PR #4, opened 2019-09-16
feat: add no-debug rule

thomlom:
This PR intends to add a new rule called `no-debug`.

Indeed, just like `console.log` statements pollutes the browser's output, debug statements can also pollute the tests:

```js
import React from 'react'

const HelloWorld = () => <h1>Hello World</h1>
const { debug } = render(<HelloWorld />)
debug()
// <div>
//   <h1>Hello World</h1>
// </div>
```

This rule would prevent developers to push unwanted debug statements to the codebase by warning them that it has been used.

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,6 +1,4 @@
 {
-  "*.js": [
-    "npm run lint:fix",
-    "git add"
-  ]
+  "*.js": ["eslint --fix", "git add"],
+  "*.md": ["prettier --write", "git add"]
 }
--- a/README.md
+++ b/README.md
@@ -66,5 +66,6 @@ To enable this configuration use the `extends` property in your `.eslintrc` conf
 | -------------------------------------------------------- | --------------------------------------------- | ---------------- |
 | [await-async-query](docs/rules/await-async-query.md)     | Enforce async queries to have proper `await`  | ![recommended][] |
 | [no-await-sync-query](docs/rules/no-await-sync-query.md) | Disallow unnecessary `await` for sync queries | ![recommended][] |
+| [no-debug](docs/rules/no-debug.md)                       | Disallow the use of `debug`                   |                  |
 
 [recommended]: https://img.shields.io/badge/recommended-lightgrey?style=flat-square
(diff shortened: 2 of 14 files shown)


thomlom:
This rule would not be useful for the base package: `@testing-library/dom` but for `'@testing-library/react'` or `'@testing-library/vue'`.
Should we implement the shareable configurations in this PR or wait a little?

Belco90:
I think is better to just have this rule now not added to recommended config and in a different task we address everything related to the frameworks configs (vue and react) and add proper doc to the readme (my idea was to include a rules table similar to [eslint-jest](https://github.com/jest-community/eslint-plugin-jest#rules) one indicating in a "framework" column if they are enabled in any framework config). Are you happy with this?

About the rule, I think you need to limit the selector for Identifiers within a CallExpresion, as just `debug` would trigger an error when it shouldn't, right? (with corresponding tests)

I guess you are still pending to update doc and including the rule in the index file as the PR is flagged as WIP

thomlom:
It looks like a good solution to me!

For the rule, it would work for a first version but there are some edge cases to consider:

- What if the user is using the [debug](https://www.npmjs.com/package/debug) package and giving it the same name? In that case, we should throw an error for every `debug` call provided it comes from the `render` method.
- But then, what if the user renames the `debug` variable as `printDOM` or `log` for example? Same question if the user assigns `debug` to a new variable:

```js
const { debug: log }  = render(...)
// OR
const { debug }  render(...)
const printDOM = debug
```

What do you think?

Belco90:
I'm happy to release a first version of those edge cases and improve them in a second step, but if you feel it would be easy to implement and should be included on first version, let's go for it. I just want to avoid including tons of different edge cases and focus on a first release with some basic rules and see what's the feedback in general about the plugin, but as I said those edge cases would need to be addressed sooner or later.

thomlom:
I think the first edge case should be included in the first release as what we want to disallow is the use of `debug` from the `render` method only instead of preventing any `debug` calls.

I agree for the second edge case, I don't see so far why you would rename `debug`. I guess we'll see once the plugin is released! 🙂

PR #5, opened 2019-09-20
Refactor: remove By replace statement

thomlom:
I don't know if the `replace` statement had a purpose to be there, but I figured a slightly simpler implementation of the generation utils.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -3,9 +3,8 @@
 const combineQueries = (variants, methods) => {
   const combinedQueries = [];
   variants.forEach(variant => {
-    const variantPrefix = variant.replace('By', '');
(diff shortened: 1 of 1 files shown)


Belco90:
I did that on purpose, so constants keeping query variants and methods mirror the same as in testing-library doc (in case it has to be tested/compared/checked in other places), but that implies kept both lists with 'By' prefix/suffix for both, so I need to remove it when combining them.

I think I prefer to leave it as it was to mirror the doc, what do you think?

thomlom:
Oh, I see now. Yes, we can leave it as is, it was just a small tweak 🙂

Belco90:
I'm gonna close this PR then. Thanks.

PR #6, opened 2019-09-20
feat: add no-debug

thomlom:
This PR intends to add a new rule called `no-debug`.

Indeed, just like `console.log` statements pollutes the browser's output, debug statements can also pollute the tests:

```js
import React from 'react'

const HelloWorld = () => <h1>Hello World</h1>
const { debug } = render(<HelloWorld />)
debug()
// <div>
//   <h1>Hello World</h1>
// </div>
```

This rule would prevent developers to push unwanted debug statements to the codebase by warning them that it has been used.

### DIFF
--- a/lib/rules/no-debug.js
+++ b/lib/rules/no-debug.js
@@ -1,0 +1,68 @@
+'use strict';
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow unnecessary debug usages in the tests',
+      category: 'Best Practices',
+      recommended: true,
+      url: 'TODO',
+    },
+    messages: {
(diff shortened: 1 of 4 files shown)


lib/rules/no-debug.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow unnecessary debug usages in the tests',
+      category: 'Best Practices',
+      recommended: true,
+      url: 'TODO',
```
Belco90: You already have `getDocsUrl` to handle this so please add corresponding URL

PR #7, opened 2019-09-20
feat: add shareable framework configs

Belco90:
This PR adds shareable framework configs (angular, react, vue) and updates corresponding documentation.

Apart from the setup itself, I've enabled `no-debug` rule for all these 3 frameworks included.

Additionally, I've added some snapshots tests for exposed configs so we have a way to test if they are changed or not flexible enough to be updated easily when adding/modifying rules. Related to this, I added couple of eslint plugins for jest as this is the first test file purely written in jest syntax.

### DIFF
--- a/tests/index.js
+++ b/tests/index.js
@@ -6,4 +6,16 @@ describe('shareable configs', () => {
   it('should export proper "recommended" config', () => {
     expect(configs.recommended).toMatchSnapshot();
   });
+
+  it('should export proper "angular" config', () => {
+    expect(configs.angular).toMatchSnapshot();
+  });
+
+  it('should export proper "react" config', () => {
+    expect(configs.react).toMatchSnapshot();
+  });
+
+  it('should export proper "vue" config', () => {
+    expect(configs.vue).toMatchSnapshot();
+  });
 });
--- a/tests/index.js
+++ b/tests/index.js
@@ -3,19 +3,10 @@
 const { configs } = require('../lib');
 
 describe('shareable configs', () => {
-  it('should export proper "recommended" config', () => {
-    expect(configs.recommended).toMatchSnapshot();
-  });
-
-  it('should export proper "angular" config', () => {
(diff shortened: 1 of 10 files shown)


tests/index.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
thomlom: What about looping over the configurations to make the test lighter?

```js
'use strict';

const { configs } = require('../lib');

describe('shareable configs', () => {
  Object.entries(configs).forEach(([configName, config]) => {
    it(`should export proper ${configName} config`, () => {
      expect(config).toMatchSnapshot()
    })
  })
});

```

Belco90: Even easier than that is to use [test.each](https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout) so we have one definition with different values. I just add them like this to specify proper configs with snapshots, but I'm happy to implement these tests like that.

Belco90:
>     * I'm not sure if snapshots are necessary or not. At least, it's a way to "lock" the configuration and a way to warn us if we modify it.
> 
>     * As for jest formatting, I'm wondering what is its value? Isn't prettier already formatting the JS files?

Exactly, snapshots it's best way to lock the config and easiest to update. About jest formatting, it doesn't format jest syntax and eslint doesn't recognize jest methods (describe, test, it, etc) so I've installed it also for that reason.

tests/index.js
thomlom: Yes, why not!

thomlom:
A simpler way to make ESLint recognize Jest methods is to add: 

```
{
  "env": {
    "jest": true
  }
}
```

to `.eslintrc.json`. But in the end, it does no harm to keep this Jest formatting plugin. As you wish!

Belco90:
Tests updated with `test.each` to avoid duplication.

thomlom:
Great!

PR #18, opened 2019-10-01
feat: add new rule await-fire-event

Belco90:
Adds new rule `await-fire-event` for waiting special async `fireEvent` methods (only for Vue).

### DIFF
--- a/lib/rules/await-fire-event.js
+++ b/lib/rules/await-fire-event.js
@@ -1,0 +1,64 @@
+'use strict';
+
+const { getDocsUrl } = require('../utils');
+
+function isAwaited(node) {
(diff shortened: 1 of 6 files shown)


lib/rules/await-fire-event.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function isPromiseResolved(node) {
+  const parent = node.parent.parent;
+
+  // fireEvent.click().then(...)
+  if (parent.type === 'CallExpression') {
+    return hasThenProperty(parent.parent);
+  }
+}
```
Belco90: These methods are partly duplicated, but I prefer to leave them like that for now as they are not exactly the same from await-async-query and see in the future how to refactor them.

PR #29, opened 2019-10-21
feat: add prefer-explicit-assert rule

### LINKED ISSUE #28: No get* to assert element is in document
I have seen some people started using get* (because it is short) to assert elements are in DOM:

```js
getByText("Text")
```
(note that there is no assignment or anything, just the call)

I would prefer if line was changed to:
```js
expect(queryByText('Text')).toBeInTheDocument()
```

It's longer, but I think we should be explicit in tests on what exactly is being tested. Also, someone new to RTL has no idea get* selectors throw and this is why one can use them to be lazy.

Belco90:
New rule: ~no-get-by-assert~ prefer-explicit-assert

This closes #28

### DIFF
--- a/docs/rules/no-get-by-assert.md
+++ b/docs/rules/no-get-by-assert.md
@@ -1,0 +1,51 @@
+# Disallow `getBy*` queries as assertion method itself (no-get-by-assert)
+
+Testing Library `getBy*` queries throw an error if the element is not
+found. Some users like this behavior to use the query itself as an
+assert for the element existence in their tests, but other users don't
+and prefer to explicitly assert the element existence, so this rule is
+for users from the latter.
+
+## Rule Details
+
+This rule aims to encourage users to explicitly assert existence of
+elements in their tests rather than just use `getBy*` queries and expect
+it doesn't throw an error so it's easier to understand what's the
+expected behavior within the test.
+
+Examples of **incorrect** code for this rule:
+
+```js
+// just calling `getBy*` query expecting not to throw an error as an
+// assert-like method, without actually using the returned element
+getByText('foo');
+```
+
+Examples of **correct** code for this rule:
+
+```js
+// wrapping the get query within a `expect` and use some matcher for
+// making the assertion more explicit
+expect(getByText('foo')).toBeDefined();
+
+// even more explicit if you use `@testing-library/jest-dom` matcher
+// for checking the element is present in the document
+expect(getByText('foo')).toBeInTheDocument();
+
+// Doing something with the element returned without asserting is absolutely fine
+await waitForElement(() => getByText('foo'));
+fireEvent.click(getByText('bar'));
+const quxElement = getByText('qux');
+
+// call directly something different than Testing Library query
+getByNonTestingLibraryVariant('foo');
+```
+
+## When Not To Use It
+
+If you prefer to use `getBy*` queries implicitly as an assert-like
+method itself, then this rule is not recommended.
+
+## Further Reading
+
+- [getBy query](https://testing-library.com/docs/dom-testing-library/api-queries#getby)
--- a/docs/rules/prefer-explicit-assert.md
+++ b/docs/rules/prefer-explicit-assert.md
@@ -1,4 +1,4 @@
-# Disallow `getBy*` queries as assertion method itself (no-get-by-assert)
+# Suggest using explicit assertions rather than just `getBy*` queries (prefer-explicit-assert)
 
 Testing Library `getBy*` queries throw an error if the element is not
 found. Some users like this behavior to use the query itself as an
--- a/docs/rules/prefer-explicit-assert.md
+++ b/docs/rules/prefer-explicit-assert.md
@@ -51,6 +51,18 @@ const quxElement = getByText('qux');
 getByNonTestingLibraryVariant('foo');
 ```
 
+## Options
+
+This rule accepts a single options argument:
+
+- `customQueries`: this array option allows to extend default Testing
+  Library queries with custom ones for including them into rule
+  inspection.
+
+  ```js
+  "testing-library/prefer-expect-query-by": ["error", {"customQueries": ["getByIcon", "getBySomethingElse"]}],
+  ```
+
 ## When Not To Use It
 
 If you prefer to use `getBy*` queries implicitly as an assert-like
--- a/docs/rules/prefer-explicit-assert.md
+++ b/docs/rules/prefer-explicit-assert.md
@@ -26,6 +26,9 @@ Examples of **incorrect** code for this rule:
 // assert-like method, without actually either using the returned element
 // or explicitly asserting
 getByText('foo');
+
+const utils = render(<Component />);
+utils.getByText('foo');
 ```
 
 Examples of **correct** code for this rule:
@@ -34,6 +37,9 @@ Examples of **correct** code for this rule:
 // wrapping the get query within a `expect` and use some matcher for
 // making the assertion more explicit
 expect(getByText('foo')).toBeDefined();
+
+const utils = render(<Component />);
+expect(utils.getByText('foo')).toBeDefined();
 
 // ⚠️ `getBy*` should be replaced by `queryBy*` when combined with `prefer-expect-query-by` rule
 expect(queryByText('foo')).toBeDefined();
--- a/docs/rules/prefer-explicit-assert.md
+++ b/docs/rules/prefer-explicit-assert.md
@@ -55,12 +55,12 @@ getByNonTestingLibraryVariant('foo');
 
 This rule accepts a single options argument:
 
-- `customQueries`: this array option allows to extend default Testing
+- `customQueryNames`: this array option allows to extend default Testing
   Library queries with custom ones for including them into rule
   inspection.
 
   ```js
-  "testing-library/prefer-expect-query-by": ["error", {"customQueries": ["getByIcon", "getBySomethingElse"]}],
+  "testing-library/prefer-expect-query-by": ["error", {"customQueryNames": ["getByIcon", "getBySomethingElse"]}],
   ```
 
 ## When Not To Use It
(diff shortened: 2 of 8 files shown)


lib/rules/no-get-by-assert.js
```diff
@@ -0,0 +1,51 @@
...
+    messages: {
+      noGetByAssert: 'Disallowed use of `getBy*` query as implicit assert',
+    },
+    fixable: null,
+    schema: [],
+  },
+
+  create: function(context) {
+    return {
+      [`CallExpression > Identifier[name=${/^getBy(LabelText|PlaceholderText|Text|AltText|Title|DisplayValue|Role|TestId)$/}]`](
```
Meemaw: Could this be extendable by the user if needed? For example we have some custom selectors e.g. `getByIcon` that I would like to fall under the same rule but should not be part of the default configuration.

lib/rules/no-get-by-assert.js
```diff
@@ -0,0 +1,51 @@
...
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Disallow `getBy*` queries as assertion method itself',
+      category: 'Best Practices',
+      recommended: false,
+      url: getDocsUrl('no-get-by-assert'),
+    },
+    messages: {
+      noGetByAssert: 'Disallowed use of `getBy*` query as implicit assert',
```
Meemaw: Is this the message user will see on hover in VSCode for example? Should it include some information to use the `expect(queryBy*)...` pattern?

lib/rules/no-get-by-assert.js
Belco90: Yes, I can make it extendable so users can add their own queries. I always have doubts about just checking `getBy*` default queries + custom ones or check everything starting by `getBy`, but I think I prefer the first one and make it extendable so you and other users can add custom queries and avoid false positives with other things.

lib/rules/no-get-by-assert.js
Belco90: Yes, this is the message you'll get when linting the code (or your IDE pass the linter for you). To be honest I don't know what to put here, so I just put something simple mirroring other rules from ESLint. Suggestions are more than welcome.

docs/rules/no-get-by-assert.md
```diff
@@ -0,0 +1,51 @@
...
+// assert-like method, without actually using the returned element
+getByText('foo');
+```
+
+Examples of **correct** code for this rule:
+
+```js
+// wrapping the get query within a `expect` and use some matcher for
+// making the assertion more explicit
+expect(getByText('foo')).toBeDefined();
```
thomlom: Shouldn't it be `expect(queryByText)` instead of `expect(getByText)`? 🤔 
Such an example would be conflicting with the `prefer-expect-query-by` rule.

docs/rules/no-get-by-assert.md
```diff
@@ -0,0 +1,51 @@
+# Disallow `getBy*` queries as assertion method itself (no-get-by-assert)
```
thomlom: The rule name sounds a bit strange to me. When it comes to preferences, ESLint rules are usually named `prefer-...`, I think it should be the case here.
`no-get-by-assert` implies here that doing assertions with `getBy` queries is wrong. I think something like `prefer-query-by-assert` is more clear to the user as discussed in the [issue](https://github.com/Belco90/eslint-plugin-testing-library/issues/28)

lib/rules/no-get-by-assert.js
thomlom: I would go for something like `Use queryBy assertions instead of getBy`. What do you think?

lib/rules/no-get-by-assert.js
```diff
@@ -0,0 +1,51 @@
...
+  create: function(context) {
+    return {
+      [`CallExpression > Identifier[name=${/^getBy(LabelText|PlaceholderText|Text|AltText|Title|DisplayValue|Role|TestId)$/}]`](
+        node
+      ) {
+        if (
+          !isDirectlyCalledByFunction(node) &&
+          !isReturnedByArrowFunctionExpression(node) &&
+          !isDeclared(node) &&
+          !isReturnedByReturnStatement(node)
```
thomlom: I really like this approach. It makes the edge cases clear and readable 👌

docs/rules/no-get-by-assert.md
Belco90: That's actually from a different rule so I would avoid having this one encouraging two things: asserting explicitly and asserting with `queryBy*`, so just leaving this rule for the former and `prefer-expect-query-by` for the latter.

docs/rules/no-get-by-assert.md
Belco90: I'm happy to rename this to `prefer-*` rule but I would avoid using query-by for same reason I commented before, so I would go for something like `prefer-explicit-assert`. What do you think?

lib/rules/no-get-by-assert.js
Belco90: Same as previous mentions to `queryBy`: user can't be forced to do that, we can only mention the fact that the `getBy` must be transformed into explicit assert.

lib/rules/no-get-by-assert.js
Belco90: I like it too, but I'm always concerned about if I left one edge case out 😨

docs/rules/no-get-by-assert.md
thomlom: `prefer-explicit-assert` sounds good to me!

docs/rules/no-get-by-assert.md
thomlom: Yes, but as `prefer-expect-query-by` is in the recommended rules, I found it a little bit weird to say `expect(getBy)` is correct (although it is for this rule technically speaking) 🙂

Belco90: Well, in the example of this specific rule that's fine. It would be weird to see `expect(queryByText('foo')).toBeDefined();` when the rule only talk about `getBy*`, so I wanted to clarify that specific use is valid for this rule.

thomlom: Sure! What about putting a side note stating that though `expect(getBy)` is valid for this rule, it is recommended to use `queryBy` for asserting elements' presence? Then, the user would know it's valid to use `expect(getBy)` while still being aware it's not recommended. What do you think? 🙂

Belco90: That sounds absolutely reasonable. I'll address the feedback here later.

emmenko: My 2 cents: I understand both points. However, since this is a documentation to something that is to be considered "valid", I would expect that we write things that are "valid", which is not really the case with `expect(getBy)`, as we also have a recommended rule for it.

At least we should mention to wrap it into an `expect` and then to change it to `queryBy`, and link it to the other rule for more reference.

Belco90:
All feedback for renaming the rule, add proper references to `prefer-expect-query-by` rule, add option for custom queries and improve error message have been addressed. Please let me know what you think now :)

lib/rules/prefer-explicit-assert.js
```diff
@@ -0,0 +1,73 @@
...
+    type: 'suggestion',
+    docs: {
+      description:
+        'Suggest using explicit assertions rather than just `getBy*` queries',
+      category: 'Best Practices',
+      recommended: false,
+      url: getDocsUrl('prefer-explicit-assert'),
+    },
+    messages: {
+      noGetByAssert:
```
thomlom: For consistency, I think it's better to rename the message id too: `preferExplicitAssert`

Belco90: Oops, good catch! Let me fix that.

Belco90: Done, I need your approval again 😛

docs/rules/prefer-explicit-assert.md
```diff
@@ -0,0 +1,77 @@
...
+
+// call directly something different than Testing Library query
+getByNonTestingLibraryVariant('foo');
+```
+
+## Options
+
+This rule accepts a single options argument:
+
+- `customQueries`: this array option allows to extend default Testing
```
emmenko: Nit: maybe `customQueryNames`?

tests/lib/rules/prefer-explicit-assert.js
```diff
@@ -0,0 +1,86 @@
...
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018 } });
+ruleTester.run('prefer-explicit-assert', rule, {
+  valid: [
+    {
+      code: `getByText`,
```
emmenko: Can you also check for when queries are not destructured?

```js
const rendered = render(<App />)

rendered.getByText('...')
// ...
```

docs/rules/prefer-explicit-assert.md
Belco90: That was actually my first approach and then I renamed it, so I'm happy to go back to that one if it feels the right name.

lib/rules/prefer-explicit-assert.js
thomlom: Just saw @emmenko comment on another test case so I'll wait for it to be included before approving 😉

tests/lib/rules/prefer-explicit-assert.js
Belco90: Nice spot! This actually found some issues in the rule, so I'm adding those tests and fixing the rule implementation.

docs/rules/prefer-explicit-assert.md
emmenko: Ah ok, I think it's easier to understand if the name is more specific. I don't have a strong opinion though, up to you to decide.

Belco90: Done, I agree with you in having it more specific.

tests/lib/rules/prefer-explicit-assert.js
Belco90: Tests added, rule fixed and more examples added to the doc

PR #37, opened 2019-10-28
chore: enforce 100% coverage on CI

### LINKED ISSUE #36: Code coverage improvements
Now that we have 100% coverage in the repo, it would be nice to set proper coverage threshold and check it on corresponding PRs.

For accomplishing this we would have to:

- set coverage threshold to 100%
- add necessary script to execute the tests in coverage mode
- execute the coverage mode in the CI
- prevent PRs to be merged in coverage has been decreased

thomlom:
Closes #36

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,11 @@
 module.exports = {
   testMatch: ['**/tests/**/*.js'],
+  coverageThreshold: {
+    global: {
+      branches: 100,
+      functions: 100,
+      lines: 100,
+      statements: 100,
+    },
+  },
 };
--- a/package.json
+++ b/package.json
@@ -28,7 +28,9 @@
     "lint": "eslint ./",
     "lint:fix": "npm run lint -- --fix",
     "format": "prettier --write README.md {lib,docs,tests}/**/*.{js,md}",
-    "test": "jest",
+    "test:local": "jest",
+    "test:ci": "jest --coverage",
+    "test": "is-ci test:ci test:local",
     "semantic-release": "semantic-release"
   },
   "dependencies": {},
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -28,7 +28,9 @@
     "lint": "eslint ./",
     "lint:fix": "npm run lint -- --fix",
     "format": "prettier --write README.md {lib,docs,tests}/**/*.{js,md}",
-    "test": "jest",
+    "test:local": "jest",
+    "test:ci": "jest --coverage",
+    "test": "is-ci test:ci test:local",
```
emmenko: Curious as to why not simply running `test:ci` on CI? It seems like an unnecessary setup for me... 🤔

jest.config.js
```diff
@@ -1,3 +1,11 @@
 module.exports = {
   testMatch: ['**/tests/**/*.js'],
+  coverageThreshold: {
```
Belco90: Didn't know that this already prevents CI from continuing as it makes jest fails if threshold is not met!

package.json
Belco90: Mmm it's interesting: on one hand doesn't seem like a necessary dependency as you mentioned, but on the other hand it simplifies the execution of tests and external scripts to just `npm test` without doing something special (we would have to customize travis job to execute `test:ci` rather than default `test` script). So I think is worth the cost of having this new dependency to simplify CI config around tests.

thomlom: If I'm not mistaken, the default Travis CI configuration for Node.js includes a `test` script. I just didn't want to modify the CI config 🙂

jest.config.js
thomlom: Jest is awesome for that. Just set a coverage threshold and you're good to go as long as you run `jest --coverage`!

package.json
benmonro: why have these 2 seperated?  if you want to run w/out coverage you can do `--no-coverage`

emmenko: It's fine, was just curious if there was some other reason 🙂

Belco90: @benmonro I usually prefer having 2 different tests scripts for local and ci just in case there are more things to set up for each (there is another project where I had to set some specific env vars and junit reporter), so it's easier to maintain in that way I think. And even better: it's transparent from executing the script with the package that Thomas added :)

PR #39, opened 2019-10-31
fix: prefer-explicit-assert options example

Meemaw:


thomlom:
Thank you for spotting the typo and correcting it! 👍

Belco90:
This PR type should have been `docs` instead of `fix` so 1) it has proper type and 2) it doesn't trigger a new package release (not necessary for this kind of change)

image

Belco90:
@all-contributors please add @Meemaw for docs

PR #42, opened 2019-11-01
Avoid raising preferExplicitAssert warning on destructuring

### LINKED ISSUE #40: prefer-explicit-assert does not work as expected
Screen Shot 2019-10-31 at 4 28 35 PM

afontcu:
Closes #40.

I was trying to get my head around the issue linked above, but I can't seem to be able to write a test case that reproduces the behavior we are getting in the wild.

Right now, tests pass/fail as the library should behave: `valid` test is passing, while `invalid` tests are failing (meaning, AFAIK, that given that piece of code no warning was raised).
However, in the real world™, `const { getByText } = render(Comp)` is raising an error. 

I opened up a draft PR in order to get this sorted out, so any help would be appreciated :)


afontcu:
(Build obviously fails as I had to add `--no-verify` flag when commiting)

thomlom:
Thanks for opening the PR @afontcu. I'm investigating it at the moment. So far, it seems that the test fails when the destructuring statement is wrapped in a callback function. If you add the test below in the `valid` tests, you'll get two errors:

```js
    {
      code: `
      describe("App", () => {
        it("test", () => {
          const { getByText } = render(<Preview/>)
        })
      })
      `,
    },
```

I don't know why at the moment but I'm taking a look at it 😉

thomlom:
Ok, I found where the bug comes from. The selector used by the visited function `'CallExpression Identifier'(node)` is incorrect since it triggers the function for all descendants of a `CallExpression`. 

As `getByText` is an identifier inside an `it` call, it will be considered and will trigger the rule. Here is two screenshots to better understand what's going on:

**Descendants** selector:

Screenshot 2019-11-01 at 15 09 00

**Children selector**:

Screenshot 2019-11-01 at 15 08 48

Doing it this way seems a good solution at first sight but it makes other tests fail, particularly this one:

```js
    ...ALL_QUERIES_METHODS.map(queryMethod => ({
      code: `const utils = render()

      utils.get${queryMethod}('foo')`,
      errors: [
        {
          messageId: 'preferExplicitAssert',
          line: 3,
          column: 13,
        },
      ],
    })),
```

thomlom:
I found a fix to the bug, I'm doing a check on if the visited node is in a destructuring statement or not: 

```js
const isInDestructuringStatement = node =>
  node.parent.type === 'Property' &&
  node.parent.parent.type === 'ObjectPattern';
```

And in the big `if` statement:

```js
        if (
          isValidQuery(node, customQueryNames) &&
          !isInDestructuringStatement(node) &&
          ...
```

Do you want to take it from here and add the corresponding tests @afontcu ?

afontcu:
sure, I take a look! It's my first toying around with eslint rules 🙌 

I'll ping you as soon as I get it ready to review :)
thanks!

afontcu:
Closing in favor of #43 (something got messed up with git history, and didn't want to spend much time on it).

PR #43, opened 2019-11-01
Fix prefer-explicit-assert raising error on destructuring statement

### LINKED ISSUE #40: prefer-explicit-assert does not work as expected

afontcu:
Closes #40.

I've added a check for Array destructuring. I've never seen it before, but I thought it wouldn't hurt.

Thanks for your help!

### DIFF
--- a/lib/rules/prefer-explicit-assert.js
+++ b/lib/rules/prefer-explicit-assert.js
@@ -24,6 +24,11 @@ const isDeclared = node => node.parent.type === 'VariableDeclarator';
 const isReturnedByReturnStatement = node =>
   node.parent.type === 'ReturnStatement';
 
+const isInDestructuringStatement = node =>
+  (node.parent.type === 'Property' &&
+    node.parent.parent.type === 'ObjectPattern') ||
+  node.parent.type === 'ArrayPattern';
+
 module.exports = {
   meta: {
     type: 'suggestion',
@@ -63,6 +68,7 @@ module.exports = {
 
         if (
           isValidQuery(node, customQueryNames) &&
+          !isInDestructuringStatement(node) &&
           !isDirectlyCalledByFunction(callExpressionNode) &&
           !isReturnedByArrowFunctionExpression(callExpressionNode) &&
           !isDeclared(callExpressionNode) &&
--- a/lib/rules/prefer-explicit-assert.js
+++ b/lib/rules/prefer-explicit-assert.js
@@ -1,13 +1,13 @@
 'use strict';
 
-const { getDocsUrl, ALL_QUERIES_METHODS } = require('../utils');
+const { findParent, getDocsUrl, ALL_QUERIES_METHODS } = require('../utils');
 
 const ALL_GET_BY_QUERIES = ALL_QUERIES_METHODS.map(
   queryMethod => `get${queryMethod}`
 );
 
 const findCallExpressionParent = node =>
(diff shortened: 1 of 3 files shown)


lib/rules/prefer-explicit-assert.js
```diff
@@ -24,6 +24,12 @@ const isDeclared = node => node.parent.type === 'VariableDeclarator';
 const isReturnedByReturnStatement = node =>
   node.parent.type === 'ReturnStatement';
 
+const isInDestructuringStatement = node =>
+  (node.parent.type === 'Property' &&
+    node.parent.parent.type === 'ObjectPattern') ||
+  (node.parent.type === 'ArrayPattern' &&
+    node.parent.parent.type === 'VariableDeclarator');
```
thomlom: I don't think checking `VariableDeclarator` is necessary, is it?

afontcu: oh, I see, it's not. Fixed!

This made me think of something else, though. Should the following line be considered as a valid one? Right now it is consideren invalid:

`const a = [ getByText('foo'), 'bar' ]`

thomlom: That's interesting... I guess it should be considered as valid since we don't use it as an assertion here. We need to fix that as well.

Thanks for spotting the issue! 🙂

afontcu: okay, I'll work on that and update the PR accordingly! :)

thomlom: All right! Thanks for working on that 😄

afontcu: Okay, just updated the PR. I think checking if grandparent is an `ArrayExpression` will suffice, but please do doublecheck it as I'm relying on tests I wrote and the AST explorer which is just learned about when you shared the screenshots 😇 

thanks for helping me out!

Meemaw:
Any estimates on when this will be merged? Great work regardless! 👍

afontcu:
> Any estimates on when this will be merged? Great work regardless! 👍

AFAICT it's ok from my side, let's see if Thomas can take a look :)

lib/rules/prefer-explicit-assert.js
thomlom: I got it working by doing this: 

1. Add a `findParent` function in the `utils` file:

```javascript

const findParent = (node, cb) => {
  if (cb(node)) {
    return node;
  } else if (node.parent) {
    return findParent(node.parent, cb);
  } else {
    return null;
  }
};
```

2. Modify both `isDeclared` and `findCallExpressionParent`:

```javascript
const findCallExpressionParent = node =>
  findParent(node, node => node.type === 'CallExpression');

// ... 

const isDeclared = node => {
  return !!findParent(node, node => node.type === 'VariableDeclarator');
};
```

3. I removed the `isInArrayDeclaration` function

4. I added a last test:

```javascript
    {
      code: `const a = { foo: getByText('bar') }`,
    },
```

I think it should be good by doing so 🙂

afontcu: It looks way smarter to check for positive cases rather than trying to take into account every posssible wrong edge case. Updating the PR.

afontcu:
Thank you for guiding me through!

thomlom:
@all-contributors please add @afontcu  for code and test.

PR #56, opened 2019-12-19
feat(rule): add data-testid rule

tknickman:
Here is the PR for the `data-testid` rule. Let me know what you think! Happy to make any adjustments necessary. 

Fixes https://github.com/Belco90/eslint-plugin-testing-library/issues/55


README.md
```diff
@@ -138,6 +138,7 @@ To enable this configuration use the `extends` property in your
 | [no-dom-import](docs/rules/no-dom-import.md)                   | Disallow importing from DOM Testing Library                         | ![angular-badge][] ![react-badge][] ![vue-badge][]                        | ![fixable-badge][] |
 | [prefer-expect-query-by](docs/rules/prefer-expect-query-by.md) | Disallow the use of `expect(getBy*)`                                | ![recommended-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |                    |
 | [prefer-explicit-assert](docs/rules/prefer-explicit-assert.md) | Suggest using explicit assertions rather than just `getBy*` queries |                                                                           |                    |
+| [data-testid](docs/rules/data-testid.md)                       | Ensure `data-testid` values match a provided regex.                 |                                                                           |                    |
```
emmenko: Nitpick: I feel like `data-testid` is a bit confusing, maybe name this rule something like `consistent-data-testid`?

docs/rules/data-testid.md
```diff
@@ -0,0 +1,46 @@
...
+const bar = props => <div data-testid="TestId">...</div>;
+
+const baz = props => <div>...</div>;
+```
+
+## Options
+
+| Option        | Details                                                                                                                                                                                                                                                           | Example                                       |
+| ------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------- |
+| testIdPattern | A regex used to validate the format of the `data-testid` value. `{componentName}` can optionally be used as a placeholder and will be substituted with the name of the file OR the name of the files parent directory in the case when the fileName is `index.js` | `'^{componentName}(\_\_([A-Z]+[a-z]_?)+)_\$'` |
```
emmenko: Shouldn't be `fileName` instead of `componentName`? 🤔

docs/rules/data-testid.md
```diff
@@ -0,0 +1,46 @@
...
+| excludePaths  | An array of path strings to exclude from the check                                                                                                                                                                                                                | `["__tests__"]`                               |
```
emmenko: I'm wondering if this is really necessary, as it can be achieved with the eslint config `overrides`. Something like this I suppose:

```js
{
  "rules": {
    
  },
  "overrides": [
    {
      "files": ["*.js"],
      "ignorePatterns": "__tests__",
      "rules": {
        "testing-library/consistent-data-testid": [
          2,
          { "testIdPattern": "^TestId(__[A-Z]*)?$" }
        ]
      }
    }
  ]
}
```

README.md
Belco90: This name sounds better imho

docs/rules/data-testid.md
Belco90: Indeed. What would `componentName` be if the component is `MyComponent` but the file is called `my-component.js`?

README.md
tknickman: Agreed. Will update it!

docs/rules/data-testid.md
tknickman: Yep agreed I think that makes more sense here. 

Ideally we could figure out the component name via the AST but that's pretty non-trivial and expensive. Internally we enforce file structure and component organization to make sure each component is contained within it's own file, so this implicit filename -> component name assumption works for us there - but that is definitely not a universal assumption so I will change this, thanks for pointing it out!

Also, this was also originally inspired from the BEM stylelint plugin, which uses similar logic to infer componentName in scss files (https://github.com/postcss/postcss-bem-linter#define-components-and-utilities-implicitly-based-on-their-filename).

tknickman:
Thanks for the review! I made the requested changes here: https://github.com/Belco90/eslint-plugin-testing-library/pull/56/commits/c7aa60f0e957252badd18f0ce48958c07276750f

Let me know what you guys think, happy to continue iterating on it 😄

tknickman:
@Belco90 no problem!. That's a great call, I totally forgot that was configurable now in RTL. I'll add that as a configuration option.

tknickman:
Forgot to update the docs, that commit is coming in 2 mins!

Belco90:
Also, the test coverage should be increased: https://travis-ci.org/Belco90/eslint-plugin-testing-library/jobs/636430039

I don't know why this wasn't blocking the PR, need to check it in a different ticket.

tknickman:
Coverage and docs updated, let me know if there's anything else I can help with!

Belco90:
Brilliant, merging now for real.

Belco90:
@all-contributors please add @tknickman for code, doc and test

tknickman:
Thanks!

Issue #59, opened 2020-01-15
Remove `prefer-expect-query-by` rule from shareable configs


Belco90:
After checking `prefer-expect-query-by` rule with different people and getting different feedback about it, this rule is definitely too opinionated to be enabled by default in shareable configs. Also, the motivation of the rule is not really clear, so we need to improve the rule doc to clarify what's the use case and corresponding motivation for it.

emmenko:
Do you know what the confusion is about? Is the doc not clear enough? The main use case is when asserting that an element is not in the document, in which case it's recommended to use `expect(queryBy` as [mentioned in the docs](https://testing-library.com/docs/guide-disappearance#asserting-elements-are-not-present). The rule then ensures some sort of consistency of always using `queryBy` with `expect`.

Anyway, it's fine for me to not have it enabled by default, I'm just curious to know what people find confusing about it.

Belco90:
I found this chat in spectrum about it: https://spectrum.chat/testing-library/general/prefer-expect-query-by-linting-rule~802a01db-1afd-4ca3-b753-1dfcfcc5d1ca

After that, I discussed with @thomlom about the rule and we agreed it looks too opinionated to be enabled by recommended preset. Then I shared with some coworkers and they found it as a weird rule by default too.

About if doc is clear enough or not, examples could be improved and I would leave clear what use cases for this rules are (adding link to [jest-dom section in testing library](https://testing-library.com/docs/ecosystem-jest-dom) doc would help).

If you have anything else in mind we could improve here, let me know. I'll create a PR soon so we can improve it there too.

emmenko:
I see, thanks for the clarification. Yeah let's do it like this then.

Belco90:
I guess that ideally we could have a rule enabled by default  that forces you to use `queryBy` instead of `getBy` when asserting elements are not present, but I'm not sure how to implement that (checking `expect().not.toBeInTheDocument` and `expect().toBeNull()` or something similar? I'm afraid we would have to check tons of combination for different matchers and negations), so I think best think we can do for now is just not enabling it by default and specify which are the use cases.

Belco90:
Closed by #60

Belco90:
[...]

PR #60, opened 2020-01-15
Remove `prefer-expect-query-by` rule from shareable configs

Belco90:
As mentioned in the original issue #59 :

- remove `prefer-expect-query-by` rule from shareable configs
- improve doc

### DIFF
--- a/docs/rules/prefer-expect-query-by.md
+++ b/docs/rules/prefer-expect-query-by.md
@@ -29,6 +29,12 @@ test('some test', () => {
 test('some test', async () => {
   const utils = render(<App />);
   await wait(() => {
+    expect(utils.getByText('Foo')).toBeInTheDocument();
+  });
+  await wait(() => {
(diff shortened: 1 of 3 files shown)


docs/rules/prefer-expect-query-by.md
```diff
@@ -24,12 +26,14 @@ test('some test', () => {
 ```
 
 ```js
-test('some test', () => {
-  const rendered = render(<App />);
-  expect(rendered.getByText('Foo')).toBeInTheDocument();
-  expect(rendered.getAllByText('Foo')[0]).toBeInTheDocument();
```
emmenko: Should we also keep the examples without `not`? At the end it's still valid to use `expect(queryByText('Foo')).toBeInTheDocument()`

Belco90: Of course 🤦‍♂️. Let me get this back.

Belco90: Fixed

PR #64, opened 2020-01-20
feat: add support for resolves/rejects matchers in async queries

### LINKED ISSUE #51: Support resolves/rejects matchers for asynchronous queries
As stated in #48, we are not handling the case where a user wraps a query in an `expect` statement and then uses the `resolves` or `rejects` matchers. For example, the following statement will trigger an error even if it's a valid case:

```js
expect(findByText("foo")).resolves.toBe("bar")
```

thomlom:
Closes #51

### DIFF
--- a/lib/rules/await-async-query.js
+++ b/lib/rules/await-async-query.js
@@ -30,7 +30,11 @@ module.exports = {
     const testingLibraryQueryUsage = [];
     return {
       [`CallExpression > Identifier[name=${ASYNC_QUERIES_REGEXP}]`](node) {
-        if (!isAwaited(node.parent.parent) && !isPromiseResolved(node)) {
+        if (
+          !isAwaited(node.parent.parent) &&
+          !isPromiseResolved(node) &&
+          !hasClosestExpectResolvesRejects(node)
+        ) {
           testingLibraryQueryUsage.push(node);
         }
       },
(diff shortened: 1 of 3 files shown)


lib/rules/await-async-query.js
```diff
@@ -101,3 +105,18 @@ function isPromiseResolved(node) {
   // promise.then(...)
   return hasAThenProperty(parent);
 }
+
+function hasClosestExpectResolvesRejects(node) {
+  if (!node.parent) {
```
emmenko: Is this necessary? Because on L34 the function is called with `isAwaited(node.parent.parent)`, which implies that `node.parent` cannot be `undefined`.

thomlom: Good catch. I initially wrote a utility function called `findClosestExpectCall`that was supposed to be reusable. At that time the `node.parent` thing was necessary because the function was recursive. Then I refactored it and indeed, it's not necessary.

thomlom:
> Awesome! Just one small thing: can you add some examples using `resolves` and `rejects` to `await-async-query` valid examples within its doc?
> 
> Btw those util methods you added for checking this look like could be extracted in the future for checking proper resolves/rejects in other rules, so let's keep an eye on that in case we can reuse it in the future 💪

About the util methods, we could look into converting the codebase to typescript, especially this package: [typescript-eslint/experimental-utils](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/experimental-utils)!

emmenko:
> About the util methods, we could look into converting the codebase to typescript, especially this package

I would also be in favor of that, as it makes working with the AST much easier and safer with the type declarations.

lib/rules/await-async-query.js
thomlom: I tried to remove the `node.parent` check and test broke which is, in the end, logical. While `isAwaited(node.parent.parent)` implies `node.parent` cannot be `undefined`, it's very possible that `hasClosestExpectResolvesRejects` goes beyond `node.parent.parent` (until the `Program` node) in case you're not using an `expect` statement.

thomlom:
Can we merge it then @Belco90 ?

Belco90:
Sure! On it, sorry.

Belco90:
> > About the util methods, we could look into converting the codebase to typescript, especially this package
> 
> I would also be in favor of that, as it makes working with the AST much easier and safer with the type declarations.

I think I'll create a project within this github repo so we can discuss which things we need to address to migrate the plugin to TS and split proper work.

PR #65, opened 2020-01-21
feat: no-get-by-for-checking-element-not-present

### LINKED ISSUE #61: Replace `prefer-expect-query-by` by better rule
Motivated by #59 it looks like we need to replace `prefer-expect-query-by`* by better implementation to only force users to replace `getBy` queries by `queryBy` when:

- [asserting elements are not present](https://testing-library.com/docs/guide-disappearance#asserting-elements-are-not-present)
- [waiting for disappearance](https://testing-library.com/docs/guide-disappearance#waiting-for-disappearance)

This should be checked in several cases. A few I can think of at the moment:

- asserting with `.not.toBeInTheDocument()`
- asserting with `.toBeNull()`
- asserting with `.toBe(null)` _this is actually an equivalent for the previous one, probably more variations to cover as `.toBeFalsy()` but I think just this one is enough for now?_
- waiting for disappearance with `waitForElementToBeRemoved`

I can't think better name for this rule than `no-get-by-for-element-absence`, any suggestions?

*_We could maintain `prefer-expect-query-by` when the new rule is implemented, but I think it wouldn't be that useful anymore and could even lead to confusion_

thomlom:
Closes #61


thomlom:
@Belco90 If you ever see something missing in the docs or misexplained, could you handle it? I have to admit I'm still not really convinced by the usefulness of that rule and I think you have the use cases in mind.

lib/index.js
```diff
@@ -7,7 +7,7 @@ const rules = {
   'no-await-sync-query': require('./rules/no-await-sync-query'),
   'no-debug': require('./rules/no-debug'),
   'no-dom-import': require('./rules/no-dom-import'),
-  'prefer-expect-query-by': require('./rules/prefer-expect-query-by'),
+  'no-get-by-for-absent-elements': require('./rules/no-get-by-for-absent-elements'),
```
emmenko: Naming suggestion: I find "absent" a bit confusing, maybe something like `no-get-by-for-asserting-element-not-present`, which has a more direct relation to the documentation page [Asserting elements are not present](https://testing-library.com/docs/guide-disappearance#asserting-elements-are-not-present). Thoughts?

tests/lib/rules/no-get-by-for-absent-elements.js
```diff
@@ -0,0 +1,70 @@
...
+const ruleTester = new RuleTester({
+  parserOptions: { ecmaVersion: 2017, sourceType: 'module' },
+});
+
+const getByVariants = ALL_QUERIES_METHODS.reduce(
+  (variants, method) => [...variants, ...[`get${method}`, `getAll${method}`]],
+  []
+);
+
+ruleTester.run('prefer-expect-query-by', rule, {
```
emmenko:
```suggestion
ruleTester.run('no-get-by-for-absent-elements', rule, {
```

tests/lib/rules/no-get-by-for-absent-elements.js
```diff
@@ -0,0 +1,70 @@
...
+      {
+        code: `expect(${queryName}('Hello')).toBeUndefined()`,
+        errors: [{ messageId: 'expectQueryBy' }],
+      },
+      {
+        code: `expect(${queryName}('Hello')).toBeFalsy()`,
+        errors: [{ messageId: 'expectQueryBy' }],
+      },
+      {
+        code: `expect(rendered.${queryName}('Hello')).toBeFalsy()`,
```
emmenko: We should probably define all combinations for both `${queryName}` and `rendered.${queryName}`, just to make sure it works in both cases.

docs/rules/no-get-by-for-absent-elements.md
```diff
@@ -1,75 +1,62 @@
...
+  expect(getAllByText('Foo')).toBeTruthy();
 });
 ```
 
 ```js
 test('some test', async () => {
   const utils = render(<App />);
-  await wait(() => {
+  await waitForElementToBeRemoved(() => {
     expect(utils.queryByText('Foo')).toBeInTheDocument();
```
emmenko: Shouldn't this be (same below)

```suggestion
    return utils.queryByText('Foo')
```

docs/rules/no-get-by-for-absent-elements.md
```diff
@@ -1,75 +1,62 @@
-# Disallow the use of `expect(getBy*)` (prefer-expect-query-by)
+# Disallow the use of `expect(getBy*)` when elements may be asbent (no-get-by-for-absent-elements)
```
tdeekens:
```suggestion
# Disallow the use of `expect(getBy*)` when elements may be absent (no-get-by-for-absent-elements)
```

docs/rules/no-get-by-for-absent-elements.md
Belco90: Indeed. There are two cases for this rule as mentioned in the issue: asserting elements are not present (rule must check this when not.toBeInTheDocument() or toBeNull() found) or waiting for element to be removed (this is not an assertion but an async util that must no use getBy inside).

This is also why I suggested “absence” in the rule so it does not refer directly to any of the motivation for the rule itself (“not present” and “disappearing”)

Belco90: I can elaborate more tomorrow as I’m on mobile now

lib/index.js
thomlom: I like it better that way, it's clearer than `no-get-by-for-absent-elements` 🙂

Belco90: Sorry but this name is not correct. As I mentioned before, this rule doesn't always apply to an assertion with `expect` so the rule can't be named like that (I'm adding a deeper review for it as it seems this rule is always checking for assertions). So we would have to go back to its previous name or rename it to something like `no-get-by-for-checking-element-not-present`

emmenko: Yeah sounds good. As long as the rule name gives a good indication of what it does, I'm fine with any name.

I think `no-get-by-for-checking-element-not-present` is also good 👌

docs/rules/no-get-by-for-asserting-element-not-present.md
```diff
@@ -1,75 +1,58 @@
...
 test('some test', async () => {
   const utils = render(<App />);
-  await wait(() => {
+  await waitForElementToBeRemoved(() => {
     expect(utils.getByText('Foo')).toBeInTheDocument();
   });
-  await wait(() => {
+  await waitForElementToBeRemoved(() => {
     expect(utils.getAllByText('Foo')).toBeInTheDocument();
   });
```
Belco90: This examples are wrong and it's my fault. You are not supposed to assert inside `waitForElementToBeRemoved`, but just pass a callback with the query for the element you want to wait for being removed, so you need to remove all the `expect` within `waitForElementToBeRemoved`. This piece of the doc probably misled everyone here, sorry.

docs/rules/no-get-by-for-asserting-element-not-present.md
```diff
@@ -1,75 +1,58 @@
...
-  await wait(() => {
-    expect(utils.queryAllByText('Foo')).toBeInTheDocument();
-  });
-  await wait(() => {
-    expect(utils.queryByText('Foo')).not.toBeInTheDocument();
-  });
-  await wait(() => {
-    expect(utils.queryAllByText('Foo')).not.toBeInTheDocument();
-  });
+  await waitForElementToBeRemoved(() => utils.queryByText('Foo'));
```
Belco90: This example is fine as it's not asserting inside.

docs/rules/no-get-by-for-asserting-element-not-present.md
```diff
@@ -1,75 +1,58 @@
...
+  await waitForElementToBeRemoved(() => utils.queryAllByText('Foo'));
```
Belco90: This one is wrong as you can't use a query returning arrays within `waitForElementToBeRemoved`, so this is another reason to remove `getAllBy` from this rule.

lib/rules/no-get-by-for-asserting-element-not-present.js
```diff
@@ -0,0 +1,78 @@
...
+const { getDocsUrl } = require('../utils');
+
+const falsyMatchers = ['toBeNull', 'toBeFalsy', 'toBeUndefined'];
+
+module.exports = {
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description:
+        'Disallow the use of getBy* queries in expect calls when elements may be absent',
```
Belco90:
```suggestion
        'Disallow the use of `getBy*` queries when checking elements are not present',
```

tests/lib/rules/no-get-by-for-asserting-element-not-present.js
```diff
@@ -0,0 +1,70 @@
...
+ruleTester.run('prefer-expect-query-by', rule, {
+  valid: getByVariants.reduce(
+    (validRules, queryName) => [
+      ...validRules,
+      { code: `expect(${queryName}('Hello')).toBeInTheDocument()` },
+      { code: `expect(${queryName}('Hello')).toBe("foo")` },
+      { code: `expect(${queryName}('Hello')).toBeTruthy()` },
+      { code: `expect(rendered.${queryName}('Hello')).toEqual("World")` },
+      { code: `expect(rendered.${queryName}('Hello')).not.toBeFalsy()` },
+    ],
```
Belco90: We need valid tests for `waitForElementToBeRemoved`, which is the second part of the rule.

tests/lib/rules/no-get-by-for-asserting-element-not-present.js
```diff
@@ -0,0 +1,70 @@
...
+    []
+  ),
+  invalid: getByVariants.reduce(
```
Belco90: We should apply this only to `getBy`

tests/lib/rules/no-get-by-for-asserting-element-not-present.js
```diff
@@ -0,0 +1,70 @@
...
+      {
+        code: `expect(${queryName}('Hello')).toBeNull()`,
+        errors: [{ messageId: 'expectQueryBy' }],
+      },
+      {
+        code: `expect(${queryName}('Hello')).not.toBeTruthy()`,
+        errors: [{ messageId: 'expectQueryBy' }],
+      },
+      {
+        code: `expect(${queryName}('Hello')).toBeUndefined()`,
```
Belco90: I'm not sure about this one. `toBeUndefined` won't work even using `queryBy` as it returns `null` so we should not include this one in the rule probably.

tests/lib/rules/no-get-by-for-asserting-element-not-present.js
```diff
@@ -0,0 +1,70 @@
...
+        code: `expect(rendered.${queryName}('Hello')).toBeFalsy()`,
+        errors: [{ messageId: 'expectQueryBy' }],
+      },
+      {
+        code: `expect(rendered.${queryName}('Hello')).not.toBeTruthy()`,
+        errors: [{ messageId: 'expectQueryBy' }],
+      },
+      {
+        code: `(async () => {
+          await waitForElementToBeRemoved(() => {
```
Belco90: I would other variants for this one having arrow function returning implicitly and having a regular function.

README.md
```diff
@@ -134,16 +134,16 @@ To enable this configuration use the `extends` property in your
...
+| [no-dom-import](docs/rules/no-dom-import.md)                                                          | Disallow importing from DOM Testing Library                                 | ![angular-badge][] ![react-badge][] ![vue-badge][]                        | ![fixable-badge][] |
+| [prefer-explicit-assert](docs/rules/prefer-explicit-assert.md)                                        | Suggest using explicit assertions rather than just `getBy*` queries         |                                                                           |                    |
+| [consistent-data-testid](docs/rules/consistent-data-testid.md)                                        | Ensure `data-testid` values match a provided regex.                         |                                                                           |                    |
+| [no-get-by-for-asserting-element-not-present](docs/rules/no-get-by-for-asserting-element-not-present) | Disallow the use of `getBy*` queries when checking elements are not present |                                                                           |                    |
```
Belco90: Can you reorder rows so they are sorted alphabetically please?

docs/rules/no-get-by-for-asserting-element-not-present.md
```diff
@@ -0,0 +1,58 @@
...
+- `expect` is used to assert element does not exist with `.not.toBeInTheDocument()` or `.toBeNull()` matchers
+- `waitForElementToBeRemoved` async util is used to wait for element to be removed from DOM
+
+Examples of **incorrect** code for this rule:
+
+```js
+test('some test', () => {
+  const { getByText } = render(<App />);
+  expect(getByText('Foo')).not.toBeInTheDocument();
+  expect(getByText('Foo')).toBeFalsy();
```
Belco90: Can you add another example here with `toBeNull()` please=

docs/rules/no-get-by-for-asserting-element-not-present.md
```diff
@@ -0,0 +1,58 @@
...
+});
+```
+
+```js
+test('some test', async () => {
+  const utils = render(<App />);
+  await waitForElementToBeRemoved(() => utils.getByText('Foo')).toBeInTheDocument());
```
Belco90: The assertion here is wrong, this should be just:
```suggestion
  await waitForElementToBeRemoved(() => utils.getByText('Foo')));
```

tests/lib/rules/no-get-by-for-asserting-element-not-present.js
```diff
@@ -0,0 +1,90 @@
...
+ruleTester.run('prefer-expect-query-by', rule, {
+  valid: [
+    ...getByQueries.reduce(
+      (validRules, queryName) => [
+        ...validRules,
+        ...getValidAssertion(queryName, '.toBeInTheDocument()'),
+        ...getValidAssertion(queryName, '.toBe("foo")'),
+        ...getValidAssertion(queryName, '.toBeTruthy()'),
+        ...getValidAssertion(queryName, '.toEqual("World")'),
+        ...getValidAssertion(queryName, '.not.toBeFalsy()'),
```
Belco90: Can you add `.not.toBeNull()` here?

tests/lib/rules/no-get-by-for-asserting-element-not-present.js
```diff
@@ -0,0 +1,90 @@
...
+        {
+          code: `(async () => {
+            await waitForElementToBeRemoved(() => {
+              return ${queryName}("hello")
+            })
+          })()`,
+        },
+        {
+          code: `(async () => {
+            await waitForElementToBeRemoved(() =>  ${queryName}("hello"))
```
Belco90: What about 3rd test passing a regular function to `waitForElementToBeRemoved`?

docs/rules/no-get-by-for-asserting-element-not-present.md
thomlom: Looks like I `git commit` too fast 🤷‍♂

docs/rules/no-get-by-for-asserting-element-not-present.md
```diff
@@ -0,0 +1,58 @@
...
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+test('some test', () => {
+  const { getByText } = render(<App />);
+  expect(getByText('Foo')).toBeInTheDocument();
```
emmenko:
```suggestion
  expect(getByText('Foo')).toBeInTheDocument();
  expect(queryByText('Foo')).not.toBeInTheDocument();
  expect(queryByText('Foo')).toBeFalsy();
```

lib/rules/no-get-by-for-asserting-element-not-present.js
```diff
@@ -0,0 +1,78 @@
...
+    docs: {
+      category: 'Best Practices',
+      description:
+        'Disallow the use of `getBy*` queries when checking elements are not present',
+      recommended: 'error',
+      url: getDocsUrl('no-get-by-for-asserting-element-not-present'),
+    },
+    messages: {
+      expectQueryBy:
+        'Use `expect(getBy*)` only when elements are present, otherwise use `expect(queryBy*)`.',
```
emmenko:
```suggestion
        'Use `getBy*` only when elements are present, otherwise use `queryBy*` when checking that elements are not present.',
```

thomlom: Is `when cheking that elements are not present` necessary? I feel like it's too much, what about: 

`Use getBy* only when checking elements are present, otherwise use queryBy*`

Belco90:
Awesome. Tomorrow morning I'll create the PR for `await-async-utils` so maybe we can wait these 2 together to release v2?

PR #67, opened 2020-01-22
docs: fix contributors badge count

Belco90:
Contributors badge count is not properly updated. Applying [this change mentioned in their docs](https://allcontributors.org/docs/en/bot/faq#all-contributors-badge-count-does-not-update) should fix it.


Belco90:
Now other badges are completely broken and I don't know why 😰
image

PR #69, opened 2020-01-24
feat: new rule await-async-utils

### LINKED ISSUE #49: Enforce await on async utilities
Currently https://github.com/Belco90/eslint-plugin-testing-library/blob/master/docs/rules/await-async-query.md rule only supports findBy* and findByAll*. What do you think about enabling it for other async utilities, too? 

https://testing-library.com/docs/dom-testing-library/api-async

```
const ASYNC_QUERIES_REGEXP = /^find(All)?By(LabelText|PlaceholderText|Text|AltText|Title|DisplayValue|Role|TestId)$/;
```

Belco90:
Closes #49

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-4-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-4-orange.svg?style=flat-square)](#contributors-)
 
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -24,7 +24,7 @@
 
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-4-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-4-orange.svg?style=flat-square)](#contributors-)
```
Belco90: I removed this by mistake in that PR trying to fix the badges.

lib/rules/await-async-utils.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
```
Belco90: This rule implementation is almost the same as `await-async-query` but I didn't try to optimize this as we may migrate this to TS soon.

lib/rules/await-async-utils.js
```diff
@@ -0,0 +1,103 @@
...
+
+  create: function(context) {
+    const testingLibraryUtilUsage = [];
+    return {
+      [`CallExpression > Identifier[name=${ASYNC_UTILS_REGEXP}]`](node) {
+        if (!isAwaited(node.parent.parent) && !isPromiseResolved(node)) {
+          testingLibraryUtilUsage.push(node);
+        }
+      },
+      'Program:exit'() {
```
emmenko: Ah interesting, I didn’t know about this

lib/rules/await-async-utils.js
```diff
@@ -0,0 +1,103 @@
...
+};
+
+function isAwaited(node) {
+  return VALID_PARENTS.includes(node.type);
+}
+
+function isPromiseResolved(node) {
+  const parent = node.parent;
+
+  const hasAThenProperty = node =>
```
emmenko: Does this need to be declared inside the `isPromiseResolved` function? Just to avoid confusion with arg names.

lib/rules/await-async-utils.js
Belco90: Me either! I learnt about it thanks to @thomlom, you can find more info [here](https://eslint.org/docs/developer-guide/selectors#listening-for-selectors-in-rules)

lib/rules/await-async-utils.js
Belco90: Not really, let me move this outside.

thomlom:
Quick question: why not merging `await-async-query` with `await-async-utils` instead of maintaining these two rules?

Belco90:
> Quick question: why not merging `await-async-query` with `await-async-utils` instead of maintaining these two rules?

Well, even if the implementation looks similar (but it's not the same 100%), the async stuff being checked is different so as we already had a rule for async queries, we would have different rule for async utils to 1) being able to config what you want to lint and 2) not mix different concerns (ones are queries to get elements, others are utils to wait until something happen).

I think mixing both in the same rule would be a mistake because 1) the name of the rule wouldn't make sense with what is checking and 2) we could rename the rule and include everything but then every single async function from testing library should go under that rule and it seems too much.

Belco90:
Merging as @thomlom approved outside github (he's busy dealing with personal stuff) and @emmenko approved before my last little change.

PR #72, opened 2020-01-29
feat: add no-manual-cleanup rule

### LINKED ISSUE #63: New rule: Disallow manual cleanups
The `cleanup` util to unmount React trees mounted with `render` will be call automatically if:

- [testing framework you are using supports global `afterEach`](https://testing-library.com/docs/react-testing-library/api#cleanup)
- [you don't skip the auto cleanup with RTL_SKIP_AUTO_CLEANUP](https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup)

So in order to avoid adding manual cleanups when is automatically setup, we can add a rule to prevent this. It looks like an easy rule to implement that could be named `no-manual-cleanups` that should not be enabled in any shareable config by default.

thomlom:
Closes #63


README.md
```diff
@@ -142,6 +144,7 @@ To enable this configuration use the `extends` property in your
 | [no-debug](docs/rules/no-debug.md)                                                                  | Disallow the use of `debug`                                                 | ![angular-badge][] ![react-badge][] ![vue-badge][]                        |                    |
 | [no-dom-import](docs/rules/no-dom-import.md)                                                        | Disallow importing from DOM Testing Library                                 | ![angular-badge][] ![react-badge][] ![vue-badge][]                        | ![fixable-badge][] |
 | [no-get-by-for-checking-element-not-present](docs/rules/no-get-by-for-checking-element-not-present) | Disallow the use of `getBy*` queries when checking elements are not present |                                                                           |                    |
+| [no-manual-cleanup](docs/rules/no-manual-cleanup.md)                                                | Disallow the use of `cleanup`                                               | ![react-badge][] ![vue-badge][]                                           |                    |
```
Belco90: I don't think we should enable this rule for any shareable config for reasons explained in the original issue.

lib/rules/no-manual-cleanup.js
```diff
@@ -0,0 +1,52 @@
...
+    type: 'problem',
+    docs: {
+      description: ' Disallow the use of `cleanup`',
+      category: 'Best Practices',
+      recommended: false,
+      url: getDocsUrl('no-manual-cleanup'),
+    },
+    messages: {
+      noManualCleanup:
+        "`cleanup` is performed automatically after each test by {{library}}, you don't need manual cleanups.",
```
Belco90: This is not right: the automatic cleanup is performed depending on your testing framework, not on your testing library package. This rule should just mention that manual cleanups are dissalowed as the testing framework taking care of it automatically.

thomlom: We are both not right. While it's true the automatic cleanup is performed depending on the testing framework, not all testing libraries package have a `cleanup` method, like Angular's testing library. 

It does not make sense for me to make this rule fire even if the package doesn't have a `cleanup` method.

I agree this rule should be removed from shareable configs as we don't know the test runner used though.

Belco90: Fair enough about short-circuiting when importing from a testing library package without `cleanup`, but still the error message is wrong as the one handling the automatic cleanup is the testing framework (e.g. jest), not testing library package (e.g. `@testing-library/angular`).

(Damn, "testing library" and "testing framework" are not making easier this discussion).

thomlom: Sure, I'll fix it too, thanks for the feedback! 👍

By the way, there are some rules like `no-debug` where we don't support frameworks like Svelte Testing Library, we should probably open a new issue to include them.

Belco90: That's actually one of the reasons I don't like to write code for checking specific testing library packages as if a new one comes up tomorrow we could have to update rules. I prefer to leave them agnostic from any package, specially those not enabled by default, so users can set them up under their responsibility. Of course, we should mention in proper rule doc which are those testing library packages supposed to be compatible with the rule for clarification.

thomlom: What I meant is that we have shareable configs for Angular, Vue, etc. But we are not adding any support for Svelte, Preact, etc. I think we should extend the rules that already work with specific packages (such as no-debug, or no-dom-import) to other potential testing libraries 🙂

Belco90: Ah of course 🤦‍♂️

And probably we should open an issue about including more frameworks for sure!

docs/rules/no-manual-cleanup.md
```diff
@@ -0,0 +1,40 @@
...
+```js
+import { cleanup } from 'any-other-library';
+
+const utils = require('any-other-library');
+utils.cleanup();
+```
+
+## Further Reading
+
+- [cleanup API in React Testing Library](https://testing-library.com/docs/react-testing-library/api#cleanup)
```
Belco90: Can you add skipping auto cleanup here:
[Skipping Auto Cleanup](https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup)

PR #74, opened 2020-02-01
Wrong link to docs in README

arvigeus:
Missing .md extension


arvigeus:
Another one I found: [prefer-explicit-assert](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/prefer-explicit-assert.md) suggests using it in conjunction with prefer-expect-query-by, which seems not existing.

Belco90:
> Another one I found: [prefer-explicit-assert](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/prefer-explicit-assert.md) suggests using it in conjunction with prefer-expect-query-by, which seems not existing.

Ah we forgot to remove that. Can you address that change too? Thanks!

Belco90:
We can address the suggestion for non-existing rule in a different PR.

@all-contributors please add @arvigeus for docs

arvigeus:
I'll remove the rule later today. Thanks!

Belco90:
Don't worry! Addressed already in #78 

Thanks for catching this!

PR #79, opened 2020-02-04
fix: support async render

### LINKED ISSUE #76: `debug` doesn't seem to work when wrapped in a test case
I just encountered a case where `debug` doesn't seem to work.

```js
describe("Home", () => {
  test("displays no posts if there are no posts", async () => {
    const { debug } = await render(<Home />);
    debug();
  });
})
```

Apparently, the `render` method doesn't seem to be detected when it's wrapped in a test case. Moving the `render` method outside the test case seems to trigger the `no-debug` rule.

I'm going to see how to fix it as soon as I can. But if anyone feels like contributing, feel free to fix it 🙂 

By the way, it's not the first time something doesn't work because some code is wrapped in a test case. I really think we should provide more real-world examples to all of our rules.

**UPDATE**: it turns out the bug is caused by my `await` statement.

thomlom:
Closes #76

### DIFF
--- a/lib/rules/no-debug.js
+++ b/lib/rules/no-debug.js
@@ -144,15 +144,21 @@ module.exports = {
   },
 };
 
-function isRenderVariableDeclarator(node, renderFunctions) {
-  const isRenderFunction = callNode =>
-    ['render', ...renderFunctions].some(name => name === callNode.callee.name);
+function isRenderFunction(callNode, renderFunctions) {
+  return ['render', ...renderFunctions].some(
+    name => name === callNode.callee.name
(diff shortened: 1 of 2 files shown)


lib/rules/no-debug.js
```diff
@@ -149,3 +143,18 @@ module.exports = {
     };
   },
 };
+
+function isRenderVariableDeclarator(node, renderFunctions) {
+  const isRenderFunction = callNode =>
```
Belco90: Can you declare this one outside of `isRenderVariableDeclarator`?

PR #81, opened 2020-02-10
fix(no-debug): ensure callee exists

marudor:
The code used in the tests currently results in an exception as callee does not exist.
This ensures it's there.


Belco90:
@all-contributors please add @marudor for code and test

Issue #85, opened 2020-03-13
Update await-async-utils with changes from dom-testing-library v7


Belco90:
After release of [dom-testing-library v7](https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0) there are couple of things to be updated for `await-async-utils` rule:

- include `waitFor` as valid async util
- update the doc to mention `wait`, `waitForElement` and `waitForDomChange` are now deprecated
- make sure when `waitForElementToBeRemoved` receives an element rather than callback doesn't cause false positives

Belco90:
This will be released on v3

PR #88, opened 2020-03-14
feat(prefer-wait-for): new rule prefer-wait-for

### LINKED ISSUE #86: RFC: fixer for `wait` (V7)
Would it be an option to add a fixable rule for the changes made to `wait`?
The implementation could look like [this](https://astexplorer.net/#/gist/52cd2d7008f89a1baa11819d361f3bb5/22406c2aef432778000274ff6178e0b11ba7b543)

```js
export default function(context) {
  return {
    "CallExpression > Identifier[name=wait]"(node) {
      context.report({
        node: node,
        message: "waitForRename",
        fix: fixer => fixer.replaceText(node, "waitFor")
      });
    },
    "CallExpression[callee.name=waitFor][arguments.length=0]"(node) {
      context.report({
        node: node,
        message: "emptyCallBack",
        fix: fixer => fixer.replaceText(node, "waitFor(() => {})")
      });
    }
  };
}
```

input:

```js
async () => {
	await wait(() => {});
  	await wait();
}
```

output:

```js
async () => {
	await waitFor(() => {});
  	await waitFor(() => {});
}
```

Belco90:
This closes #86 

@timdeschryver let me know what you think! I had to include the empty callback fixer within the the other selectors fixers as otherwise tests were not passing. It's weird tho, the code was executed but apparently not taken into account for the test result.


benmonro:
should the docs for the other rules related to async utils be updated as well?

Belco90:
@benmonro that will be addressed here https://github.com/testing-library/eslint-plugin-testing-library/issues/85

docs/rules/prefer-wait-for.md
```diff
@@ -0,0 +1,48 @@
+# Use `waitFor` instead of deprecated wait methods (prefer-wait-for)
+
+Since dom-testing-library v7 new `waitFor` async util has been added. This new method satisfies the use cases of `wait`, `waitForElement`, and `waitForDomChange`, so those have been deprecated.
```
thomlom:
```suggestion
`dom-testing-library` v7 released a new async util called `waitFor` which satisfies the use cases of `wait`, `waitForElement`, and `waitForDomChange` making them deprecated.
```

docs/rules/prefer-wait-for.md
```diff
@@ -0,0 +1,48 @@
...
+
+Deprecated wait async utils are:
+
+- `wait`
+- `waitForElement`
+- `waitForDomChange`
+
+> This rule will auto fix deprecated async utils for you, including necessary empty callback for `waitFor`. This means `wait();` will be replaced with `waitFor(() => {});`
+
+Examples of **incorrect** code for this rule:
```
thomlom:
```suggestion
Examples of **correct** code for this rule:
```

docs/rules/prefer-wait-for.md
```diff
@@ -0,0 +1,48 @@
...
+const foo = async () => {
+  // new waitFor method
+  await waitFor(() => {});
+
+  // previous waitForElementToBeRemoved is not deprecated
+  await waitForElementToBeRemoved(() => {});
+};
+```
+
+Examples of **correct** code for this rule:
```
thomlom:
```suggestion
Examples of **incorrect** code for this rule:
```

lib/rules/prefer-wait-for.js
```diff
@@ -0,0 +1,74 @@
...
+        node: node,
+        messageId: 'preferWaitFor',
+        data: {
+          methodName: node.name,
+        },
+        fix: fixer => {
+          const { parent } = node;
+          const [arg] = parent.arguments;
+
+          if (arg) {
```
thomlom: You could use a ternary here: 

```js
return arg ? fixer.replaceText(node, 'waitFor') : fixer.replaceText(parent, 'waitFor(() => {})');
```

lib/rules/prefer-wait-for.js
```diff
@@ -0,0 +1,74 @@
...
+              fixer.insertTextBefore(arg, `() => {}, `),
+            ];
+          }
+          return fixer.replaceText(parent, 'waitFor(() => {})');
+        },
+      });
+    };
+
+    return {
+      'CallExpression > Identifier[name=wait]'(node) {
```
thomlom: What if the user is using a custom `wait` util? The rule will report it even if it's not imported from Testing Library?

Belco90: Yep. I think it's too much from where it was imported? But I don't know what to do to be honest.

Belco90: Obviously it would be better to check where it's imported from, but we have been checking this in other rules without caring about it. Could be an improvement for a separated ticket for all the rules within the plugin?

Belco90: On the other hand... It would be interesting to check from where it's imported so we can also replace the import itself. Otherwise, the fix of this rule could leave a broken file.

docs/rules/prefer-wait-for.md
```diff
@@ -0,0 +1,48 @@
+# Use `waitFor` instead of deprecated wait methods (prefer-wait-for)
+
+`dom-testing-library` v7 released a new async util called `waitFor` which satisfies the use cases of `wait`, `waitForElement`, and `waitForDomChange` making them deprecated.
+
+## Rule Details
+
+This rule aims to use `waitFor` async util rather than previous deprecated ones.
+
+Deprecated wait async utils are:
```
timdeschryver:
```suggestion
Deprecated `wait` async utils are:
```

docs/rules/prefer-wait-for.md
```diff
@@ -0,0 +1,48 @@
...
+Examples of **incorrect** code for this rule:
+
+```js
+const foo = async () => {
+  await wait();
+  await wait(() => {});
+  await waitForElement(() => {});
+  await waitForDomChange();
+  await waitForDomChange(mutationObserverOptions);
+  await waitForDomChange({ options: true });
```
timdeschryver: I think it would be better to use a "real" config here 

```suggestion
  await waitForDomChange({ timeout: 100});
```

docs/rules/prefer-wait-for.md
```diff
@@ -0,0 +1,48 @@
...
+
+This rule aims to use `waitFor` async util rather than previous deprecated ones.
+
+Deprecated wait async utils are:
+
+- `wait`
+- `waitForElement`
+- `waitForDomChange`
+
+> This rule will auto fix deprecated async utils for you, including necessary empty callback for `waitFor`. This means `wait();` will be replaced with `waitFor(() => {});`
```
timdeschryver:
```suggestion
> This rule will auto fix deprecated async utils for you, including the necessary empty callback for `waitFor`. This means `wait();` will be replaced with `waitFor(() => {});`
```

timdeschryver:
Having the empty callback with the rest makes sense 👍

tests/lib/rules/prefer-wait-for.js
```diff
@@ -0,0 +1,167 @@
...
+        await waitForElementToBeRemoved(() => {});
+      }`,
+    },
+    {
+      code: `async () => {
+        await waitForSomethingElse(() => {});
+      }`,
+    },
+    {
+      code: `import { wait } from 'some-testing-library'`,
```
timdeschryver: I just thought of this, but shouldn't these be renamed as well? 
Otherwise it wont compile, right? 🤔

Belco90: Yes, that's what I was discussing with @thomlom in a previous comment. We would need to check from where it's imported to being able to 1) check if it's the `wait` we are expecting and 2) rename the import too. Haven't done this before, so I'm checking how to do it properly.

timdeschryver: We could use something similar as https://github.com/cartant/eslint-plugin-rxjs/blob/master/source/rules/just.ts

Belco90: I'll give it a try later to see if I can mix that with the current rule code. Thanks.

timdeschryver: I'm also writing something to see how we could do it

timdeschryver: @Belco90 what do you think of this?
It will also take care of https://github.com/testing-library/eslint-plugin-testing-library/pull/88#discussion_r392661245, and only rename testing-library occurrences 

https://astexplorer.net/#/gist/6994663f6be1b74b6b6fb41b14320790/latest

Belco90: Ok, I'll try to put all this together with the tests to see if this works as expected. Thanks!

timdeschryver: 👍 feel free to ping me for some help or to do another review

Belco90: Thanks! I've been little bit busy and only had time to prepare the tests, but should have some spare time today to check everything together. I'll let you know if need some help.

timdeschryver: There's no rush :)

lib/rules/prefer-wait-for.js
emmenko: I would suggest keep it like this for now and wait for users to ask for it, in case it's a valid use case.

docs/rules/prefer-wait-for.md
```diff
@@ -0,0 +1,48 @@
...
+
+This rule aims to use `waitFor` async util rather than previous deprecated ones.
+
+Deprecated `wait` async utils are:
+
+- `wait`
+- `waitForElement`
+- `waitForDomChange`
+
+> This rule will auto fix deprecated async utils for you, including the necessary empty callback for `waitFor`. This means `wait();` will be replaced with `waitFor(() => {});`
```
emmenko: 💯

lib/rules/prefer-wait-for.js
emmenko: Unless it's not too complicated to implement of course 😅

tests/lib/rules/prefer-wait-for.js
```diff
@@ -0,0 +1,167 @@
...
+    },
+    {
+      code: `import { waitForElement } from 'some-testing-library'`,
+    },
+    {
+      code: `import { waitForDomChange } from 'some-testing-library'`,
```
emmenko: Can you explain what the purpose of testing only the import is?

Belco90: This was just a test I came up with to check only calls to those methods were transformed to `waitFor` but then we realized that we need to transform the imports too. I got the tests updated and suggested code from @timdeschryver almost in place. I'll push it soon for another review round.

emmenko: Ah ok, thanks!

tests/lib/rules/prefer-wait-for.js
Belco90: It was a tough one but I got it! Please check the final rule implementation. @timdeschryver it was inspired by your code but I had to implement the import reporter in a different way as it wasn't working properly together with everything else. Tests passing now (as pipeline not run here because doesn't match branching rules):

image

Belco90: I found one case not covered: multiline imports. I'm on it :)

lib/rules/prefer-wait-for.js
```diff
@@ -0,0 +1,129 @@
...
+      context.report({
+        node: node,
+        messageId: 'preferWaitForImport',
+        fix(fixer) {
+          const fixers = [];
+          let waitForImported = false;
+
+          node.specifiers.forEach((specifier, idx, specifiers) => {
+            if (DEPRECATED_METHODS.includes(specifier.imported.name)) {
+              if (waitForImported) {
```
timdeschryver: Should we take into account that `waitFor` is already imported (manually), before running this rule?

Belco90: Yeah, that's a good point. It's easier to check here, I'll add that later!

tests/lib/rules/prefer-wait-for.js
Belco90: Now it's fixed, all import cases covered 💪

lib/rules/prefer-wait-for.js
Belco90: Solved with the new import fix code

PR #89, opened 2020-03-15
feat(await-async-utils): reflect waitFor changes

### LINKED ISSUE #85: Update await-async-utils with changes from dom-testing-library v7

Belco90:
This closes #85

### DIFF
--- a/docs/rules/await-async-utils.md
+++ b/docs/rules/await-async-utils.md
@@ -6,10 +6,11 @@ Ensure that promises returned by async utils are handled properly.
 
 Testing library provides several utilities for dealing with asynchronous code. These are useful to wait for an element until certain criteria or situation happens. The available async utils are:
 
-- `wait`
-- `waitForElement`
-- `waitForDomChange`
+- `waitFor` _(introduced in dom-testing-library v7)_
 - `waitForElementToBeRemoved`
+- `wait` _(**deprecated** in dom-testing-library v7)_
+- `waitForElement` _(**deprecated** in dom-testing-library v7)_
+- `waitForDomChange` _(**deprecated** in dom-testing-library v7)_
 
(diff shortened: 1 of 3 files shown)


docs/rules/await-async-utils.md
```diff
@@ -6,10 +6,11 @@ Ensure that promises returned by async utils are handled properly.
 
 Testing library provides several utilities for dealing with asynchronous code. These are useful to wait for an element until certain criteria or situation happens. The available async utils are:
 
-- `wait`
-- `waitForElement`
-- `waitForDomChange`
+- `waitFor` _(introduced in dom-testing-library v7)_
```
emmenko: 💯

docs/rules/await-async-utils.md
```diff
@@ -54,16 +52,13 @@ test('something correctly', async () => {
   );
 
   // `then` chained method is correct
-  waitForDomChange(() => {
-    return { container };
-  })
+  waitFor(() => {}, { timeout: 100 })
```
thomlom: Why adding a `timeout` to the example? 🤔

Belco90: It was just a stupid config object. I updated it on tests, I can do the same here.

Belco90: No, sorry. It's actually fine like this. This is a real example from new [`waitFor` optional second argument](https://testing-library.com/docs/dom-testing-library/api-async#waitfor).

PR #94, opened 2020-03-20
feat(no-wait-for-empty-callback ): new rule no-wait-for-empty-callback 

### LINKED ISSUE #92: New rule: no-wait-for-empty-cb
New `waitFor` method added in [dom-testing-library v7](https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0) introduces a new restriction on empty async utils: a callback must be passed as param. Empty callbacks are allowed but not recommended. Quoting changelog itself:

> it's recommended to avoid an empty callback and instead insert an assertion in that callback function. This is because otherwise your test is relying on the "next tick of the event loop" before proceeding, and that's not consistent with the philosophy of the library:

Because of that, a new rule called `no-wait-for-empty-cb` could enforce to pass non-empty callbacks to `waitFor`.

What would be considered a non-empty callback tho? I would say any of these:

- empty arrow function (i.e. `() => {}`)
- empty regular function (i.e. `function () {}`)
- a var referencing any of previously mentioned
- a function or var called `noop`?
- falsy values?

timdeschryver:
Closes #92 

This rule covers:

- empty arrow function (i.e. `() => {}`)
- empty regular function (i.e. `function () {}`)
- a function or var called `noop`

Inspired by eslint no-empty-function rule

### DIFF
--- a/tests/lib/rules/no-wait-for-empty-callback.js
+++ b/tests/lib/rules/no-wait-for-empty-callback.js
@@ -17,24 +17,23 @@ ruleTester.run('no-wait-for-empty-callback', rule, {
       code: `${m}(() => {
           screen.getByText(/submit/i)
         })`,
-      errors: [
-        {
-          messageId: 'noWaitForEmptyCallback',
-        },
-      ],
     })),
     ...ALL_WAIT_METHODS.map(m => ({
       code: `${m}(function() {
           screen.getByText(/submit/i)
         })`,
-      errors: [
-        {
-          messageId: 'noWaitForEmptyCallback',
-        },
-      ],
     })),
     {
+      code: `waitForElementToBeRemoved(someNode)`,
+    },
+    {
+      code: `waitForElementToBeRemoved(() => someNode)`,
+    },
+    {
       code: `waitSomethingElse(() => {})`,
+    },
+    {
+      code: `wait(() => {})`,
     },
   ],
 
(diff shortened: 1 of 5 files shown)


timdeschryver:
@Belco90 this is what we're aiming for, right?
Not sure how we can tackle `a var referencing any of previously mentioned`

Belco90:
Yes! This is exactly what we need. Let's leave the var referencing for future improvement if necessary. I'll add couple of comments tho I know this is just a draft.

tests/lib/rules/no-wait-for-empty-callback.js
```diff
@@ -0,0 +1,69 @@
...
+  valid: [
+    ...ALL_WAIT_METHODS.map(m => ({
+      code: `${m}(() => {
+          screen.getByText(/submit/i)
+        })`,
+      errors: [
+        {
+          messageId: 'noWaitForEmptyCallback',
+        },
+      ],
```
Belco90: Errors block is not necessary for valid tests right?

tests/lib/rules/no-wait-for-empty-callback.js
```diff
@@ -0,0 +1,69 @@
...
+    })),
+    {
+      code: `waitSomethingElse(() => {})`,
+    },
+  ],
+
+  invalid: [
```
Belco90: Would be nice to add more complex invalid cases like:

- callbacks receiving args but with empty body (e.g. `(foo, bar) => {}`)
- callbacks with body containing just comments (e.g. `() =>  { /* I'm empty anyway */ }`)
-  wait with no callback. What's the rule doing on this case? I'm sure it won't report anything but could cause an error in our rule code? Better to test that.

All of them for both default and arrow functions.

tests/lib/rules/no-wait-for-empty-callback.js
timdeschryver: Right! Started with the invalid ones and copy pasted 😅
Will update this PR later this week

PR #96, opened 2020-03-21
ci(await-async-query): update node versions

Belco90:
BREAKING CHANGE: Drop support for node v8. Min version allowed is node
v10.12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,17 @@
 language: node_js
 
-node_js:
-  - 8
-  - 10
+env:
+  - FORCE_COLOR=true
 
-branches:
-  only: master
+node_js:
+  - 10.12
+  - 12
+  - node
 
 jobs:
   include:
     - stage: release
+      if: branch = master AND type != pull_request
       node_js: lts/*
       deploy:
         provider: script


.travis.yml
```diff
@@ -1,15 +1,17 @@
 language: node_js
 
-node_js:
-  - 8
-  - 10
+env:
+  - FORCE_COLOR=true
```
thomlom: What is `FORCE_COLOR` doing?

.travis.yml
```diff
@@ -1,15 +1,17 @@
...
 
-branches:
-  only: master
+node_js:
+  - 10.12
+  - 12
+  - node
```
thomlom: Why do you have to specify `node`? Isn't `10.12` and `12` enough?

.travis.yml
```diff
@@ -1,15 +1,17 @@
...
 
 jobs:
   include:
     - stage: release
+      if: branch = master AND type != pull_request
```
thomlom:
```suggestion
      if: branch = master AND type = push
```

What do you think?

.travis.yml
Belco90: To display proper colors in CI output.

.travis.yml
Belco90: `node` is last active version, which at the moment is v13. With this we cover same versions as ESLint: 10.12, 12 and 13

.travis.yml
Belco90: Mmm I'm not sure. There are other types which I'm not sure if we are interested to trigger or not. Ignoring pull requests is the usual behavior I've seen in other packages.

PR #97, opened 2020-03-22
Plugin v3

### LINKED ISSUE #93: Release v3
Recent update on dom-testing-library to v7 has originated few improvements for some existing rules and ideas for some new others. I'll wrap all those changes under release of plugin v3 for several reasons:

- One of them implies a breaking change so I think it's better to put everything under same major version rather than be releasing different minors and one major at some point.
- Most of them are related to new `waitFor` async util so I'd like to test all changes together before releasing them to check they make sense as a whole.
- Changes are closed to be finished so this won't take too long.

Changes under v3 will be:

- [x] **BREAKING CHANGE**: drop node v8 support #96 
- [x] **BREAKING CHANGE**: replace rule `no-get-by-for-checking-element-not-present` with `prefer-presence-queries` #90 
- [x] Feature: update rule `await-async-utils` with new `waitFor` async util #89 
- [x] Feature: new rule `prefer-wait-for` #88
- [x] Feature: new rule `no-wait-for-empty-callback` #92 
- [x] Feature: new rule `prefer-screen-queries` #95 (_optional, only if it ready when releasing v3. Otherwise it will go on v3.1_)

All PRs for mentioned changes will point to branch `v3` where everything will be tested together before releasing everything. I'll test this myself in a huge codebase with more than 3k tests, but I'd like to get other people running this on their projects too.

Thoughts?

Belco90:
This closes #93


docs/rules/no-wait-for-empty-callback.md
```diff
@@ -0,0 +1,40 @@
+# Empty callbacks inside `waitFor` and `waitForElementToBeRemoved` are not preferred (no-wait-for-empty-callback)
+
+## Rule Details
+
+This rule aims to ensure the correct usage of `waitFor` and `waitForElementToBeRemoved`, in the way that they're intended to be used.
+If an empty callback is passed, these methods will just wait next tick of the event loop before proceeding, and that's not consistent with the philosophy of the library.
+**Instead, insert an assertion in that callback function.**
```
Belco90: @timdeschryver I've updated the rule details with proper suggestion fix from dom-testing-library doc. I'm even thinking about [providing a suggestion](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) for this rule!

Belco90: Mmm it needs to provide a suggestion fix too, not just the description, so we can't use that here.

timdeschryver: @Belco90 thanks and TIL eslint suggestions.
So if I understand correctly we can't provide a suggestion without a fix?

Belco90: I'm afraid so:

> The suggestion objects represent individual suggestions that could be applied and require either a desc key string that describes what applying the suggestion would do or a messageId key (see below), and a fix key that is a function defining the suggestion result.

Belco90:
v3 of the plugin tested in a large codebase with more +3k tests. Everything looks fine, the `prefer-wait-for` fix is 💯

timdeschryver:
🥳 Awesome to hear that ✋

Belco90:
@all-contributors please add @timdeschryver for code, doc, ideas, review and test

PR #98, opened 2020-03-22
feat: new rule prefer-presence-queries

### LINKED ISSUE #90: Feature request: convert no-get-by-for-checking-element-not-present to be prefer-appropriate-query-for-expect
```jsx
// reports:
expect(getByText("Foo")).not.toBeInTheDocument() // as it already does
expect(queryByText("Foo")).toBeInTheDocument()

// does not report:
expect(queryByText("Foo")).not.toBeInTheDocument()
expect(getByText("Foo")).toBeInTheDocument()
```
In addition as discussed below in the comments, the use of `get*` queries inside of waitForElementToBeRemoved will no longer report an error either in a callback or directly as a parameter.

```js
  //does not report:
  await waitForElementToBeRemoved(screen.getByTitle("Loading..."));
  await waitForElementToBeRemoved(() => screen.getByTitle("Loading..."));
```

Belco90:
This closes #90 

BREAKING CHANGE: previous rule `no-get-by-for-checking-element-not-present` has been replaced by this one.


docs/rules/prefer-presence-queries.md
```diff
@@ -0,0 +1,63 @@
+# Enforce specific queries when checking appearance or disappearance of elements (prefer-presence-queries)
+
+The (DOM) Testing Library allows to query DOM elements using different types of queries such as `getBy*` and `queryBy*`. Using `getBy*` throws an error in case the element is not found, while `queryBy` returns null instead of throwing. These differences are useful in some situations:
+
+- using `getBy*` queries when asserting if element is present, so if the element is not found the error thrown will offer better info than asserting with other queries which not throw.
```
benmonro:
```suggestion
- using `getBy*` queries when asserting if element is present, so if the element is not found the error thrown will offer better info than asserting with other queries which will not throw an error.
```

docs/rules/prefer-presence-queries.md
```diff
@@ -0,0 +1,63 @@
...
+- using `queryBy*` queries when asserting if element is not present, so the test doesn't fail immediately when the element it's not found and the assertion can be executed properly.
```
benmonro:
```suggestion
- using `queryBy*` queries when asserting if element is not present, so the test doesn't fail immediately when the element is not found and the assertion can be executed properly.
```

docs/rules/prefer-presence-queries.md
```diff
@@ -0,0 +1,63 @@
...
+```js
+test('some test', () => {
+  render(<App />);
+
+  // check element is present with `queryBy*`
+  expect(screen.queryByText('button')).toBeInTheDocument();
+  expect(screen.queryByText('button')).toBeTruthy();
+  expect(screen.queryByText('button')).toBeNull();
+  expect(screen.queryByText('button')).not.toBeNull();
+  expect(screen.queryByText('button')).not.toBeFalsy();
```
benmonro: what about the following:

```jsx
const button = screen.queryByText("button");
expect(button).toBeInTheDocument()

// or this one

expect(screen.queryAllByText("button")[1]).toBeInTheDocument()
```

docs/rules/prefer-presence-queries.md
```diff
@@ -0,0 +1,63 @@
...
+
+  // check element is NOT present with `getBy*`
+  expect(screen.getByText('loading')).not.toBeInTheDocument();
+  expect(screen.getByText('loading')).not.toBeTruthy();
+  expect(screen.getByText('loading')).not.toBeNull();
+  expect(screen.getByText('loading')).toBeNull();
+  expect(screen.getByText('loading')).toBeFalsy();
```
benmonro: i'm fine w/ these but less concerned about them since the tests will actually fail if they are present

lib/rules/prefer-presence-queries.js
```diff
@@ -0,0 +1,91 @@
...
+
+const PRESENCE_MATCHERS = ['toBeInTheDocument', 'toBeTruthy', 'toBeDefined'];
+const ABSENCE_MATCHERS = ['toBeNull', 'toBeFalsy'];
+
+module.exports = {
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description:
+        'Disallow the use of `getBy*` queries when checking elements are not present',
```
benmonro: I would word this to be more generic
```suggestion
        'Ensure appropriate get*/query* queries are used with their respective matchers',
```
or something to that effect

docs/rules/prefer-presence-queries.md
```diff
@@ -0,0 +1,63 @@
...
+
+  // check element is NOT present with `queryBy*`
+  expect(screen.queryByText('loading')).not.toBeInTheDocument();
+  expect(screen.queryByText('loading')).not.toBeTruthy();
+  expect(screen.queryByText('loading')).not.toBeNull();
+  expect(screen.queryByText('loading')).toBeNull();
+  expect(screen.queryByText('loading')).toBeFalsy();
+});
+```
+
```
benmonro: might also be worth making a note that the rule doesn't apply to findBy queries

docs/rules/prefer-presence-queries.md
Belco90: getAllBy and querAllBy are not checked by the rule.

benmonro: it should be possible though right? at least for queryAllBy

docs/rules/prefer-presence-queries.md
Belco90: I think it's clear it only apply to `getBy*` and `queryBy*` queries within rule details.

docs/rules/prefer-presence-queries.md
Belco90: Yes, it's possible. I'll give it a try later and apply typos suggestions.

benmonro: cool and thanks again!

docs/rules/prefer-presence-queries.md
Belco90: `findBy*` example added to correct examples

docs/rules/prefer-presence-queries.md
timdeschryver: I'm unsure if these must be included in this rule.

Newcomers to TL, might not know that these queries throw; by writing tests like this it makes these explicit and easier to understand.

I've seen `getBy` queries get removed because they didn't know they fail underneath.

timdeschryver: (also, not against it) just wanted to share my experience :)

benmonro: @timdeschryver yeah i was on the fence initially as the tests themselves would catch that, but i figured A) they're not hurting B) when used w/ the eslint plugin for IDEs it can be a quick check before you run the test...

lib/rules/prefer-presence-queries.js
```diff
@@ -0,0 +1,94 @@
+'use strict';
+
+const { getDocsUrl, ALL_QUERIES_METHODS } = require('../utils');
+
+const QUERIES_REGEXP = new RegExp(
+  `^(get|query)(All)?(${ALL_QUERIES_METHODS.join('|')})$`
```
timdeschryver: what about `find`? https://testing-library.com/docs/dom-testing-library/api-queries#findby

lib/rules/prefer-presence-queries.js
```diff
@@ -0,0 +1,94 @@
...
+
+      if (expectCallNode) {
+        const expectStatement = expectCallNode.parent;
+        let matcher = expectStatement.property.name;
+        let isNegatedMatcher = false;
+
+        if (matcher === 'not') {
+          isNegatedMatcher = true;
+          matcher =
+            expectStatement.parent &&
```
timdeschryver: I think this can removed? On line 39, there is the same assignment.
If these properties can be `undefined` perhaps it would be safer to rewrite line 39, the same way like this assignment.

Belco90: It's not the same assignment tho. On line 39 is getting property name of expect statement. Here is getting property name from **parent** of expect statement. This would be clearer if getting the matcher with some util similar to `findClosestCallNode`.

You are right about guarding against nil on both cases.

lib/rules/prefer-presence-queries.js
Belco90: I guess it should be added but wasn't mentioned on original discussion.

docs/rules/prefer-presence-queries.md
Belco90: I don't know if I'm following. The problem is reporting `getBy*` for asserting element absence? Or reporting `toBeFalsy` matcher?

@timdeschryver I'd recommend go through the related issue for original discussion.

Belco90: And this rule won't be enabled by default, so it's not that intrusive imho.

lib/rules/prefer-presence-queries.js
timdeschryver: Oh shoot you're right.
I was so sure because I read it twice, sorry for the confussion 😅

You could create a util function, but I'm not sure if it would add the value as `findClosestCallNode` does

docs/rules/prefer-presence-queries.md
timdeschryver: Again, sorry for the confusion - these does add value.
The sentence "I'm unsure if these must be included in this rule." was a leftover...

lib/rules/prefer-presence-queries.js
Belco90: Thoughts about this: I don't think findBy should be included in this rule as it's the only async query so there is no alternative async query for it. When checking appearance you can use findBy and when checking disappearance you can use findBy too (wrapped with waitForElementToBeRemoved) so I think the rule is fine with only getBy and queryBy.

This said, I'll make sure I've used the appropriate language for describing the rule so it always refers to element presence/absence (sync, checked with getBy/queryBy) and never to element appearing/disappearing (async, checked with findBy).

timdeschryver: This makes sense, thanks 👍

Belco90: I hope I can update the doc accordingly and add the last check for var declarations tomorrow.

docs/rules/prefer-presence-queries.md
benmonro: @Belco90 do you think the first case i mentioned is possible as well?

i.e. 
```jsx
const button = screen.queryByText("button");
expect(button).toBeInTheDocument()
```

Belco90: Yes, still on it. I’ll request your review again when added

Belco90:
@benmonro after thinking about it again, I don't think this case should be reported:

```
const button = screen.queryByText("button");
expect(button).toBeInTheDocument()
```

This is not so simple to identify as vars containing the dom elements are updated dynamically. Take a look at this example:
image

There is a really simple component with a text and a button. When you click the button, the text disappear. In the test (which you can see it's passing) I'm getting the message with `getByText` and saving it within a var to check it's present, this is fine. But then I'm checking with the same var that the element is **not** present after clicking the button. That element was get with a `getBy*` query and I'm checking that it's **not** present so this rule would complain from the case you mentioned above. But same var and query have been used to check is present in a different line. Should this rule complain then? If so, the user would have to switch the query to `queryBy*` one and then the rule would complain about the opposite: using a `queryBy*` for checking element is present.

This seems like a vicious circle, so I don't think this rule should check queries values saved in vars as it adds lots of complexity. This would be the final implementation for the rule then.

benmonro:
@Belco90 hmm yeah I see what you mean.  What if we just check for it's use immediately after the query?  Reason I ask is that it gives people a way to work around the rule which is what I'm trying to prevent...

Belco90:
> @Belco90 hmm yeah I see what you mean. What if we just check for it's use immediately after the query? Reason I ask is that it gives people a way to work around the rule which is what I'm trying to prevent...

I'm pretty sure there are others ways to skip this and other rules. There are different ways it could be solved, but I'm not sure is worth the effort for now. Even checking if it's used immediately after the query has lots of variants. Would you report this?

```
// they are not used immediately after their declaration
const button = screen.queryByText("button");
const submit = screen.queryByText('submit');
expect(button).toBeInTheDocument();
expect(submit).toBeInTheDocument();
```

Or this?
```
// not used immediately either
const button = screen.queryByText("button");
expect(button).toHaveClass('hidden');
expect(button).toBeInTheDocument();
```

As you can see, tons of variants would take a while to cover. This could be discussed after this first approach and release it as a minor release.

benmonro:
@Belco90 ok fair enough this is definitely better so yeah let's ship it and see what happens

PR #99, opened 2020-03-25
feat(prefer-screen-queries): add prefer-screen-queries

### LINKED ISSUE #95: RFC: new rule to prefer the usage of screen
It's preferred to use `screen` when using the query functions.

```js
// bad
const component = render()
component.getBy()

// good
screen.getBy()
```

For a first implementation we can verify that the queries used have `screen` as callee.
This doesn't take the following into account, but it makes the rule simpler.

```
const screen = render()
screen.getBy
```

timdeschryver:
Closes #95


lib/utils.js
```diff
@@ -44,8 +44,8 @@ const ASYNC_QUERIES_COMBINATIONS = combineQueries(
 );
 
 const ALL_QUERIES_COMBINATIONS = [
-  SYNC_QUERIES_COMBINATIONS,
-  ASYNC_QUERIES_COMBINATIONS,
+  ...SYNC_QUERIES_COMBINATIONS,
+  ...ASYNC_QUERIES_COMBINATIONS,
```
timdeschryver: I hope this is OK (`ALL_QUERIES_COMBINATIONS ` wasn't used until now)

docs/rules/prefer-screen-queries.md
```diff
@@ -0,0 +1,28 @@
...
+
+const utils = render(<Component />);
+utils.getByText('foo');
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import { screen } from '@testing-library/any-framework';
+
```
Belco90: Can you add here the render line to have a full example?

```suggestion
render(<Component />);
```

lib/utils.js
Belco90: Damn 🤦‍♂️ Thanks for this fix!

Belco90:
I edited to PR to point it to `v3` branch and now you get some conflicts. Sorry about that.

timdeschryver:
No problem! Should be good now, thanks for the review

PR #104, opened 2020-04-01
fix(no-debug): enable for angular

timdeschryver:
Angular also uses `debug` 😁


Belco90:
@all-contributors please add @timdeschryver for bug

PR #117, opened 2020-04-20
feat: add basic config for cypress

emmenko:
I noticed that some of the default rules (like for `async`) don't really work for Cypress, as [commands are not promises](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Commands-Are-Not-Promises).

So I thought it would be good to start a default config for Cypress as well.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -37,6 +37,12 @@ module.exports = {
         'testing-library/no-dom-import': ['error', 'angular'],
       },
     },
+    cypress: {
+      plugins: ['testing-library'],
+      rules: {
+        'testing-library/no-debug': 'warn',
+      },
+    },
     react: {
       plugins: ['testing-library'],
       rules: {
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -37,6 +37,12 @@ module.exports = {
         'testing-library/no-dom-import': ['error', 'angular'],
       },
     },
+    cypress: {
+      plugins: ['testing-library'],
+      rules: {
+        'testing-library/no-debug': 'warn',
+      },
```
emmenko, line 44: Notice here that we don't spread `recommendedRules`

Belco90: I guess we should update the comment in Frameworks section within README saying "Note that frameworks configurations enable their specific rules + recommended rules."

emmenko:
Ah ok, I missed that thread. No problem, let me close this PR then. There is no rush for this.

Thanks :)

Issue #122, opened 2020-05-04
False positives on `await-async-query` and `no-await-sync-query`


Belco90:
Related to #120, #114

Since we are checking testing library queries when calling queries member expressions for `await-async-query` and `no-await-sync-query`, we are getting some false positives in some cases:

- when using queries in cypress files
- when using methods with the same name in other places (e.g. `findByTestId` in enzyme tests in a codebase mixing both enzyme and testing library)

The main issue here is that we are checking the queries methods in every single file no matter where the method is coming from, but we should only check those imported from testing library or returned by something imported from testing library. However, I think a first approach could be at least check if testing library is imported in the file to check the methods or not.

afontcu:
Hi! Looks like #137 was merged but this issue is still valid, right? (Just to double check 😇)

Belco90:
Yes, it was only fixed for `no-await-sync-query` but I'm afraid the issue is still there for `await-async-query`.

kentcdodds:
I'm seeing this as well. Also `prefer-screen-queries` is firing on `cy.findByRole`... etc.

Belco90:
Yes, still pending to fix `await-async-query`.

About `prefer-screen-queries` one, that's new I think. There is [this workaround](https://github.com/testing-library/eslint-plugin-testing-library/issues/120#issuecomment-622345740) for another cypress related issue that might work in the meantime.

I hope I can find some time soon to fix those, specially the `await-async-query` one which is one of the first rules so it needs to be updated properly.

thomlom:
As of today, there are two rules that have similar implementations:

- [await-async-query](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/await-async-query.md)
- [await-async-utils](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/await-async-utils.md)

#137 seems to have fixed `await-async-utils` but didn't include the fix for `await-async-query`. I see two solutions to this: 

1. We merge the rules `await-async-utils` and `await-async-query` since they both check the usage of `await` for imported methods. This would result in a breaking change though.
2. We create a util/function that allows to check if an async imported method (whatever the name) has been used throughout the code. Then we make use of that util in these two rules.

WDYT?

Belco90:
1. We could achieve that in next v4 as is still pending.
2. We have now a module for common functionality named `node-utils.ts`. We can move there that shared code.

[...]

PR #137, opened 2020-05-16
fix(await-async-utils): false positives

### LINKED ISSUE #118: await-async-utils false positives
The await-async-utils rule is finding function calls that aren't actually from testing library. For example, if I define a function `wait` in my test code:

```js
function wait(delay) {
  return new Promise((resolve) => {
    setTimeout(resolve, delay);
  });
}
```

The lint rule will match it since it has the same name as the testing library's async util.

Would it make sense to first detect the import of testing-library, before applying the rule?

dmitry-lobanov:
Closes #118; Related #122

Fixes false positives `no-await-sync-query`

### DIFF
--- a/lib/rules/await-async-utils.ts
+++ b/lib/rules/await-async-utils.ts
@@ -1,6 +1,6 @@
 import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils';
 
-import { getDocsUrl, ASYNC_UTILS } from '../utils';
+import { getDocsUrl, ASYNC_UTILS, LIBRARY_MODULES } from '../utils';
 import { isCallExpression, hasThenProperty } from '../node-utils';
 
 export const RULE_NAME = 'await-async-utils';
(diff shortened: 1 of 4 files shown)


lib/rules/await-async-utils.ts
```diff
@@ -49,17 +49,49 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
   defaultOptions: [],
 
   create(context) {
-    const testingLibraryUtilUsage: TSESTree.Identifier[] = [];
+    const asyncUtilsUsage: Array<{ node: TSESTree.Identifier | TSESTree.MemberExpression, name: string }> = [];
+    const importedAsyncUtils: string[] = [];
+
     return {
+      'ImportDeclaration > ImportSpecifier,ImportNamespaceSpecifier'(node: TSESTree.Node) {
```
Belco90, line 56: More clear now ✨

tests/lib/rules/await-async-utils.test.ts
```diff
@@ -114,33 +143,46 @@ ruleTester.run(RULE_NAME, rule, {
...
           doSomethingElse();
           ${asyncUtil}(() => getByLabelText('email'));
         });
       `,
-      errors: [{ line: 4, messageId: 'awaitAsyncUtil' }],
+      errors: [{ line: 5, messageId: 'awaitAsyncUtil' }],
+    })),
+    ...ASYNC_UTILS.map(asyncUtil => ({
+      code: `
+        import * as asyncUtil from '@testing-library/dom';
```
Belco90, line 156: 💯

dmitry-lobanov:
Hi @Belco90! Thanks, I'm glad to contribute.

Issue #127, opened 2020-05-06
new rule request: prefer-find-by


benmonro:
as described in Kent's blog post, it would be nice to have an autofix rule that converts

```js
await waitFor(() => {
  expect(screen.getByText("foo")).toBeInTheDocument()
}
```
into this:
```js
expect (await screen.findByText("foo")).toBeInTheDocument()
```

Belco90:
Hello Ben! Thanks for your suggestion. Indeed, this would be a nice addition.

Do you want to give it a try? PRs are welcome! I must admit this one could be a hard one to implement tho, as there are plenty of things you can find within `waitFor` callback body.

benmonro:
I wish I could but I'm swamped at work.  I noticed we have a lot of them in our code so figured I'd add the issue. :)

gndelia:
I _might_ be able to take a peek in this one in the incoming days 😄

Belco90:
That would be great! Let me know if you want to discuss some details about the rule or its implementation.

gndelia:
yeah, I have some doubts about being able to fix it in all scenarios... given the callback could potentially have tons of statements - it just seems "feasible" when the statement inside the callback is only one (like a arrow function without brackets, for instance)

Thoughts?

Belco90:
Exactly. We can potentially find _anything_ within callback body passed to `waitFor` method, so just trying to fix it for a single statement seems like the perfect approach for first implementation.

One of the latest ESLint features is the [ability to suggest fixes](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions). This looks like a good place to try it! But I'd leave it as an optional step at the end, as just implementing the rule and fixing the code for single statements is a lot of work.

gndelia:
we could also narrow the fix if that statement inside the callback contains an `expect` with a get/query* method - 

should we also include the deprecated methods (for instance `wait`) as part of the fix?

Belco90:
That seems a good idea. And actually, let's forget about the possibility of adding suggestions and just focus on single statement within `waitFor` cb body.

I think it's not difficult to include deprecated wait utils when it's done for waitFor, so I'd add them too.

gndelia:
I created a draft PR for this one.

[...]

Issue #138, opened 2020-05-17
Support ESLint 7.x


MichaelDeBoey:
ESLint v7.0.0 is [released](https://eslint.org/blog/2020/05/eslint-v7.0.0-released) 🎉

It would be awesome to have official ESLint 7 support. 👊
I'm happy to help where I can of course 🙂

Belco90:
Hey Michael! Thanks for your suggestion and help. Indeed, it would be nice to update the plugin to get compatibility with ESLint v7. I was thinking about adding an ESLint matrix to CI in a similar way as [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react/blob/master/.travis.yml) address it. I don't know if there is a better approach for that.

Anyway, I don't expect to cover so many node and ESLint versions as that plugin. I think just with node 10.12, 12 and 14 (we already have this) and ESLint v5, v6 and v7 (this is missing at the moment) would be enough.

Thoughts?

MichaelDeBoey:
@Belco90 I've opened #139 to support ESLint 7.x

Belco90:
Oh wow, amazing job with that one and all related PRs mate 🎉

Belco90:
I still think we should merge this one. All the tests are passing for ESLint 5, 6 and 7. We are not affected by breaking changes from v7. It doesn't seem the dependencies not upgraded to v7 yet are breaking anything in our side.

nickmccurdy:
Can they be installed with yarn though? If I remember correctly, yarn errors on peer dependency warnings by default. Even if a dependency of this package works, if it doesn't say it works with ESLint 7 and that's the only pinned version, the entire installation would fail.

Belco90:
First notice I have about it. I need to check that behavior when installing with yarn.

BridgeAR:
Installing with yarn works properly as well.

PR #144, opened 2020-05-22
feat: add prefer-find-by rule

gndelia:
This one tackle #127 

While current code works, I still have to polish some stuff and add more scenarios (and update the docs)

### DIFF
--- a/docs/rules/prefer-find-by.md
+++ b/docs/rules/prefer-find-by.md
@@ -1,0 +1,78 @@
+# Suggest using `findBy*` methods instead of the `waitFor` + `getBy` queries (prefer-find-by)
+
+findBy* queries are a simple combination of getBy* queries and waitFor. The findBy\* queries accept the waitFor options as the last argument. (i.e. screen.findByText('text', queryOptions, waitForOptions))
+
+## Rule details
+
+This rule aims to use `findBy*` or `findAllBy*` queries to wait for elements, rather than using `waitFor`, or the deprecated methods `waitForElement` and `wait`.  
+This rules analyzes those cases where `waitFor` is used with just one query method, in the form of an arrow function with only one statement (that is, without a block of statements). Given the callback could be more complex, this rule does not consider function callbacks or arrow functions with blocks of code
+
+Examples of **incorrect** code for this rule
+
+```js
+// arrow functions with one statement, using screen and any sync query method
+const submitButton = await waitFor(() =>
+  screen.getByRole('button', { name: /submit/i })
+);
+const submitButton = await waitFor(() =>
+  screen.getAllByTestId('button', { name: /submit/i })
+);
+
+// arrow functions with one statement, calling any sync query method
+const submitButton = await waitFor(() =>
+  queryByLabel('button', { name: /submit/i })
+);
+
+const submitButton = await waitFor(() =>
+  queryAllByText('button', { name: /submit/i })
+);
+```
+
+Examples of **correct** code for this rule:
+
+```js
+// using findBy* methods
+const submitButton = await findByText('foo');
+const submitButton = await screen.findAllByRole('table');
+
+// using waitForElementToBeRemoved
+await waitForElementToBeRemoved(() => screen.findAllByRole('button'));
+await waitForElementToBeRemoved(() => queryAllByLabel('my label'));
+await waitForElementToBeRemoved(document.querySelector('foo'));
+
+// using waitFor with a function
+await waitFor(function() {
+  foo();
+  return getByText('name');
+});
+
+// passing a reference of a function
+function myCustomFunction() {
+  foo();
+  return getByText('name');
+}
+await waitFor(myCustomFunction);
+
+// using waitFor with an arrow function with a code block
+await waitFor(() => {
+  baz();
+  return queryAllByText('foo');
+});
+
+// using a custom arrow function
+await waitFor(() => myCustomFunction());
+
+// using expects inside waitFor
+await waitFor(() => expect(screen.getByText('bar')).toBeDisabled());
+await waitFor(() => expect(getAllByText('bar')).toBeDisabled());
+```
+
+## When Not To Use It
+
+- Not encouraging use of findBy shortcut from testing library best practices
+
+## Further Reading
+
+- Documentation for [findBy\* queries](https://testing-library.com/docs/dom-testing-library/api-queries#findby)
+
+- Common mistakes with RTL, by Kent C. Dodds: [Using waitFor to wait for elements that can be queried with find\*](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find)
--- a/lib/rules/prefer-find-by.ts
+++ b/lib/rules/prefer-find-by.ts
@@ -1,0 +1,88 @@
+import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils';
+import {
+  isIdentifier,
+  isCallExpression,
+  isMemberExpression,
+  isArrowFunctionExpression,
+} from '../node-utils';
+import { getDocsUrl, SYNC_QUERIES_COMBINATIONS } from '../utils';
+
+export const RULE_NAME = 'prefer-find-by';
+
+type Options = [];
+export type MessageIds = 'preferFindBy';
+
+export const WAIT_METHODS = ['waitFor', 'waitForElement', 'wait']
+
+export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Suggest using find* instead of waitFor to wait for elements',
+      category: 'Best Practices',
+      recommended: 'warn',
+    },
+    messages: {
+      preferFindBy: 'Prefer {{queryVariant}}{{queryMethod}} method over using await {{fullQuery}}'
+    },
+    fixable: null,
+    schema: []
+  },
+  defaultOptions: [],
+
+  create(context) {
+
+    function reportInvalidUsage(node: TSESTree.CallExpression, { queryVariant, queryMethod, fullQuery }: { queryVariant: string, queryMethod: string, fullQuery: string}) {
+      context.report({
+        node,
+        messageId: "preferFindBy",
+        data: { queryVariant, queryMethod, fullQuery },
+      });
+    }
+
+    const sourceCode = context.getSourceCode();
+
+    return {
+      'AwaitExpression > CallExpression'(node: TSESTree.CallExpression) {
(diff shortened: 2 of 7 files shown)


docs/rules/prefer-find-by.md
```diff
@@ -0,0 +1,43 @@
...
+const submitButton = await screen.findAllByRole('table');
+```
+
+## When Not To Use It
+
+TBD
+
+## Further Reading
+
+TBD
```
Belco90, line 43: I would link to [findBy query doc](https://testing-library.com/docs/dom-testing-library/api-queries#findby) here, as it mentions that `findBy` is a shortcut for `waitFor` + `getBy`. I would use the text from that callout as the main inspiration for this url in the description at the top.

Belco90:
Looking good so far!

docs/rules/prefer-find-by.md
```diff
@@ -0,0 +1,43 @@
...
+  queryAllByText('button', { name: /submit/i })
+);
+```
+
+Examples of **correct** code for this rule:
+
+```js
+const submitButton = await findByText('foo');
+
+const submitButton = await screen.findAllByRole('table');
```
benmonro, line 34: I would add some correct usages of waitFor here as well.

docs/rules/prefer-find-by.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+## When Not To Use It
+
+TBD
```
benmonro, line 39:
```suggestion
Don't use those rule if you don't care if people aren't following best practices of testing library.
```

lib/rules/prefer-find-by.ts
```diff
@@ -0,0 +1,88 @@
...
+export const WAIT_METHODS = ['waitFor', 'waitForElement', 'wait']
+
+export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Suggest using find* instead of waitFor to wait for elements',
+      category: 'Best Practices',
+      recommended: false,
```
benmonro, line 24: Why not recommend this one?

lib/rules/prefer-find-by.ts
```diff
@@ -0,0 +1,88 @@
...
+    },
+    messages: {
+      preferFindBy: 'Prefer {{queryVariant}}{{queryMethod}} method over using await {{fullQuery}}'
+    },
+    fixable: null,
```
benmonro, line 29: Can we not make the simple ones fixable?

docs/rules/prefer-find-by.md
Belco90: I wouldn't go that _direct_ here. Better something nicer like "Not encouraging use of `findBy` shortcut from testing library best practices".

lib/rules/prefer-find-by.ts
Belco90: This is still a draft Ben. Leave the guy work on this! We are discussing in the original issue if addressing it here or in a 2nd PR.

lib/rules/prefer-find-by.ts
Belco90: I think it should be recommended, indeed.

lib/rules/prefer-find-by.ts
benmonro: Was just an idea.  Not trying to clock this or anything...

lib/rules/prefer-find-by.ts
gndelia: "warn" or "error" ?

gndelia:
I pushed a new version.

- CI passes: More test cases!
- Docs updated: I added links, details, and more examples

still pending to make it fixable

lib/rules/prefer-find-by.ts
Belco90: Nice question. I think warn is enough? This rules doesn't prevent any malfunctioning code, just warns about a preferred way of writing it.

lib/rules/prefer-find-by.ts
```diff
@@ -0,0 +1,88 @@
...
+  isMemberExpression,
+  isArrowFunctionExpression,
+} from '../node-utils';
+import { getDocsUrl, SYNC_QUERIES_COMBINATIONS } from '../utils';
+
+export const RULE_NAME = 'prefer-find-by';
+
+type Options = [];
+export type MessageIds = 'preferFindBy';
+// TODO check if this should be under utils.ts - there are some async utils
```
Belco90, line 14: We have `ASYNC_UTILS` in `utils.ts` but it includes more methods than here.

Belco90: Maybe it's a good idea to keep all different async utils under an enumerate type and then build `ASYNC_UTILS` all the enumerate options?

Belco90:
> I pushed a new version.
> 
>     * CI passes: More test cases!
> 
>     * Docs updated: I added links, details, and more examples
> 
> 
> still pending to make it fixable

Nice work mate. It would be nice to have the fixable code here, but let me know what you think about it and if you prefer to address it in a separated PR in the future.

lib/rules/prefer-find-by.ts
gndelia: I added `warn`

gndelia:
This PR is ready to be reviewed.

I gave [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) a try (instead of directly fixing the code), but the problem is that types in ESLint 6 differ from ESLint 7. I have a local branch with the `suggestions` which _I believe_ are working, but the tests always pass for the suggestions; It's like I cannot force them to fail. I believe that once #138  is finished, suggestions should work better

So this PR goes without suggestions for fix available 😞

docs/rules/prefer-find-by.md
```diff
@@ -0,0 +1,74 @@
...
+
+Examples of **incorrect** code for this rule
+
+```js
+// arrow functions with one statement, using screen and any sync query method
+const submitButton = await waitFor(() =>
+  screen.getByRole('button', { name: /submit/i })
+);
+const submitButton = await waitFor(() =>
+  screen.getAllTestId('button', { name: /submit/i })
```
Belco90, line 18:
```suggestion
  screen.getAllByTestId('button', { name: /submit/i })
```

docs/rules/prefer-find-by.md
```diff
@@ -0,0 +1,74 @@
...
+await waitFor(myCustomFunction);
+
+// using waitFor with an arrow function with a code block
+await waitFor(() => {
+  baz();
+  return queryAllByText('foo');
+});
+
+// using a custom arrow function
+await waitFor(() => myCustomFunction());
```
Belco90, line 63: Can you add an extra correct usage of `waitFor` within expects?

```javascript
await waitFor(() => expect(screen.getByText('bar').toBeDisabled());
```

tests/lib/rules/prefer-find-by.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      code: `await waitForElementToBeRemoved(document.querySelector('foo'))`
+    },
+    ...SYNC_QUERIES_COMBINATIONS.map((queryMethod: string) => ({
+      code: `
+        await waitFor(() => {
+          foo()
+          return ${queryMethod}()
+        })
+      `
+    })),
```
Belco90, line 44: Could you add valid tests for the example requested in the doc? A valid `waitFor` returning expect with getBy inside.

docs/rules/prefer-find-by.md
gndelia: done

tests/lib/rules/prefer-find-by.test.ts
gndelia: done

lib/rules/prefer-find-by.ts
gndelia: after giving some though, I am not sure of moving this one out. It's not a list of all `wait` methods, but just a list of the ones this rule cares about. Not sure how useful it could be outside of here, as it's not _all_ the `wait` methods. Either way, we could centralize them in the future if we see the chance

gndelia:
BTW, I added the new rule to the main README.

image




> Looking good mate, thanks for your work! I just requested couple of extra examples and tests, everything else seems fine. We are leaving the fix step for a future PR then right?

yes. The problem is that when testing `suggestions`, no matter what the `expected` output I write, the tests always passes. Let's look at this example.

```ts
code: `
            const submitButton = await ${waitMethod}(() => screen.${queryMethod}('foo', { name: 'baz' }))
          `,
          errors: [{
            messageId: 'preferFindBy',
            data: {
              queryVariant: queryMethod.includes('All') ? 'findAllBy': 'findBy',
              queryMethod: queryMethod.split('By')[1],
              fullQuery: `${waitMethod}(() => screen.${queryMethod}('foo', { name: 'baz' }))`,
            },
            suggestions: [{
              messageId: 'preferFindBy',
              // this expected output should match the suggestion output of the rule
              output: `const submitButton = await screen.${queryMethod.includes('All') ? 'findAllBy': 'findBy'}${queryMethod.split('By')[1]}('foo', { name: 'baz' })`
            }]
          }]
        })
```

No matter what output I write in the test, the tests pass, making me think that there's something wrong with the test assertion declaration.

I think the problem might be solved with ESlint 7 - furthermore, the contract changed in that version so it'll be easier to implement once that is solved.

We could probably go with that in another PR, does that sound good for you?

Belco90:
@gndelia that sounds good. I remember I had a similar problem when implementing the fix for the `waitFor` migration. We can wait a little bit until we find proper way to test it, but I think we should go for fix rather than suggestion as there is just single way of changing a getBy + waitFor by a findBy.

docs/rules/prefer-find-by.md
```diff
@@ -0,0 +1,78 @@
+# Use `find*` query methods to wait for elements instead of waitFor (prefer-find-by)
```
Belco90, line 1: Can you use here same description from README please?

```suggestion
# Suggest using `findBy*` methods instead of the `waitFor` + `getBy` queries (prefer-find-by)
```

Belco90:
I've been thinking about this rule name. Could we regret it in the future for being so generic? Could we have in the future other cases where we prefer `findBy` over other options? I don't know if could be worth it to think about a better name for the rule. ESLint rules names should be short tho, but I'd like to think about it. Could something like `no-get-by-wait-for-combo` work? Maybe I'm overthinking around this.

benmonro:
I think when it comes to rule names less is more. Either prefer-find-by or no-wait-for-get

gndelia:
we could expand the rules or rename (breaking change) if we get a collision in the future. Not sure which other case `findBy*` offers, to prefer it over something else...

docs/rules/prefer-find-by.md
gndelia: done

PR #150, opened 2020-06-05
fix(prefer-screen-queries): take container into account

### LINKED ISSUE #149: Shouldn't prefer-screen-queries takes into account querySelector?
`prefer-screen-queries` is not taken into account that the user can do this:

```
const utils = render(<Component />);
utils.container.querySelector('div');
```

This kind of selector can lead the user to a bad pattern of tests according to the [testing-library](https://testing-library.com/docs/guide-which-query).

I don't understand eslint under the hood, but I propose to change the `prefer-screen-queries` to take the snippet above into account as a bad pattern.

What do you guys think about that?

lourenci:
It fixes #149.

### DIFF
--- a/docs/rules/prefer-screen-queries.md
+++ b/docs/rules/prefer-screen-queries.md
@@ -2,7 +2,7 @@
 
 ## Rule Details
 
-DOM Testing Library (and other Testing Library frameworks built on top of it) exports a `screen` object which has every query (the `container` and a `debug` method). This works better with autocomplete and makes each test a little simpler to write and maintain.
+DOM Testing Library (and other Testing Library frameworks built on top of it) exports a `screen` object which has every query (plus the `container` and a `debug` method). This works better with autocomplete and makes each test a little simpler to write and maintain.
 This rule aims to force writing tests using queries directly from `screen` object rather than destructuring them from `render` result.
 
 Examples of **incorrect** code for this rule:
--- a/docs/rules/prefer-screen-queries.md
+++ b/docs/rules/prefer-screen-queries.md
@@ -2,26 +2,30 @@
 
 ## Rule Details
 
-DOM Testing Library (and other Testing Library frameworks built on top of it) exports a `screen` object which has every query (and a `debug` method). This works better with autocomplete and makes each test a little simpler to write and maintain.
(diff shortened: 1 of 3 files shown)


tests/lib/rules/prefer-screen-queries.test.ts
```diff
@@ -4,9 +4,11 @@ import { ALL_QUERIES_COMBINATIONS } from '../../../lib/utils';
 
 const ruleTester = createRuleTester();
 
+const ALL_QUERIES_COMBINATIONS_PLUS_CONTAINER = [...ALL_QUERIES_COMBINATIONS, 'container']
```
gndelia, line 7: I think that it would be better to keep `ALL_QUERIES_COMBINATIONS` and then use container for its specific scenarios. As this would test as valid scenarios that do not match real tests. For instance, in the line 12 it would translate to

```ts
screen.container()
```

which is not RTL valid code. Same below for

```js
within(component).container()
```

docs/rules/prefer-screen-queries.md
```diff
@@ -2,26 +2,30 @@
 
 ## Rule Details
 
-DOM Testing Library (and other Testing Library frameworks built on top of it) exports a `screen` object which has every query (and a `debug` method). This works better with autocomplete and makes each test a little simpler to write and maintain.
+DOM Testing Library (and other Testing Library frameworks built on top of it) exports a `screen` object which has every query (the `container` and a `debug` method). This works better with autocomplete and makes each test a little simpler to write and maintain.
```
gndelia, line 5: Minor spelling mistake, I think ? English is not my native language 😄 

```suggestion
DOM Testing Library (and other Testing Library frameworks built on top of it) exports a `screen` object which has every query (plus the `container` and a `debug` method). This works better with autocomplete and makes each test a little simpler to write and maintain.
```

gndelia:
Thank you for the PR! 

I added a couple of small comments 😄

docs/rules/prefer-screen-queries.md
lourenci: It's not mine either. 🙂

Belco90: Agree on @gndelia suggestion as container and debug are extra things apart from the queries, so it's better to say "screen has every query plus container and debug).

Belco90:
@all-contributors please add @lourenci for bug, code and test

PR #152, opened 2020-06-09
fix(prefer-wait-for): only report when imported from TL

### LINKED ISSUE #151: prefer-wait-for rule autofix is modifying Cypress wait() functions
Hello,

I have tried to add the rule 'prefer-wait-for' to my code but an unexpected interaction is provoked with my Cypress files. The autofix from the rule is also changing the calls to cy.Wait() for cy.WaitFor() in my Cypress test files

Screenshot 2020-06-09 at 14 04 13

timdeschryver:
Closes #151

### DIFF
--- a/lib/rules/prefer-wait-for.ts
+++ b/lib/rules/prefer-wait-for.ts
@@ -34,9 +34,6 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
   defaultOptions: [],
 
   create(context) {
-    const importNodes: TSESTree.ImportDeclaration[] = [];
-    const waitNodes: TSESTree.Identifier[] = [];
-
     const reportImport = (node: TSESTree.ImportDeclaration) => {
       context.report({
         node: node,
@@ -44,7 +41,6 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
         fix(fixer) {
           const excludedImports = [...DEPRECATED_METHODS, 'waitFor'];
 
-          // TODO: refactor `importNodes` to TSESTree.ImportSpecifier[] ? (to not have to traverse the list twice)
           // get all import names excluding all testing library `wait*` utils...
           const newImports = node.specifiers
             .filter(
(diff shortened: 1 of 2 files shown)


lib/rules/prefer-wait-for.ts
```diff
@@ -119,36 +115,44 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
 
-        importNodes.forEach(importNode => {
-          reportImport(importNode);
+          context
+            .getDeclaredVariables(importSpecifier)
+            .forEach(variable =>
+              variable.references.forEach(reference =>
+                reportWait(reference.identifier)
+              )
+            );
```
timdeschryver, line 136: Is there a way we could make this cleaner? 👀

tests/lib/rules/prefer-wait-for.test.ts
```diff
@@ -41,6 +41,25 @@ ruleTester.run(RULE_NAME, rule, {
...
+
+      async () => {
+        await foo.wait();
+      }`,
+    },
+    {
+      code: `
+      cy.wait();
+      `,
+    },
```
gndelia, line 62: I think we should consider testing `require` scenarios as well

lib/rules/prefer-wait-for.ts
gndelia: perhaps

```ts
context
  .getDeclaredVariables(importSpecifier)
  .flatMap(variable => variable.references)
  .forEach(({ identifier })=> reportWait(identifier))
```

? But IMO it's not too much of a difference - I think this is fine

And I think flatMap does not work in node <= 10

Belco90: Indeed, we can't use [flatMap in node 10](https://node.green/#ES2019-features-Array-prototype--flat--flatMap-)

Belco90: Maybe we could make the cod cleaner by extracting the first forEach callback to its own function? It's the only thing I can think of.

tests/lib/rules/prefer-wait-for.test.ts
Belco90: It would be nice if we can have an additional test for it, indeed.

timdeschryver:
This won't be just additional tests, it would also mean that we have to modify the rule.
e.g. the following is currently valid (but invalid with `import`):

```ts
const { wait } = require("@testing-library/foo");

async () => {
  await wait();
}
```

Belco90:
Ah of course. Fair enough, I'll approve it then.

Belco90:
@all-contributors please add @miguelerja for bug

PR #153, opened 2020-06-09
feat(prefer-find-by): make the rule fixable

gndelia:
This feature extends prefer-find-by rule to make it fixable

One doubt I had was whether to fix or not this scenario

```ts
waitFor(() => query*)
```

as `queryBy*` methods do not throw if not found, but `findBy*` do. At first I thought of not changing the behavior, but then I consider that if null was always returned, eventually `waitFor` would timeout, throwing as well (just like the `findBy*` method would)


Belco90:
`waitFor` + `queryBy*` should be the same as `findBy` too. As far as I know, `waitFor` will wait until the query returns an element, so `null` (from `queryBy`) or error (from `getBy`) are both considered as still waiting for element, ending up with the same behavior under `waitFor`.

So I think that scenario should be fixed too.

lib/rules/prefer-find-by.ts
```diff
@@ -26,23 +26,26 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
+    const sourceCode = context.getSourceCode();
 
-    function reportInvalidUsage(node: TSESTree.CallExpression, { queryVariant, queryMethod, fullQuery }: { queryVariant: string, queryMethod: string, fullQuery: string}) {
+    function reportInvalidUsage(node: TSESTree.CallExpression, { queryVariant, queryMethod, fullQuery, callArguments, caller }: { queryVariant: string, queryMethod: string, fullQuery: string, callArguments: TSESTree.Expression[], caller?: string }) {
       context.report({
         node,
         messageId: "preferFindBy",
         data: { queryVariant, queryMethod, fullQuery },
+        fix(fixer) {
+          const newCode = `${caller ? `${caller}.` : ''}${queryVariant}${queryMethod}(${callArguments.map((node) => sourceCode.getText(node)).join(', ')})`
```
Belco90, line 43: Damn, even checking the test I can't figure out how this replaces `waitFor` + `getBy` with `findBy`

Belco90: Do you want me to write a suggestion for this jsdoc? We need something simple, not elaborated description necessary here.

gndelia: As you want. I can do it, but I will probably do it in the weekend as I have some work to do 😄

Belco90: No problem, there is no rush around this!

gndelia: Done!
Hovering the method shows the docs:

image

Belco90: Awesome! It wasn't even necessary to re-add the type, but thanks!

PR #161, opened 2020-06-12
fix: ensure node exists before using in isArrowFunctionExpression

jrparish:
This fixes an issue where the linter blows up when there are no arguments in `wait()` or `waitFor()`

image


jrparish:
> I see this is failing for `prefer-find-by` from your screenshot. Could you add an additional test within that rule with empty `wait` please?

For sure! I had the test locally but wasn't sure if it was desired since not entirely related 😄

Belco90:
@all-contributors please add @jrparish for bug, code and test

Issue #165, opened 2020-06-15
Add rules from "Common mistakes with React Testing Library"


nickmccurdy:
There are some great suggestions in [Common mistakes with React Testing Library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library), a few of which we could probably check statically:

1. [Using wrapper as the variable name for the return value from render](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-wrapper-as-the-variable-name-for-the-return-value-from-render)
1. [Wrapping things in act unnecessarily](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) (might be tricky, but maybe we could filter AST nodes and only error if something is used that isn't from testing-library)
1. [Having multiple assertions in a single waitFor callback
](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback)

Already implemented but should maybe be enabled by default:
1. [Using cleanup](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-cleanup)
1. [Using container to query for elements](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-container-to-query-for-elements)
1. [Passing an empty callback to waitFor](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#passing-an-empty-callback-to-waitfor)

Related issues: #133 #134 #162

Belco90:
Hey Nick, thanks for bringing this discussion here!

Indeed, there are some other common mistakes we can prevent. From your list of new rules:

1. This one is an easy one to implement. Just something like `no-render-wrapper` or something similar would be enough. Should be a fixable rule? I'd rather not, as we would have to rename all appearances of `wrapper` const defined by the user and we can mess up the code 
2. Not really sure about this one. Can we really check this statically? I'm afraid about throwing errors for something that is actually fine.
3. We were discussing around this one when implementing the `prefer-find-by` one. I think it needs more discussion to clarify the exact behavior for it. Do we care if there are several expects but also other stuff? Should be fixable? What if it's fixable the first thing within `waitFor` callback isn't an expect? (I guess in this case we won't automatically fix it).

About already implemented ones:
[...]

PR #168, opened 2020-06-17
feat: recommend no-wait-for-empty-callback

nickmccurdy:


Belco90:
I'd say this shouldn't be a breaking change, or at least we haven't treat changes in presets as breaking changes. I should read more about this.

nickmccurdy:
I thought they would be breaking changes since they could cause additional failures when the configs are in use. If you disagree I can rebase these to not include `BREAKING CHANGE:` and it will be a minor/patch release instead.

Belco90:
Don't worry, I can address that when squashing and merging the PR. I'll read more about it and see how other libraries treat this.

Belco90:
Ok so other plugins are handling this as a major release. I think it would be a good idea then to wrap all preset changes into a milestone, merge them into a v4 branch and wait for some other additions we have pending. So we release a v4 including several breaking changes for each preset update, and also add new rules and improvements. What do you think? I can create the branch and the milestone later.

nickmccurdy:
That sounds good, what should I do with these pull requests for now though? Also, would it make sense to also have all the other changes related to #165 in this new release?

timdeschryver:
a v4 is a good idea.
Looks like we will be adding more rules in a short time period.

@nickmccurdy when @Belco90 created a v4 branch,  we can rebase this PR to the v4 branch.

Belco90:
Branch for `v4` created. You can update the PRs to point that branch now. I'm creating the milestone to track everything from #165 there.

PR #177, opened 2020-06-19
feat: add no-container rule

### LINKED ISSUE #171: new rule: no-container
__Inspiration:__ [Using container to query for elements](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-container-to-query-for-elements)

This would error on `container` methods like `querySelector` that are often used instead of get/query functions, as there might be other reasons why you'd need to access the container in a test (such as assertions). It would be nice to give suggestions of specific get/query functions as well, but I'm not sure how far we could get.

thebinaryfelix:
Closes #171 

This PR adds a new `no-container` rule for disallowing the use of container methods like `querySelector`, following [best practices](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-container-to-query-for-elements).

A refactor of the `no-debug` rule was made in order to extract two helper functions to **node-utils.ts** so they could be used by `no-container`.

Preview of how the message will appear (in VS Code):

screenshot-no-container-rule


docs/rules/no-container.md
```diff
@@ -0,0 +1,36 @@
+# Disallow the use of `container` methods (no-container)
+
+By using `container` methods like `.querySelector` you may lose a lot of the confidence that the user can really interact with your UI. Also, the test becomes harder to read, and it will break more frequently.
```
Belco90, line 3: Could you specify this applies to Testing Library frameworks built on top of DOM Testing Library please?

docs/rules/no-container.md
```diff
@@ -0,0 +1,36 @@
...
+
+If you use [custom render functions](https://testing-library.com/docs/example-react-redux) then you can set a config option in your `.eslintrc` to look for these.
+
+```
+   "testing-library/no-container": ["error", {"renderFunctions":["renderWithRedux", "renderWithRouter"]}],
+```
+
+## Further Reading
+
+- [querying with `screen`](https://testing-library.com/docs/dom-testing-library/api-queries#screen)
```
Belco90, line 36: Could you add here corresponding link to one of Testing Library framework's `container` please? React one for example: https://testing-library.com/docs/react-testing-library/api#container-1

docs/rules/no-container.md
```diff
@@ -0,0 +1,36 @@
...
+
+```js
+render(<Example />);
+screen.getByRole('button', { name: /click me/i });
+```
+
+If you use [custom render functions](https://testing-library.com/docs/example-react-redux) then you can set a config option in your `.eslintrc` to look for these.
+
+```
+   "testing-library/no-container": ["error", {"renderFunctions":["renderWithRedux", "renderWithRouter"]}],
```
Belco90, line 31: Nice! Thanks for extracting the utils methods for working with this. We may need to set this globally in the plugin through ESLint settings in the future @nickmccurdy @timdeschryver @gndelia

lib/index.ts
```diff
@@ -34,6 +36,7 @@ const recommendedRules = {
   'testing-library/await-async-query': 'error',
   'testing-library/await-async-utils': 'error',
   'testing-library/no-await-sync-query': 'error',
+  'testing-library/no-container': 'error',
```
Belco90, line 39: Per previous comment, this should added to angular, react and vue presets individually, rather than here.

lib/rules/no-container.ts
```diff
@@ -0,0 +1,82 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow the use of container methods',
+      category: 'Best Practices',
+      recommended: 'error',
+    },
+    messages: {
+      noContainer:
+        'Unexpected use of container methods. Prefer the use of "screen.someMethod()".',
```
Belco90, line 25: What do you think about this message? Lifted from Testing Library docs.

```suggestion
        'Avoid using container to query for elements. Use some Testing Library query instead.',
```

gndelia:
Thanks for your PR! There are some conflicts with the files. Could you rebase/update your PR ? Thanks!

docs/rules/no-container.md
```diff
@@ -0,0 +1,36 @@
...
+
+```js
+const { container } = render(<Example />);
+const button = container.querySelector('.btn-primary');
+```
+
+```js
+const { container: alias } = render(<Example />);
+const button = alias.querySelector('.btn-primary');
+```
```
gndelia, line 19: Can we add some scenarios here?

I can think of 

```js
// using container from screen
const button = screen.container.querySelector('.btn-primary')

// using container from the result from calling render, without destructuring
const view = render(<Example />)
const button = view.container.querySelector('.btn-primary')
```

tests/lib/rules/no-container.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    {
+      code: `
+        const { container } = render(<Example />)
+        expect(container.firstChild).toBeDefined()
+      `,
+    },
+    {
+      code: `
+        const { container: alias } = render(<Example />)
+        expect(alias.firstChild).toBeDefined()
```
Belco90, line 27: Nice one 💯

lib/rules/no-container.ts
gndelia, line 25: I'm thinking this rule is more about disabling container, but there's no mention about using screen or not. So perhaps the text could go something like

```suggestion
        'Unexpected use of container methods. Prefer using query methods, such as "getByRole()"',
```

tests/lib/rules/no-container.test.ts
```diff
@@ -0,0 +1,82 @@
...
+          renderFunctions: ['renderWithRedux'],
+        },
+      ],
+      errors: [
+        {
+          messageId: 'noContainer',
+        },
+      ],
+    },
+  ],
```
gndelia, line 81: can we also add these scenarios that I mentioned in the docs?

```ts
// using container from screen
const button = screen.container.querySelector('.btn-primary')

// using container from the result from calling render, without destructuring
const view = render(<Example />)
const button = view.container.querySelector('.btn-primary')
```

lib/rules/no-container.ts
gndelia: Perhaps someone else can come up with a better error text too 😅

thebinaryfelix:
> I added some comments. I Will continue reviewing it later. Thank you very much for this PR 🚀

I'll take a look and work on them later, thanks for the review!

tests/lib/rules/no-container.test.ts
```diff
@@ -0,0 +1,82 @@
...
+import rule, { RULE_NAME } from '../../../lib/rules/no-container';
+
+const ruleTester = createRuleTester({
+  ecmaFeatures: {
+    jsx: true,
+  },
+});
+
+ruleTester.run(RULE_NAME, rule, {
+  valid: [
```
Belco90, line 11: Could you add an additional valid test for using `container` not related to Testing Library? Something like:

```javascript
function getExampleDOM() {
  const container = document.createElement('div')
  container.innerHTML = `
    <label for="username">Username</label>
    <input id="username" />
    <button>Print Username</button>
  `
  const button = container.querySelector('button')

  button.addEventListener('click', () => console.log('clicked'));
  return container;
}

const exampleDOM = getExampleDOM();

screen.getByText(exampleDOM, 'Print Username').click();
```

This is a simplified code from [DOM Testing Library example](https://testing-library.com/docs/dom-testing-library/example-intro).

lib/rules/no-container.ts
Belco90: Though I just realized it could mean any method, not just queries.

README.md
```diff
@@ -140,6 +140,7 @@ To enable this configuration use the `extends` property in your
 | [await-fire-event](docs/rules/await-fire-event.md)                     | Enforce async fire event methods to be awaited                             | ![vue-badge][]                                                            |                    |
 | [consistent-data-testid](docs/rules/consistent-data-testid.md)         | Ensure `data-testid` values match a provided regex.                        |                                                                           |                    |
 | [no-await-sync-query](docs/rules/no-await-sync-query.md)               | Disallow unnecessary `await` for sync queries                              | ![recommended-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |                    |
+| [no-container](docs/rules/no-container.md)                             | Disallow the use of `container` methods                                    | ![recommended-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |                    |
```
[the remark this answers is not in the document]
thebinaryfelix: Nice, I didn't realize that 😅  I just thought about the literal meaning of something being "recommended". I'll remove this badge and other configs that set this rule as recommended.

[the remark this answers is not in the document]
gndelia: Me too!

I learned something new today 😀

lib/rules/no-container.ts
thebinaryfelix: Yes, it's a good move to keep consistency between lint messages and the lib's docs. I think we could be more descriptive with the suggestion of what to do, so by mixing your suggested message with the one from @gndelia we could get something like:
`Avoid using container to query for elements. Prefer using query methods from Testing Library, such as "getByRole()"`, what do you think?

gndelia: Sounds good to me, but I'd wait for @Belco90  confirmation as well. Minor (my English is not native so I might be wrong), but I think it should be `Avoid using container to query elements` (dropping for)

docs/rules/no-container.md
gndelia: sorry @thebinaryfelix  can we remove this one ?

```ts
const button = screen.container.querySelector('.btn-primary')
```

it was my mistake (2nd time I do this error 😢 ) - `container` is not available in the `screen` object - so this code makes no sense. I am sorry!

tests/lib/rules/no-container.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    {
+      code: `
+        const button = screen.container.querySelector('.btn-primary')
+      `,
+      errors: [
+        {
+          messageId: 'noContainer',
+        },
+      ],
+    },
```
gndelia, line 93: Same here, this code is invalid. I am sorry I said it was valid, but `container` is not exposed in the `screen` object. We should drop this

gndelia:
I think that by using destructuring, it'd be also possible to bypass this rule

```
const { container: { querySelector } } = render(<Example />);
querySelector('foo');
```

I don't generally use that double-nested destructuring because I find it really hard to read 😆 but it's valid code. The rule should probably cover it. Can we add it to the docs as invalid, and in the tests? I followed the code and I think it does not cover this scenario (but I might be wrong)

lib/rules/no-container.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+      },
+
+      CallExpression(node: TSESTree.CallExpression) {
+        function showErrorForChainedContainerMethod(
+          innerNode: TSESTree.MemberExpression
+        ) {
+          if (isMemberExpression(innerNode)) {
+            if (isIdentifier(innerNode.object)) {
+              const isScreen = innerNode.object.name === 'screen';
```
gndelia, line 75: we should drop this line. `screen` does not expose the `container` property - see the types [here](https://github.com/testing-library/dom-testing-library/blob/master/types/screen.d.ts). Sorry if I caused confusion when mentioning it before! My bad

thebinaryfelix: Sure! No problem :)

thebinaryfelix:
> I think that by using destructuring, it'd be also possible to bypass this rule
> 
> ```ts
> const { container: { querySelector } } = render(<Example />);
> querySelector('foo');
> ```
> 
> I don't generally use that double-nested destructuring because I find it really hard to read laughing but it's valid js code. The rule should probably cover it. Can we add it to the docs as invalid, and in the tests? I followed the code and I think it does not cover this scenario (but I might be wrong)

You're right! It didn't cover this scenario, I've adjusted the code now. If you think of any other case, please let me know.

thebinaryfelix:
> > I think that by using destructuring, it'd be also possible to bypass this rule
> > ```ts
> > const { container: { querySelector } } = render(<Example />);
> > querySelector('foo');
> > ```
> > 
> > 
> > I don't generally use that double-nested destructuring because I find it really hard to read laughing but it's valid js code. The rule should probably cover it. Can we add it to the docs as invalid, and in the tests? I followed the code and I think it does not cover this scenario (but I might be wrong)
> 
> You're right! It didn't cover this scenario, I've adjusted the code now. If you think of any other case, please let me know.

Actually there's another case where the user could rename the method by doing something like:

```
const { container: { querySelector: alias } } = render(<Example />);
alias("foo");
```
And we could verify that recursively, but do we really wanna go that far to validate it? Also, if someone has another approach for how to check that, feel free to share it.

lib/rules/no-container.ts
Belco90: It didn't sound right to me but I just got it from Testing Library doc so I didn't say anything about rewording it.

Belco90:
> Actually there's another case where the user could rename the method by doing something like:
> 
> ```
> const { container: { querySelector: alias } } = render(<Example />);
> alias("foo");
> ```
> 
> And we could verify that recursively, but do we really wanna go that far to validate it? Also, if anyone has another approach for how to check that, feel free to share it.

I'm against going that far in general. We need to cover different cases which are pretty common, but this is a really weird edge case I would say. So I'd go ahead with a basic coverage of different usages of this method (which is actually really nice) and leave that complex destructuring case as a future improvement.

lib/rules/no-container.ts
```diff
@@ -0,0 +1,120 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow the use of container methods',
+      category: 'Best Practices',
+      recommended: 'error',
+    },
+    messages: {
+      noContainer:
+        'Avoid using container to query for elements. Prefer using query methods from Testing Library, such as "getByRole()"',
```
Belco90, line 24: This message and rule doc examples only mention that it's a problem to use `container` to query elements, but rule doc description and rule implementation itself are checking all `container` methods. I guess rule doc examples and this message should be updated accordingly.

thebinaryfelix: So, you're suggesting we show a message like this: `Avoid using container methods. Prefer using the methods from Testing Library, such as "getByRole()"`?

gndelia:
@all-contributors please add @thebinaryfelix for code, tests and doc

Belco90:
I need to update contributing guidelines with a section for maintainers. Meanwhile, please let's try to squash merge PRs to make the commit messages fit with the semantic release changelogs (except when merging v4 to master for example, so we don't ruin the original commits as I did when merging v3 😓 ).

gndelia:
Hey @Belco90 I don't have access, but I think there should be a way in the settings to only allow squash merges in PR

image

(this is from another project I work for)

Perhaps that could help to avoid missing squashes. I reviewed the commit message so it fit the semantic release, but I missed the squash. Sorry for that 😓

Belco90:
No problem, not a big deal as I can update the release notes after the release itself. I don't want to force always squash merging tho, as it's useful to allow merge commits too (for example when merging v4 to master).

PR #180, opened 2020-06-20
feat(no-promise-in-fire-event): add new no-promise-in-fire-event rule

### LINKED ISSUE #166: fireEvent with Promises
Based on https://github.com/testing-library/dom-testing-library/issues/609, I think we could add a new rule. I'm not 100% sure if it's worth it, because:

- testing library already added a runtime error
- in TypeScript projects, this will already be caught (I think)
- the current implementation of the rule doesn't catch all cases, and I'm not sure if it can. Currently it catches `fireEvent(Promise)`, `fireEevent(new Promise()`, `fireEvent(findBy...)`

The implementation of the rule might look like this.
What do you tink?

```ts
export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
  name: RULE_NAME,
  meta: {
    type: 'suggestion',
    docs: {
      description: 'TK',
      category: 'Best Practices',
      recommended: false,
    },
    messages: {
      noPromiseInEvent: 'TK',
    },

    fixable: 'code',
    schema: [],
  },
  defaultOptions: [],

  create(context) {
    return {
      'ImportDeclaration[source.value=/testing-library/]'(
        node: TSESTree.ImportDeclaration
      ) {
        const fireEventImportNode = node.specifiers.find(
          specifier =>
            isImportSpecifier(specifier) &&
            specifier.imported &&
            'fireEvent' === specifier.imported.name
        ) as TSESTree.ImportSpecifier;

        const {references} = context.getDeclaredVariables(fireEventImportNode)[0];

        for (const reference of references) {
          const referenceNode = reference.identifier;
          if (
            isMemberExpression(referenceNode.parent) &&
            isCallExpression(referenceNode.parent.parent)
          ) {
            const [element] = referenceNode.parent.parent
              .arguments as TSESTree.Node[];
            if (element) {
              if (isCallExpression(element) || isNewExpression(element)) {
                const methodName = isIdentifier(element.callee)
                  ? element.callee.name
                  : isMemberExpression(element.callee) &&
                    isIdentifier(element.callee.property)
                  ? element.callee.property.name
                  : '';

                if (
                  ASYNC_QUERIES_VARIANTS.includes(methodName) ||
                  methodName === 'Promise'
                ) {
                  context.report({
                    node: element,
                    messageId: 'noPromiseInEvent',
                  });
                }
              }
            }
          }
        }
      },
    };
  },
});
```

timdeschryver:
Closes #166 

I did not make this recommended to not create a breaking change.
Should it be a recommended query and targeted against the v4 branch?
Or do we merge this into master, and then make a PR to make it recommended (if we want) against the v4 branch?


nickmccurdy:
> Should it be a recommended query and targeted against the v4 branch?
Or do we merge this into master, and then make a PR to make it recommended (if we want) against the v4 branch?

I think that's up to @Belco90

Belco90:
I'd go for targeting v4, so we can release all this together as a huge update?

nickmccurdy:
I can change the base branch to v4 if you want

docs/rules/no-promise-in-fire-event.md
```diff
@@ -0,0 +1,31 @@
...
+Examples of **incorrect** code for this rule:
+
+```js
+import { screen, fireEvent } from '@testing-library/react';
+
+// usage of findBy queries
+fireEvent.click(screen.findByRole('button'));
+
+// usage of promises
+fireEvent.click(new Promise(jest.fn())
```
gndelia, line 14: Does this cover the scenario where the promise is in a variable ? Like

```ts
const promise = new Promise(jest.fn())
fireEvent.click(promise)
```

If so, perhaps we should add it here in the docs

timdeschryver: That's great feedback!
It doesn't cover this case, not sure how we can check that :/

Issue #134, opened 2020-05-13
new rule: no-side-effects-wait-for


renatoagds:
Following the proposed for Kent at his post (https://kentcdodds.com/blog/common-mistakes-with-react-testing-library). I would like to suggest a new rule that detects side effects inside a wait-for:

```javascript
❌ // wrong
await waitFor(() => {
  fireEvent.click(a)
  expect(b).toEqual('b')
})
```

```javascript
✅ // correct
fireEvent.click(a)
await waitFor(() => expect(b).toEqual('b'))
```

Belco90:
This one duplicates #133

Belco90:
Sorry, I closed this one by mistake! Thanks again Renato, this suggestion sounds good too.

We would have to identify which _side effects_ we want to check here, I guess those from testing library itself.

Again, would like to give it a try? PRs are welcome!

renatoagds:
> We would have to identify which side effects we want to check here, I guess those from testing library itself.

@Belco90 Yes! I was think to add `fireEvent` and `userEvent` initially. Sounds good?

Belco90:
Sounds good!

renatoagds:
Just to update! I'm finishing #133 this week and will start work on this one 🚀

Belco90:
Awesome, thanks for your help 💪

PR #189, opened 2020-06-25
feat: add rule no-multiple-assertions-wait-for

### LINKED ISSUE #133: new rule: no-multiple-assertions-wait-for
Following the proposed in Kent at his post (https://kentcdodds.com/blog/common-mistakes-with-react-testing-library). I would like to suggest a new rule that detects multiple expect inside a wait-for:

```javascript
❌ // wrong
await waitFor(() => {
  expect(a).toEqual('a')
  expect(b).toEqual('b')
})
```

```javascript
✅ // correct
await waitFor(() => expect(a).toEqual('a'))
expect(b).toEqual('b')
```

renatoagds:
Add a new rule, based in a tip from [`common-mistakes`](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library), to avoid multiple assertions inside a `waitFor`.

This rule just get assertions inside `waitFor`, this way it doesn't conflict with other rules like #134 

Closes #133 

*Running demo*

Screen Shot 2020-06-24 at 23 23 49


lib/rules/no-multiple-assertions-wait-for.ts
```diff
@@ -0,0 +1,67 @@
...
+            isMemberExpression(node?.expression?.callee) &&
+            isCallExpression(node?.expression?.callee?.object)
+          ) {
+            const object: TSESTree.CallExpression = node?.expression?.callee?.object
+            const expressionName: string = (object?.callee as TSESTree.Identifier)?.name
+            return expressionName === 'expect'
+          } else {
+            return false
+          }
+        })
```
gndelia, line 47: I think here instead of confirming _every_ statement to be a `expect`, we should check filter and check the length is large or equals than two. Otherwise, a `console.log` in-between will make this `every` return `false`. For instance, I think scenario

```ts
waitFor(() => {
  expect(a).toEqual('a')
  console.log('asserting b')
  expect(b).toEqual('b')
})
```

will not be detected by this code

lib/rules/no-multiple-assertions-wait-for.ts
```diff
@@ -0,0 +1,67 @@
...
+          data: {
+            methodName,
+          },
+        });
+      }
+    }
+
+    return {
+      [`${WAIT_EXPRESSION_QUERY} > ArrowFunctionExpression > BlockStatement`]: reporttMultipleAssertion,
+      [`${WAIT_EXPRESSION_QUERY} > FunctionExpression > BlockStatement`]: reporttMultipleAssertion,
```
gndelia, line 64: 💯 

Can we add more scenarios in the test for this code? I would also add them in the `correct` section of the readme of this rule.

docs/rules/no-multiple-assertions-wait-for.md
```diff
@@ -0,0 +1,54 @@
+# Multiple assertions inside `waitFor` are not preferred (no-multiple-assertions-wait-for)
+
+## Rule Details
+
+This rule aims to ensure the correct usage of `expect` inside `waitFor`, in the way that they're intended to be used.
+When using multiples assertions inside `waitFor`, if one fails, you have to wait for timeout before see it failing.
```
timdeschryver, line 6:
```suggestion
When using multiples assertions inside `waitFor`, if one fails, you have to wait for a timeout before seeing it failing.
```

lib/rules/no-multiple-assertions-wait-for.ts
```diff
@@ -0,0 +1,63 @@
+import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils'
+import { getDocsUrl } from '../utils'
+import { isBlockStatement, findClosestCallNode, isMemberExpression, isCallExpression, isIdentifier } from '../node-utils'
+
+export const RULE_NAME = 'no-multiple-expect-wait-for';
```
timdeschryver, line 5: The rulename should match the filename (and docs)

```suggestion
export const RULE_NAME = 'no-multiple-assertions-wait-for';
```

lib/rules/no-multiple-assertions-wait-for.ts
```diff
@@ -0,0 +1,63 @@
...
+    },
+    messages: {
+      noMultipleAssertionWaitFor: 'Avoid use multiple assertions to `waitFor`',
+    },
+    fixable: null,
+    schema: [],
+  },
+  defaultOptions: [],
+  create: function(context) {
+    function reporttMultipleAssertion(
```
gndelia, line 31: There's a minor spelling mistake here
```suggestion
    function reportMultipleAssertion(
```

docs/rules/no-multiple-assertions-wait-for.md
```diff
@@ -0,0 +1,54 @@
...
+  // or
+  await waitFor(function() {
+    expect(a).toEqual('a')
+  })
+};
+```
+
+## Further Reading
+
+- [about `waitFor`](https://testing-library.com/docs/dom-testing-library/api-async#waitfor)
```
Belco90, line 54: Could you add here link to proper inspiration from Common mistakes post please? https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback

docs/rules/no-multiple-assertions-wait-for.md
```diff
@@ -0,0 +1,54 @@
...
+    expect(a).toEqual('a')
+    expect(b).toEqual('b');
+  })
+};
+```
+
+Examples of **correct** code for this rule:
+
+```js
+const foo = async () => {
```
Belco90, line 30: Could you add same examples from incorrect block but corrected please? You are just using correct examples with a single expect, so it would be nice if incorrect examples can be added here showing how it should be fixed.

lib/rules/no-multiple-assertions-wait-for.ts
```diff
@@ -0,0 +1,63 @@
...
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        "It's preferred to avoid multiple assertions in `waitFor`",
+      category: 'Best Practices',
+      recommended: false,
+    },
+    messages: {
+      noMultipleAssertionWaitFor: 'Avoid use multiple assertions to `waitFor`',
```
Belco90, line 24:
```suggestion
      noMultipleAssertionWaitFor: 'Avoid using multiple assertions within `waitFor` callback',
```

Belco90:
@all-contributors please add @renatoagds for ideas, code, doc and test.

PR #190, opened 2020-06-29
feat: add no-node-access rule

### LINKED ISSUE #62: New Rule: Disallow DOM exploration properties
This is not a great way to select elements, but is allowed:
`parentElement.parentElement.parentElement`

It would be great if there were an optional rule that discouraged the use of properties such as `parentElement`, `previousSibling`, and the rest.

The rule could be highly configurable by allowing each property key to be enabled or disabled. It could also allow a certain level of traversing so that one use of a property is okay, but a chain is not.

thebinaryfelix:
Add a new rule `no-node-access` to disallow traversing the DOM using native methods and properties.

Closes #62 

_Demo_

no-node-access


docs/rules/no-node-access.md
```diff
@@ -0,0 +1,44 @@
...
+
+```js
+document.getElementById('submit-btn').closest('button');
+```
+
+Examples of **correct** code for this rule:
+
+```js
+const buttonText = screen.getByRole('button');
+expect(buttonText.textContent).toBe('Submit');
```
timdeschryver, line 35: Should we add an example referring `@testing-library/jest-dom`?

https://github.com/testing-library/jest-dom#tohavetextcontent

tests/lib/rules/no-node-access.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      errors: [
+        {
+          messageId: 'noNodeAccess',
+        },
+      ],
+    },
+    {
+      code: `
+        const buttons = screen.getAllByRole('button');
+        const buttonA = buttons[1];
```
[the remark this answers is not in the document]
thebinaryfelix: Makes sense. Since this rule is already covering methods from `document` that return an HTMLCollection such as `document.getElementsBy...` only arrays returned from Testing Library methods will be allowed anyway, so I can just remove this check.

docs/rules/no-node-access.md
thebinaryfelix: Well, I think it doesn't hurt showing more examples 😃  I would actually prefer to show it instead of the current example

docs/rules/no-node-access.md
```diff
@@ -0,0 +1,44 @@
...
+```
```
gndelia, line 36: I see all the examples uses `screen`. Though it's the recommended way, there's another query that considers that; so for the scope of this one, using destructuring with render should work as well.

Can we add those examples here in the readme, and also in the tests?

thebinaryfelix: Nice! I thought about those and forgot to add them, thanks.

thebinaryfelix:
While adding the new tests with destructuring I've realized that only one scenario wasn't fitting in: The one where the user would create an object with methods having the same name as the ones we get from the Nodes. I might have been overthinking this, trying to cover that scenario. Without it, the rule implementation becomes really very simple. IMO I believe it's a corner case, but I might be missing something.

thebinaryfelix:
Basically now I'm just checking whether the methods called are in the forbidden list or not. Also, when this rule gets upgraded to accept configs and allowed methods, the refactoring should be smooth enough.

Belco90:
Brilliant. Thanks for your effort Mateus ❤️

docs/rules/no-multiple-assertions-wait-for.md
```diff
@@ -30,11 +30,11 @@ Examples of **correct** code for this rule:
...
   await waitFor(() => expect(a).toEqual('a'));
   expect(b).toEqual('b');
-  
+
   // or
   await waitFor(function() {
-    expect(a).toEqual('a')
-  })
+    expect(a).toEqual('a');
+  });
```
gndelia, line 37: I know it's minor, but are we adding this for any particular reason? It's not related to this rule 🤔  but @Belco90  comment makes me thing that perhaps this branch started based off master...? (the one about the readme)

gndelia:
Sorry for the delay. I will try to review it in the incoming days.

Thanks for your effort and the PR 🚀

docs/rules/no-multiple-assertions-wait-for.md
thebinaryfelix: When I did a pull from v4 to fix conflicts (due to this PR entering first #189 ) I might have used my auto format here and it fixes semicolons on `js` snippets. But indeed, I believe I started this branch from the master on my forked repo.

thebinaryfelix: But by all means if you see something wrong in these commits and needs to be fixed, let me know.

gndelia:
> Basically now I'm just checking whether the methods called are in the forbidden list or not. Also, when this rule gets upgraded to accept configs and allowed methods, the refactoring should be smooth enough.

I am ok with making a small rule and upgrade it later to accept configs and allowed methods.

My only concern here is that by just checking the list by name, I think we are basically forbidding any variable with the names of those properties.

For instance, the following case

```js
{
   code: 'const lastChild = foo(baz)'
}
```

where `foo` or `baz` might be any custom function breaks as an invalid scenario. The same applies for any variable name that is listed in [this array](https://github.com/testing-library/eslint-plugin-testing-library/pull/190/files#diff-37a775e32e4a1ab95c3a6ada682775dcR66)  

I think we should narrow a bit more the scope of the implementation to avoid cases like this to fail.   
I'd also like to read your thoughts and of the rest as well. Perhaps it's not worth the effort because this names are so generic that the chances of being used are rare, so it cancan be tackled later.

thebinaryfelix:
> {
   code: 'const lastChild = foo(baz)'
}

@gndelia this scenario wasn't supposed to trigger an error. I've corrected it now.

What I thought was, some code like the following should be corner case and thus we wouldn't need to cover it now:

```
{
  // this will show an error even though "firstChild" here is a property defined by the user
  code: 'myCustomObj.customMethod().firstChild'
}
```

> Perhaps it's not worth the effort because these names are so specific that the chances of being used are rare, so it can be tackled later.

This is exactly what I thought for this later case I mentioned. These names are so specific that might be difficult for a user to create a method / property with a matching name.

docs/rules/no-multiple-assertions-wait-for.md
Belco90: It seems like your autoformatting is updating stuff in other files. Not a big deal, but keep an eye on it as it could lead to other issues in the future.

docs/rules/no-node-access.md
```diff
@@ -0,0 +1,42 @@
...
+
+```js
+document.getElementById('submit-btn').closest('button');
+```
+
+Examples of **correct** code for this rule:
+
+```js
+const button = screen.getByRole('button');
+expect(button).toHaveTextContent('submit');
```
Belco90, line 33: Last example requirement: I think it would be nice to add a more elaborated example using `within` which is a nice way of finding elements from another one.

thebinaryfelix: > Are we waiting then for the checks you mentioned in your last comments

After including this suggestion of using `within` I think we're good to go.

thebinaryfelix:
>I think we should only run this rule if a testing-library package is imported

I've got a ~stupid~ question: shouldn't all the rules run only if a testing-library package is being imported (aka: on a testing environment)?

Aside from that, you're right this is really important for this specific rule because as you stated, if someone writes `props.children` in the source code, it'll trigger an error. I haven't thought about that.

Belco90:
> > I think we should only run this rule if a testing-library package is imported
> 
> I've got a ~stupid~ question: shouldn't all the rules run only if a testing-library package is being imported (aka: on a testing environment)?

Yes, ideally all of them should check that it's a testing library code. But not only if testing library package is imported (e.g. when using cypress testing library you don't import anything in that particular file, or if using a custom render method the user won't import from testing library directly). We should look for a generic mechanism for this, but for now we need to take care of it on each rule individually.

timdeschryver:
I think this rule is a little bit different.
 
The other rules are mainly about an interaction with the testing API (which can be validated if it's a method from a testing library).

While this rule is about accessing properties from the render result, and I'm not sure if we could verify if it comes from testing library. That's why I opted to check if a testing library is imported.

renatoagds:
> We should look for a generic mechanism for this

@Belco90 Something in mind? Tried to think about this and didn't arrive in a specific solution. 

Maybe it's a case to open a specific issue for this? Could be good to map this case (will avoid comments "unrelated" with this PR too)

docs/rules/no-node-access.md
thebinaryfelix: Done! I've also updated the examples and testing cases. I've added a verification on the import declarations to check whether a testing-library package is being imported before triggering an error message.

lib/rules/no-node-access.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  create(context) {
+    let isImportingTestingLibrary = false;
+
+    function checkTestingEnvironment(node: TSESTree.ImportDeclaration) {
+      isImportingTestingLibrary = /testing-library/g.test(node.source.value as string);
+    }
+
+    function showErrorForNodeAccess(node: TSESTree.MemberExpression) {
+      isIdentifier(node.property) &&
```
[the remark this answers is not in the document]
Belco90: Oops sorry! We approved at the same time and I merged it without realizing about this.

Issue #173, opened 2020-06-17
Rules do not work with custom name library name


stefcameron:
I love the `no-debug` rule, however, it doesn't work if you're using a custom local library name in order to, for example, provide a [custom render function](https://testing-library.com/docs/react-testing-library/setup#custom-render) to the rest of your code.

I agree that the library name is no longer `@testing-library/react`, for example, and could be _anything_, so the rule doesn't find a match, but because customizing the render function is in the docs, and I presume quite standard practice, it makes this rule an immediate non-starter, which is a shame since the rule would be really nice to have.

Seems like the rule needs an option, similar to the existing `renderFunctions` option, that we can use to specify the name of our import. In my case, it's `import ... from 'testingUtility'`.

```
import { screen } from '@testing-library/react';
...
screen.debug(); // IS a lint error
```
however,
```
import { screen } from 'testingUtility';
...
screen.debug(); // is NOT a lint error -- and should be
```

nickmccurdy:
Out of curiosity, what's the use case for this? Are you forking the library? Or adding custom types? Or is the intention to support new testing libraries that aren't necessarily in this organization yet?

timdeschryver:
This probably affects most of the (all?) rules 😅

Belco90:
Wow, didn't think about this before. @timdeschryver exacatly, this can potentially affect all the rule. To solve this properly we would need to include a global param in the plugin config so you can set your custom module from where you import Testing Library stuff. Otherwise, we can't know if the `screen` you are importing is from Testing Library or not.

nickmccurdy:
I'm still not sure what this is for. Could we alternatively solve this without any config with some more dynamic logic?

Belco90:
I don't know if it's possible to guess this from our ESLint component. I'll check how other libraries get this kind of dynamic values from custom user files.

timdeschryver:
TIL, there are [shared settings](https://eslint.org/docs/user-guide/configuring#adding-shared-settings) seems like something we can use? Not sure tho...

> ESLint supports adding shared settings into configuration file. You can add settings object to ESLint configuration file and it will be supplied to every rule that will be executed. This may be useful if you are adding custom rules and want them to have access to the same information and be easily configurable.

Belco90:
[...]

PR #192, opened 2020-07-01
[new-rule] prefer-user-event

### LINKED ISSUE #162: New rule: prefer-user-event
When https://github.com/testing-library/dom-testing-library/pull/616 gets merged, `userEvent` will be part of DOM testing library. 

We could create rules for events on `userEvent` that are preferred over their `fireEvent` counterparts.

Before we start implementing rules, there are some thing we need to clarify first:

- Kent already mentioned `click` would be a good candidate, are there any others that could benefit from this rule? I think `hover`, `unhover` could also be included in within this rule.

- Should it be one rule for all events, or a rule per event? The latter would allow us to disable/enable specific events... but we could also create a config for it...

gndelia:
Closes #162 

I reused a function from `no-debug`, so you will notice the change there because I moved  it to `utils`

I also have some doubts about which badges should this rule use: as I understand, `userEvent` is only for `react-testing-library`, so that's the only badge required. 

Working example

image

(My IDE is not configured to run on type 😅 )

This PR does not include the configuration for `custom` import (like mentioned in #173).   
Not sure if that's going to be on `v4`. I did not implement it because I assume we'll do a generic fix for all rules (apparently, the configuration of the plugin solves this as it's mentioned there)

edit: In the end, I removed the methods related to blur and focus as they are not documented/released  in `userEvent` yet


docs/rules/prefer-user-event.md
```diff
@@ -0,0 +1,126 @@
...
+
+// using fireEvent after importing the entire library
+import * as dom from '@testing-library/dom';
+dom.fireEvent.click(node);
+dom.fireEvent.change(node, { target: { value: 'foo' } });
+```
+
+## When Not To Use It
+
+When you don't want to encourage the usage of `userEvent`
```
gndelia, line 92: I did not know what to put here 😅  🤷

nickmccurdy:
> I also have some doubts about which badges should this rule use: as I understand, `userEvent` is only for `react-testing-library`, so that's the only badge required. 

I believe it should have all badges, since it's meant to work with all testing libraries.

docs/rules/prefer-user-event.md
nickmccurdy: Maybe simplify this a little

```suggestion
When you don't want to use `userEvent`, such as if a legacy codebase is still using `fireEvent` or you need to have more low-level control over firing events (rather than the recommended approach of testing from a user's perspective)
```

docs/rules/prefer-user-event.md
```diff
@@ -0,0 +1,126 @@
...
+| `fireEvent.pointerLeave()` | <ul><li>`userEvent.unhover()`</li></ul>                                                                                                                                 |
+| `fireEvent.pointerMove()`  | <ul><li>`userEvent.hover()`</li><li>`userEvent.unhover()`</li><li>`userEvent.selectOptions()`</li><li>`userEvent.deselectOptions()`</li></ul>                           |
+| `fireEvent.pointerOut()`   | <ul><li>`userEvent.unhover()`</li></ul>                                                                                                                                 |
+| `fireEvent.pointerOver()`  | <ul><li>`userEvent.hover()`</li><li>`userEvent.selectOptions()`</li><li>`userEvent.deselectOptions()`</li></ul>                                                         |
+| `fireEvent.pointerUp()`    | <ul><li>`userEvent.click()`</li><li>`userEvent.dblClick()`</li><li>`userEvent.selectOptions()`</li><li>`userEvent.deselectOptions()`</li></ul>                          |
```
nickmccurdy, line 126: I don't think we need to repeat `fireEvent`/`userEvent` and parentheses in this table.

tests/lib/rules/prefer-user-event.test.ts
```diff
@@ -0,0 +1,123 @@
...
+          code: `
+            import * as dom from '${libraryModule}'
+            dom.fireEvent.${fireEventMethod}(foo)
+          `,
+          errors: [{ messageId: 'preferUserEvent' }]
+        }))
+      ),
+    [])
+  ]
+});
```
nickmccurdy, line 123: There's a bunch of repetitive mapping and reducing here, maybe we could extract helper functions

Belco90:
> > I also have some doubts about which badges should this rule use: as I understand, `userEvent` is only for `react-testing-library`, so that's the only badge required.
> 
> I believe it should have all badges and be added to all testing library configs, since it's meant to work with all testing libraries.

I'm not sure we should enable this one in any shareable config tho. It's a good practices to use `userEvent` over `fireEvent` when available, that's a fact. But `user-event` is not included with any testing library framework, so we are not sure if the final user has that method available. Should we enable it anyway to encourage people using it? Or could that cause more issues than it solves?

Belco90:
@gndelia btw remember you are a collaborator now, so you can directly branch off your work within this repo! Not necessary to keep working in your fork anymore.

gndelia:
> > > I also have some doubts about which badges should this rule use: as I understand, `userEvent` is only for `react-testing-library`, so that's the only badge required.
> > 
> > 
> > I believe it should have all badges and be added to all testing library configs, since it's meant to work with all testing libraries.
> 
> I'm not sure we should enable this one in any shareable config tho. It's a good practices to use `userEvent` over `fireEvent` when available, that's a fact. But `user-event` is not included with any testing library framework, so we are not sure if the final user has that method available. Should we enable it anyway to encourage people using it? Or could that cause more issues than it solves?

It is indeed separated from `dom`. I went for React only because 

- it only lists React in the summary

image

- All the examples are of React in the Readme - not `dom` related

[The docs state](https://testing-library.com/docs/ecosystem-user-event)

> user-event is a companion library for **React** Testing Library that provides...


So even though I imagined it could work in other libraries, I wasn't sure how true is that today. Perhaps those docs are misleading? I am not sure



> @gndelia btw remember you are a collaborator now, so you can directly branch off your work within this repo! Not necessary to keep working in your fork anymore.

Lol you're right. I just used the code that I already had locally. Not a big deal though, next PR will come from a branch 😄

nickmccurdy:
> I'm not sure we should enable this one in any shareable config tho. It's a good practices to use `userEvent` over `fireEvent` when available, that's a fact. But `user-event` is not included with any testing library framework, so we are not sure if the final user has that method available. Should we enable it anyway to encourage people using it? Or could that cause more issues than it solves?

My thought is that since `user-event` is meant to represent more realistic user testing behavior, and since it's recommended over `fireEvent` in its own docs, the rule would be a good way to encourage it, even if that means installing another package. We also might still merge it back into `dom-testing-library` eventually.

@kentcdodds What do you think about this? Should the default config error or warn when `user-event` isn't used by default?

> It is indeed separated from `dom`. I went for React only because
> 
> * it only lists React in the summary
> 
> image
> 
> * All the examples are of React in the Readme - not `dom` related
> 
> [The docs state](https://testing-library.com/docs/ecosystem-user-event)
> 
> > user-event is a companion library for **React** Testing Library that provides...
> 
> So even though I imagined it could work in other libraries, I wasn't sure how true is that today. Perhaps those docs are misleading? I don't know
> 
> EDIT: Given it's a separated package, it might confuse user as they cannot import `userEvent` if they don't know in advance about it. Usually the rules are straightforward about "use this", but for them it might be confusing if they don't know about `userEvent` and can't just "import it" because it's not available

Oops, those docs do seem misleading! It's definitely not just for React even if RTL was the initial inspiration. We've actually been porting our tests away from React and are just using plain DOM now to make sure it can work in any testing library.

kentcdodds:
I think it should be either disabled or only warn

kentcdodds:
Actually, what would be coolest is if we detect whether user-event is installed. If it is then make it an error, otherwise disable it.

nickmccurdy:
Personally I'd rather encourage more users to try out `user-event` with a warning since it does solve a lot of problems with `fireEvent` and make testing more convenient, even if it's not ready to be merged yet. What do you think?

kentcdodds:
Yeah, there are trade-offs here. I'm fine with a warning.

gndelia:
So to sum up:

- Set it as a warning
- It should have all the badges (react, angular, cue)

- Should I put the `dom` badge as well, or should we wait for that until `userEvent` gets merged in the `dom` repo ?

tests/lib/rules/prefer-user-event.test.ts
gndelia: done - definitely the scenarios are a bit more readable now. Thanks!

Belco90:
I think you can enable it for `dom` too, as it's based on it.

docs/rules/prefer-user-event.md
```diff
@@ -0,0 +1,126 @@
+# Use [userEvent](https://github.com/testing-library/user-event) over using `fireEvent` for user interactions (prefer-user-event)
+
+From
+[testing-library/dom-testing-library#107](https://github.com/testing-library/dom-testing-library/issues/107):
+
+> [...] it is becoming apparent the need to express user actions on a web page
+> using a higher-level abstraction than `fireEvent`
+
+`userEvent` adds related event calls from browsers to make tests more realistic than its counterpart `fireEvent`, which is a low-level api.  
+See the appendix at the end to check how are the events from `fireEvent` mapped to `userEvent`
```
timdeschryver, line 10:
```suggestion
See the appendix at the end to check how are the events from `fireEvent` mapped to `userEvent`.
```

docs/rules/prefer-user-event.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Rule Details
+
+This rules enforces the usage of [userEvent](https://github.com/testing-library/user-event) methods over `fireEvent`. By default, the methods from `userEvent` takes precedence, but you add exceptions by configuring the rule in `.eslintrc`
```
timdeschryver, line 14:
```suggestion
This rule enforces the usage of [userEvent](https://github.com/testing-library/user-event) methods over `fireEvent`. By default, the methods from `userEvent` take precedence, but you add exceptions by configuring the rule in `.eslintrc`.
```

docs/rules/prefer-user-event.md
```diff
@@ -0,0 +1,126 @@
...
+
+See below for examples of valid usages of `fireEvent` methods with the configuration.
```
timdeschryver, line 16: Personally, I think this line can be removed - or that it can be replaces with a link to the options.
I'm fine either way tho, this was the only comment I would not have said it.
```suggestion
```

docs/rules/prefer-user-event.md
```diff
@@ -0,0 +1,126 @@
...
+import * as dom from '@testing-library/dom';
+dom.fireEvent.click(node);
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+// any userEvent method
+userEvent.click();
+// fireEvent method that does not have an alternative in userEvent
```
timdeschryver, line 39: Sorry, I feel like a jerk to nit about the style, but since this is public I think it should look right 😅

```suggestion

// fireEvent method that does not have an alternative in userEvent
```

docs/rules/prefer-user-event.md
```diff
@@ -0,0 +1,126 @@
...
+fireEvent.cut(node);
+import * as dom from '@testing-library/dom';
```
timdeschryver, line 41:
```suggestion

import * as dom from '@testing-library/dom';
```

docs/rules/prefer-user-event.md
```diff
@@ -0,0 +1,126 @@
...
+
+// using fireEvent after importing the entire library
+import * as dom from '@testing-library/dom';
+dom.fireEvent.click(node);
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+// any userEvent method
```
timdeschryver, line 37: Because most samples include an import, I would add it here too:

```suggestion
import userEvent from '@testing-library/user-event';
// any userEvent method
```

docs/rules/prefer-user-event.md
```diff
@@ -0,0 +1,126 @@
...
+// fireEvent method that does not have an alternative in userEvent
+fireEvent.cut(node);
+import * as dom from '@testing-library/dom';
+dom.fireEvent.cut(node);
+
+// a function called fireEvent that's not imported from testing-library
+function fireEvent() {
+  // do stuff
+}
+fireEvent();
```
timdeschryver, line 48: I think we can assume this is the case for most of the rules, thus I would remove it.
It sure is important for us (in our tests), but I don't see an added value to add it to the docs.

```suggestion
```

gndelia: iirc it was just for the sake of coverage, otherwise, I wouldn't get the 100%... but I changed so many times the tests that I'll review it again to see if I can remove it.

timdeschryver: It sure can exists in the specs.
But it shouldn't live in the docs, imho.

gndelia: oh lol, I just reviewed the comment from the emails and I thought you were commenting in the test, not in the docs. Yeah, I think that makes sense. I'll remove it

lib/rules/prefer-user-event.ts
```diff
@@ -0,0 +1,138 @@
...
+function buildErrorMessage(fireEventMethod: string) {
+  const allMethods = MappingToUserEvent[fireEventMethod].map((method: string) => `userEvent.${method}()`)
+
+  let userEventMethods = ''
+  if (allMethods.length > 2) {
+    userEventMethods += allMethods.slice(0, allMethods.length - 2).join(', ')
+  }
+
+  userEventMethods += `${allMethods.length > 1 ? ' or ' : ''}${allMethods[allMethods.length - 1]}`
+  return userEventMethods
```
nickmccurdy, line 48: I think this is easier to read with variables extracted.
```suggestion
  const { length } = allMethods

  const init = length > 2 ? allMethods.slice(0, length - 2).join(', ') : ''
  const last = `${length > 1 ? ' or ' : ''}${allMethods[length - 1]}`
  return `${init}${last}`
```

Feel free to refactor more if you want, I may be missing something.

gndelia:
@nickmccurdy @timdeschryver  all changes applied. Thanks for the suggestions!

README.md
```diff
@@ -125,25 +125,26 @@ To enable this configuration use the `extends` property in your
...
+| [prefer-find-by](docs/rules/prefer-find-by.md)                                   | Suggest using `findBy*` methods instead of the `waitFor` + `getBy` queries               | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] | ![fixable-badge][] |
+| [prefer-presence-queries](docs/rules/prefer-presence-queries.md)                 | Enforce specific queries when checking element is present or not                         |                                                                   |                    |
+| [prefer-screen-queries](docs/rules/prefer-screen-queries.md)                     | Suggest using screen while using queries                                                 | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |                    |
+| [prefer-user-event](docs/rules/prefer-user-event.md)                             | Suggest using `userEvent` library instead of `fireEvent` for simulating user interaction | ![angular-badge][] ![react-badge][] ![vue-badge][]                |                    |
```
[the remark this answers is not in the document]
gndelia: added the dom badge. Not quite sure what you mean by `config updates`  😅 . I updated the rule readme with all the suggestions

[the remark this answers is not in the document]
nickmccurdy: I meant adding it to the DOM config. It looks like you already added it and merged it into the other configs though, thanks!

Belco90:
@gndelia You have some conflicts with v4 branch. Can you resolve them please? I'll merge it right after that.

gndelia:
@Belco90 ✔️

j10sanders:
Should this be added to the docs? Thanks

PR #139, opened 2020-05-17
feat: Support ESLint 7.x

MichaelDeBoey:
ESLint v7.0.0 is [released](https://eslint.org/blog/2020/05/eslint-v7.0.0-released) 🎉

(dev)Dependencies should be compatible with ESLint 7 too before we can merge this one:

- [x] [`@typescript-eslint/experimental-utils`](https://github.com/typescript-eslint/typescript-eslint/issues/1550)
  - [x]  [Release](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v3.0.0)
- [x] [`@typescript-eslint/eslint-plugin`](https://github.com/typescript-eslint/typescript-eslint/issues/1550)
  - [x]  [Release](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v3.0.0)
- [x] [`@typescript-eslint/parser`](https://github.com/typescript-eslint/typescript-eslint/issues/1550)
  - [x]  [Release](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v3.0.0)
- [x] [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import/issues/1772)
  - [x]  [PR](https://github.com/benmosher/eslint-plugin-import/pull/1715) + [commit](https://github.com/benmosher/eslint-plugin-import/commit/6a110dd16a7fd775f08601054bf14ffd503eea7b)
  - [x]  [Release](https://github.com/benmosher/eslint-plugin-import/blob/master/CHANGELOG.md#2211---2020-06-07)
- [x] [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest/issues/568)
  - [x]  [PR](https://github.com/jest-community/eslint-plugin-jest/pull/572)
  - [x]  Released
- [x] [`eslint-plugin-jest-formatting`](https://github.com/dangreenisrael/eslint-plugin-jest-formatting/issues/88)
  - [x]  [PR](https://github.com/dangreenisrael/eslint-plugin-jest-formatting/pull/89)
  - [x]  [Release](https://github.com/dangreenisrael/eslint-plugin-jest-formatting/releases/tag/v2.0.0)
- [ ] [`eslint-plugin-node`](https://github.com/mysticatea/eslint-plugin-node/issues/223)
  - [ ]  [PR](https://github.com/mysticatea/eslint-plugin-node/pull/224)
  - [ ]  Released
[...]

PR #194, opened 2020-07-06
ESLint 7 support

nickmccurdy:
@SimenB Hope this looks good. I don't know much about what else we may have to change for ESLint 7, but everything seems to be working in CI.

I also opened up the version ranges using `>=5`, that seems to be the preference of both ESLint and Kent.


nickmccurdy:
Sorry, I completely forgot about #139 🤦‍♀️  it already has all these changes along with tracking required updates to dependencies

PR #196, opened 2020-07-08
feat: add rule no-side-effects-wait-for

### LINKED ISSUE #134: new rule: no-side-effects-wait-for

renatoagds:
Add a new rule, based in a tip from [`common-mistakes`](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#performing-side-effects-in-waitfor), to avoid side effects inside a `waitFor`.

Closes #134 

*Running demo*

Screen Shot 2020-07-08 at 01 09 11


docs/rules/no-side-effects-wait-for.md
```diff
@@ -0,0 +1,73 @@
+# Side effects inside `waitFor` are not preferred (no-side-effects-wait-for)
+
+## Rule Details
+
+This rule aims to avoid the usage of side effects actions (`fireEvent` or `useEvent`) inside `waitFor`.
```
gndelia, line 5:
```suggestion
This rule aims to avoid the usage of side effects actions (`fireEvent` or `userEvent`) inside `waitFor`.
```

docs/rules/no-side-effects-wait-for.md
```diff
@@ -0,0 +1,73 @@
...
+    expect(b).toEqual('b');
+  });
+
+  // or
+  userEvent.keyDown(input, { key: 'ArrowDown' });
+  await waitFor(function() {
+    expect(b).toEqual('b');
+  });
+};
+```
```
gndelia, line 66: I am a bit confused about the 
```ts
const foo = async () => {
```

parts. I would just remove them and leave the examples directly.

perhaps I would also add a small comment explaining what the snippet does (like `// using arrow function`) or similar to be more consistent with other rules (not all, but many do this)

docs/rules/no-side-effects-wait-for.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Further Reading
+
+- [about `waitFor`](https://testing-library.com/docs/dom-testing-library/api-async#waitfor)
+- [about `userEvent`](https://github.com/testing-library/user-event)
+- [about `fireEvent`](https://testing-library.com/docs/dom-testing-library/api-events)
+- [inspiration for this rule](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#performing-side-effects-in-waitfor)
```
gndelia, line 73: 💯

lib/rules/no-side-effects-wait-for.ts
```diff
@@ -0,0 +1,65 @@
...
+    fixable: null,
+    schema: [],
+  },
+  defaultOptions: [],
+  create: function(context) {
+    function reportSideEffects(
+      node: TSESTree.BlockStatement
+    ) {
+      const totalSideEffects = (body: Array<TSESTree.Node>): Array<TSESTree.Node> =>
+        body.filter((node: TSESTree.ExpressionStatement) => {
```
gndelia, line 37: we could probably use `some()` here, which already returns a boolean, and remove the check for `.length > 0` in line 51. Because of that, the method name could be changed to something like `hasSideEffects` or similar

gndelia:
I added some comments. Thanks for your PR ❤️ 🚀 

In addition to these comments, I think we should probably check that RTL is imported.

docs/rules/no-side-effects-wait-for.md
renatoagds: Ops! This is something extra here 😅

lib/rules/no-side-effects-wait-for.ts
```diff
@@ -60,6 +62,9 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
     return {
       [`${WAIT_EXPRESSION_QUERY} > ArrowFunctionExpression > BlockStatement`]: reportSideEffects,
       [`${WAIT_EXPRESSION_QUERY} > FunctionExpression > BlockStatement`]: reportSideEffects,
+      ImportDeclaration(node: TSESTree.ImportDeclaration) {
+        isImportingTestingLibrary = /testing-library/g.test(node.source.value as string);
```
Belco90, line 66: You already have `hasTestingLibraryImportModule` util for checking this. Could you use that one instead please?

Belco90, line 66: You already have hasTestingLibraryImportModule util for checking this. Could you use that one instead please?

tests/lib/rules/no-side-effects-wait-for.test.ts
```diff
@@ -0,0 +1,210 @@
...
+  ecmaFeatures: {
+    jsx: true,
+  },
+});
+
+ruleTester.run(RULE_NAME, rule, {
+  valid: [
+    {
+      code: `
+        import { waitFor } from '@testing-library/react';
```
Belco90, line 14: Could you add some extra test to check that a `waitFor` imported from a module not related to testing-library is always valid please? Something like:

```javascript
import { waitFor } from '@somewhere/else';

// valid as this waitFor is not related to Testing Library
await waitFor(() => {
  expect(b).toEqual('b')
  fireEvent.keyDown(input, {key: 'ArrowDown'})
});
```

docs/rules/no-side-effects-wait-for.md
```diff
@@ -0,0 +1,71 @@
...
+    userEvent.keyDown(input, { key: 'ArrowDown' });
+    expect(b).toEqual('b');
+  });
+
+  // or
+  await waitFor(function() {
+    userEvent.keyDown(input, { key: 'ArrowDown' });
+    expect(b).toEqual('b');
+  });
+};
```
gndelia, line 35: a minor comment just to make the examples more realistic: there's no `keyDown` method in `userEvent` (that one belongs to `fireEvent`). 

You could replace it with some other method, such as `click`, `type` or `hover`. The same applies for the "correct" section

tests/lib/rules/no-side-effects-wait-for.test.ts
```diff
@@ -0,0 +1,210 @@
...
+        })
+      `,
+      errors: [{ messageId: 'noSideEffectsWaitFor' }]
+    },
+    // userEvent
+    {
+      code: `
+        import { waitFor } from '@testing-library/react';  
+        await waitFor(() => {
+          userEvent.keyDown(input, {key: 'ArrowDown'})
```
gndelia, line 155: same in these comments - just to make them more realistic, let's better use the methods from userEvent API

gndelia:
I added a few more minor comments (about the docs and the tests), but once those and the ones that Mario did are completed, I think we are good to go. 🍡

renatoagds:
@Belco90 done! conflicts fixed 🤘

PR #197, opened 2020-07-15
fix: add findby variable declaration to prefer-find-by when auto fixing

### LINKED ISSUE #167: [prefer-find-by] When query methods come from destructuring, findBy* might not be defined
Example code **before** running the linter

```ts
const { getByText } = render(<Foo />)
await waitFor(() => getByText('foo'))
```

**After** automatically fixing this, the code looks like this

```ts
const { getByText } = render(<Foo />)
await findByText('foo')
```

There are a couple of issues:

1. As the title mentions, `findByText` is not defined. If the code was using `screen` or any similar wrapper, instead of the bound function directly, this is not a problem.
2. If `getByText` is not used elsewhere, it becomes unused, causing _another_ error in the linter that wasn't there before - we've introduced this error
3. Something similar might happen with `waitFor` - it might now be unused as well

We should definitely fix 1. as it otherwise it's a broken code. I guess I could track where `getByText` was defined, and add the `findBy*` function there.

Thoughts on 2. and 3 ?

gndelia:
This PR fixes #167 

The following scenario

```ts
// incorrect
const { getByText } = render(<Foo />)
await waitFor(() => getByText('foo'))
```

now becomes
```ts
// correct - and findByText is defined
const { getByText, findByText } = render(<Foo />)
await findByText('foo')
```

this should even work if they are in different levels. For instance

```ts
// incorrect
const { getByText } = render(<Foo />)
it('tests', async () => {
   await waitFor(() => getByText('foo'))
})
```

should be converted to

```ts
// correct - findByText is defined
// as long as the customRender is in any parent scope from the original code, this works
const { getByText, findByText } = customRender(<Foo />)
it('tests', async () => {
     await findByText('foo')
})
```

there are some scenarios that I assume they are out of scope, and I am seriously not sure how to fix them. For instance, if the query is passed from a function, like 

```ts
function commonCode(getByText) {
  await waitFor(() => getByText(baz))
// do stuff
}

// later, in a test
commonCode(render().getByText)
```

or similar, I can't fix it.. but I guess that sounds like a non-realistic scenario.

Another that _might_ be slightly more realistic (or actually, less complex) might be something like

```ts
const utils= render(baz)
const getByText = utils.getByText
await waitFor(() => getByText('foo'))
```

I am not fixing that either. I'd assume if they are storing the result from the `render`, then they are using that variable. These snippets seem mostly hacks to avoid the rule, rather than code that someone would write on its own.

and a demo gif!

(The spaces are due to other eslint rules I've configured on the project I tested this 👀 )

P.S: I'm not using the fork 🎉

### DIFF
--- a/tests/lib/rules/prefer-find-by.test.ts
+++ b/tests/lib/rules/prefer-find-by.test.ts
@@ -1,7 +1,7 @@
-import { InvalidTestCase } from '@typescript-eslint/experimental-utils/dist/ts-eslint'
+import { InvalidTestCase, ValidTestCase } from '@typescript-eslint/experimental-utils/dist/ts-eslint'
 import { createRuleTester } from '../test-utils';
 import { ASYNC_QUERIES_COMBINATIONS, SYNC_QUERIES_COMBINATIONS } from '../../../lib/utils';
-import rule, { WAIT_METHODS, RULE_NAME } from '../../../lib/rules/prefer-find-by';
+import rule, { WAIT_METHODS, RULE_NAME, getFindByQueryVariant, MessageIds } from '../../../lib/rules/prefer-find-by';
 
 const ruleTester = createRuleTester({
   ecmaFeatures: {
@@ -9,10 +9,26 @@ const ruleTester = createRuleTester({
   },
 });
 
+function buildFindByMethod(queryMethod: string) {
+  return `${getFindByQueryVariant(queryMethod)}${queryMethod.split('By')[1]}`
+}
+
+function createScenario<T extends ValidTestCase<[]> | InvalidTestCase<MessageIds, []>>(callback: (waitMethod: string, queryMethod: string) => T) {
+  return WAIT_METHODS.reduce((acc: T[], waitMethod) => 
+    acc.concat(
+      SYNC_QUERIES_COMBINATIONS
+        .map((queryMethod) => callback(waitMethod, queryMethod))
+    )
+  , [])
+}
+
 ruleTester.run(RULE_NAME, rule, {
   valid: [
     ...ASYNC_QUERIES_COMBINATIONS.map((queryMethod) => ({
-      code: `const submitButton = await ${queryMethod}('foo')`
+      code: `
+        const { ${queryMethod} } = setup()
+        const submitButton = await ${queryMethod}('foo')
+      `
     })),
     ...ASYNC_QUERIES_COMBINATIONS.map((queryMethod) => ({
       code: `const submitButton = await screen.${queryMethod}('foo')`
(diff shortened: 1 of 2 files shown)


tests/lib/rules/prefer-find-by.test.ts
```diff
@@ -60,35 +76,117 @@ ruleTester.run(RULE_NAME, rule, {
...
+        messageId: 'preferFindBy',
+        data: {
+          queryVariant: 'findBy',
+          queryMethod: 'Role',
+          fullQuery: `waitFor(() => getByRole('baz', { name: 'button' }))`
+        }
+      }],
+      output: `
+        const getByRole = render().getByRole
+        const submitButton = await findByRole('baz', { name: 'button' })
```
timdeschryver, line 188: This would turn runnable code into breaking code, right?
Because `findByRole` isn't defined.

If that's the case I would consider this a bad experience as it could potentially break a lot of code.

gndelia: it would.
but it's was already hapenning before this PR. There are a couple of scenarios where the fix breaks.
This PR fixes one of them. This test you've commented is still unfixed. Not quite sure if we can't fix it, as it seems there are multiple scenarios to fix (like I've mentioned in the PR description)

Maybe we went too far by making this rule autofixable? Not sure.

gndelia: In addition to that, I am not sure how often that would be a realistic scenario. I'd think that devs use destructuring or call the function from screen  or the object returned from the render function. But not quite sure. I'm open to debate it.

I still want to enforce this PR _does not introduce_ this scenario, but rather, fixes one of the many possible scenarios where `findBy*` method is added and might not be defined. the scenarios were there already, and the bug mentioned one of them (the one I've fixed)

timdeschryver: Got it, thanks for the clarification.
For the record, I'm not against it.

I do agree that most devs will use `screen`, especially since we're recommended it.

Belco90: I hope no one uses Testing Library this way to be honest 😅  I think we are fine by covering all the other cases.

tests/lib/rules/prefer-find-by.test.ts
```diff
@@ -1,18 +1,34 @@
...
+import rule, { WAIT_METHODS, RULE_NAME, getFindByQueryVariant, MessageIds } from '../../../lib/rules/prefer-find-by';
 
 const ruleTester = createRuleTester({
   ecmaFeatures: {
     jsx: true,
   },
 });
 
+function buildFindByMethod(queryMethod: string) {
+  return `${getFindByQueryVariant(queryMethod)}${queryMethod.split('By')[1]}`
```
Belco90, line 13: I'm not a fan of doing this kind of "importing the function calculating something to check the same in the tests". This can lead to silent errors, so I prefer to hardcode or reimplement this kind of behaviors in testing side. Anyway, I think for this case it's fine.

gndelia: I generally, agree, but I was using that function everywhere and it was so small... I couldn't resist the temptation (?)

Issue #198, opened 2020-07-20
Move rules settings to ESLint shared config


Belco90:
As discussed in different issues and PRs, we need to find a better approach for giving the user the possibility to define which custom render methods they use for Testing Library, or which custom modules they are importing render method.

This change will imply some breaking changes, so v4 would be the perfect opportunity to address this.

The idea is to move specific rules options duplicated across the plugin to [ESLint shared settings](https://eslint.org/docs/user-guide/configuring#adding-shared-settings).

> You can add settings object to ESLint configuration file and it will be supplied to every rule that will be executed. This may be useful if you are adding custom rules and want them to have access to the same information and be easily configurable.

For that reason, this seems the proper approach for avoid duplicating rules configs.

### What things do we need to setup there?
0. As pre-requisite we need to find a way to hook enhanced rules listener with detection helpers for all rules implementations. This has two purposes: extract the common checks (e.g. is imported from Testing Library? is a valid filename?) into a shared create rule function which already prevent rules to be executed if common checks are not met (this allows rules to focus on particular checks for the rule itself without worrying about global checks), and receive a set of detection helpers to detect Testing Library utils like "isQuery" or "isRender" (this allows rules to detect Testing Library patterns easier, reducing duplicated code, complexity and errors).

1. New shared setting to setup a custom module where Testing Libraries can be imported from other than the official package itself. There is [an example here in Testing Library docs](https://testing-library.com/docs/react-testing-library/setup#custom-render). + "aggressive reporting" by default when no custom module set + add corresponding detection helper
```javascript
// render method from custom module
import { render } from 'test-utils';
// ...
const view = render(<Foo />) // corresponding rules will get setting to check if this render method has to be reported
```

2. ~~New shared setting to setup the name pattern of the files desired to be analyzed (by default with same config from jest) + add corresponding detection helper~~ **THIS WAS FINALLY DISCARDED**
```javascript
// assume the user config the plugin to just report errors from files named *.test.js

// foo.spec.js
screen.findByText('something'); // <-- we don't report anything here since the file name doesn't match with desired filename pattern
```

3. Moving current `renderFunctions` option from existing rules to shared setting, so users can config there any custom render function they want to be considered consider by eslint-plugin-testing-library to be reported + add corresponding detection helper:
```javascript
// custom render method from outside
import { renderWithTheme } from 'test-utils';
// ...
[...]

PR #200, opened 2020-07-21
feat: add render-result-naming-convention rule

### LINKED ISSUE #170: new rule: render-result-naming-convention
__Inspiration:__ [Using wrapper as the variable name for the return value from render](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-wrapper-as-the-variable-name-for-the-return-value-from-render)

If possible, it would be nice to extend or reuse the existing core rule [`prefer-destructuring`](https://eslint.org/docs/rules/prefer-destructuring) so it only applies to `wrapper` or return values of `render`.

Belco90:
Closes #170


tests/lib/rules/render-result-naming-convention.test.ts
```diff
@@ -0,0 +1,306 @@
...
+        
+        const setup = () => {
+          // this one must have a valid name
+          const view = render(<SomeComponent />);
+          return view;
+        };
+
+        test('should not report render result called "view" from wrapping function', async () => {
+          // this isn't a render technically so it can be called "wrapper"
+          const wrapper = setup();
```
Belco90, line 169: I'm not sure if this should be valid or not. Should the rule just check whatever comes directly from `render`? Should check also return value from functions wrapping `render` too?

docs/rules/render-result-naming-convention.md
```diff
@@ -0,0 +1,77 @@
+# Enforce a valid naming for return value from `render` (render-result-naming-convention)
+
+> The name `wrapper` is old cruft from `enzyme` and we don't need that here. The return value from `render` is not "wrapping" anything. It's simply a collection of utilities that you should actually not often need anyway.
+
+## Rule Details
+
+This rule aims to ensure the return value from `render` is named properly.
+
+Ideally, you should destructure what you need from `render`. In case you need to save the collection of utils returned in a variable, its name should be on of `view` or `utils`, as `render` is not wrapping anything: it's just returning a collection of utilities. Every other name for that variable will be considered invalid.
```
gndelia, line 9: Perhaps we could also hint and link the `prefer-screen-queries` rule

lib/rules/render-result-naming-convention.ts
```diff
@@ -0,0 +1,101 @@
...
+const ALLOWED_VAR_NAMES_TEXT = ALLOWED_VAR_NAMES.map(
+  name => '`' + name + '`'
+).join(', ');
+
+export default ESLintUtils.RuleCreator(getDocsUrl)({
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'TODO',
```
gndelia, line 23: 👀

tests/lib/rules/render-result-naming-convention.test.ts
```diff
@@ -0,0 +1,306 @@
...
+ruleTester.run(RULE_NAME, rule, {
+  valid: [
+    {
+      code: `
+        import { render } from '@testing-library/react';
+        
+        test('should not report straight destructured render result', () => {
+          const { rerender, getByText } = render(<SomeComponent />);
+          const button = getByText('some button');
+        });
```
gndelia, line 21: 💯  very complete tests

tests/lib/rules/render-result-naming-convention.test.ts
gndelia: Ideally, If `setup` was in another file, it would qualify as a `customRender` function (if configured)- is that correct? We should probably apply the same criteria.

However, in a more realistic way (?), I am not sure if you can easily ensure that whatever was returned from render is returned to the outside in this function where `render` is called. Consider the following scenario

```js
const commonLogic = async (param) => {
   // this is fine according to the rule
   const view = render(<Foo prop1={param} />
   await userEvent.click(view.getByRole('button')
   // might be more logic here
   return view.getByTestId('box').textContent
}

it('something1', () => {
  // this should be fine - a render is inside and we don't return anything render related
  const result = commonLogic(param1)
  expect(result).toBe('foo') 
})

it('something2', () => {
  // same here
  const result = commonLogic(param2)
  expect(result).toBe('baz') 
})

```

here, `commonTest` is a function that has a `render` but it does not wrap it, it just avoids duplication of some logic.

Can we differentiate this scenario from the one in your test?

tests/lib/rules/render-result-naming-convention.test.ts
```diff
@@ -0,0 +1,306 @@
...
+        {
+          messageId: 'invalidRenderResultName',
+          data: {
+            varName: 'wrapper',
+          },
+          line: 5,
+          column: 17,
+        },
+      ],
+    },
```
gndelia, line 195: 🏅  very complete check of the errors, 🚀

lib/rules/render-result-naming-convention.ts
```diff
@@ -0,0 +1,101 @@
...
+    },
+  ],
+
+  create(context, [options]) {
+    const { renderFunctions } = options;
+    let renderResultName: string | null = null;
+    let renderAlias: string | undefined;
+
+    return {
+      ImportDeclaration(node: TSESTree.ImportDeclaration) {
```
gndelia, line 54: should we also consider when importing in the following format?

```ts
import * as dom from '@testing-library/react'
```

Some of the rules check that.

tests/lib/rules/render-result-naming-convention.test.ts
Belco90: > Ideally, If setup was in another file, it would qualify as a customRender function (if configured)- is that correct? 

That's correct. It would also qualify as a render function to be checked if added to `renderFunctions` rule args (which is what I do on my projects as I tend to have tons of `setup` inline wrappers).

I'm worried about exactly that example you wrote. I don't see an easy way to spot that, and I'm not sure if should be reported either. Maybe a nice to have for a future version?

docs/rules/render-result-naming-convention.md
Belco90: Nice suggestion, I'll add it later.

lib/rules/render-result-naming-convention.ts
Belco90: Oops, sorry about this.

tests/lib/rules/render-result-naming-convention.test.ts
Belco90: Thanks 😄 

I've used the test title within the code for giving each test an actual description. I'd love ESLint to provide a way for adding descriptions or titles to each tests. It's really painful to find the failing tests only from their code.

lib/rules/render-result-naming-convention.ts
Belco90: Ah of course! I forgot about that variant. I'll add this later.

tests/lib/rules/render-result-naming-convention.test.ts
gndelia: > Maybe a nice to have for a future version?

yeah, I agree 👍 . For now, being careful of not causing false positives in these scenarios should be enough

gndelia:
LGTM

For the custom render scenarios, like
```ts
import { CustomRender } from 'foo-baz';
```
the `foo-baz` part will be covered in #198 is that correct?

docs/rules/render-result-naming-convention.md
```diff
@@ -0,0 +1,77 @@
...
+
+// ...
+
+// keeping return value from `render` in a var called "utils" is correct
+const utils = render(<SomeComponent />);
+```
+
+## Further Reading
+
+- [Common Mistakes with React Testing Library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-wrapper-as-the-variable-name-for-the-return-value-from-render)
```
renatoagds, line 77: Could be good to add reference for render doc: https://testing-library.com/docs/react-testing-library/api#render

lib/index.ts
```diff
@@ -15,8 +15,9 @@ import preferPresenceQueries from './rules/prefer-presence-queries';
 import preferScreenQueries from './rules/prefer-screen-queries';
 import preferUserEvent from './rules/prefer-user-event';
 import preferWaitFor from './rules/prefer-wait-for';
-import noMultipleAssertionsWaitFor from './rules/no-multiple-assertions-wait-for'
+import noMultipleAssertionsWaitFor from './rules/no-multiple-assertions-wait-for';
 import preferFindBy from './rules/prefer-find-by';
+import invalidRenderResultName from './rules/render-result-naming-convention';
```
renatoagds, line 20: Looking at the other rules, all them are following the pattern to name the import as the rule name. Any reason to change here? Maybe could be good to keep the same pattern

Belco90:
> LGTM
> 
> For the custom render scenarios, like
> 
> ```ts
> import { CustomRender } from 'foo-baz';
> ```
> 
> the `foo-baz` part will be covered in #198 is that correct?

Yes, that should be covered there as the option for config custom testing library modules must be added before checking this.

lib/index.ts
Belco90: Oops, I think I changed this one in a rename that shouldn't affect it. I'll change it.

PR #201, opened 2020-07-26
feat: allow baseElement and container in prefer-screen-queries

### LINKED ISSUE #188: prefer-screen-queries: Don't warn if using the baseElement option
This should be acceptable (from the React Testing Library tests):

```javascript
test('baseElement isolates trees from one another', () => {
  const {getByText: getByTextInA} = render(<div>Jekyll</div>, {
    baseElement: treeA,
  })
  const {getByText: getByTextInB} = render(<div>Hyde</div>, {
    baseElement: treeB,
  })

  expect(() => getByTextInA('Jekyll')).not.toThrow(
    'Unable to find an element with the text: Jekyll.',
  )
  expect(() => getByTextInB('Jekyll')).toThrow(
    'Unable to find an element with the text: Jekyll.',
  )

  expect(() => getByTextInA('Hyde')).toThrow(
    'Unable to find an element with the text: Hyde.',
  )
  expect(() => getByTextInB('Hyde')).not.toThrow(
    'Unable to find an element with the text: Hyde.',
  )
})
```

gndelia:
closes #188 

One thing I noticed from this rule is that it does not check `screen` is imported from RTL, nor anything about custom renders. Not sure if we should all of that for the v4 in #198 or if checking imports should be done in a separated ticket for the current version and then enhanced with custom renders for the v4... I'll assume it will be easier to do everything directly in v4 to avoid doing weird rebases but I want to know your opinion folks.

For this PR I just stick with the feature requested

### DIFF
--- a/docs/rules/prefer-screen-queries.md
+++ b/docs/rules/prefer-screen-queries.md
@@ -60,6 +60,11 @@ const { rerender, unmount, asFragment } = render(<Foo />);
 rerender(<Foo />);
 asFragment();
 unmount();
+
+// using baseElement
+const { getByText } = render(<Foo />, { baseElement: treeA });
+// using container
+const { getAllByText } = render(<Foo />, { container: treeA });
 ```
 
 ## Further Reading
--- a/lib/node-utils.ts
+++ b/lib/node-utils.ts
@@ -1,4 +1,4 @@
-import { TSESTree } from '@typescript-eslint/experimental-utils';
+import { TSESTree, AST_NODE_TYPES } from '@typescript-eslint/experimental-utils';
 
 export function isCallExpression(
   node: TSESTree.Node
@@ -105,4 +105,8 @@ export function hasThenProperty(node: TSESTree.Node) {
 
 export function isArrowFunctionExpression(node: TSESTree.Node): node is TSESTree.ArrowFunctionExpression {
   return node && node.type === 'ArrowFunctionExpression'
+}
+
+export function isObjectExpression(node: TSESTree.Expression): node is TSESTree.ObjectExpression {
+  return node?.type === AST_NODE_TYPES.ObjectExpression
 }
--- a/lib/rules/prefer-screen-queries.ts
+++ b/lib/rules/prefer-screen-queries.ts
@@ -6,13 +6,20 @@ import {
   isCallExpression,
   isProperty,
   isIdentifier,
+  isObjectExpression,
 } from '../node-utils';
 
 export const RULE_NAME = 'prefer-screen-queries';
 export type MessageIds = 'preferScreenQueries';
 type Options = [];
 
+const ALLOWED_RENDER_PROPERTIES_FOR_DESTRUCTURING = ['container', 'baseElement']
 const ALL_QUERIES_COMBINATIONS_REGEXP = ALL_QUERIES_COMBINATIONS.join('|');
+
+function usesContainerOrBaseElement(node: TSESTree.CallExpression) {
+  const secondArgument = node.arguments[1]
+  return isObjectExpression(secondArgument) && secondArgument.properties.some((property) => isProperty(property) && isIdentifier(property.key) && ALLOWED_RENDER_PROPERTIES_FOR_DESTRUCTURING.includes(property.key.name))
+}
 
 export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
   name: RULE_NAME,
@@ -50,9 +57,14 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
 
     return {
       VariableDeclarator(node) {
-        const isWithinFunction = isCallExpression(node.init) && isIdentifier(node.init.callee) && node.init.callee.name === 'within';
+        if (!isCallExpression(node.init) || !isIdentifier(node.init.callee)) {
+          return
+        }
+        const isWithinFunction =  node.init.callee.name === 'within';
+        // TODO add the custom render option #198
+        const usesRenderOptions = node.init.callee.name === 'render' && usesContainerOrBaseElement(node.init);
 
-        if (!isWithinFunction) {
+        if (!isWithinFunction && !usesRenderOptions) {
           return
         }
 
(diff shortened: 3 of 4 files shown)


lib/node-utils.ts
```diff
@@ -105,4 +105,8 @@ export function hasThenProperty(node: TSESTree.Node) {
 
 export function isArrowFunctionExpression(node: TSESTree.Node): node is TSESTree.ArrowFunctionExpression {
   return node && node.type === 'ArrowFunctionExpression'
+}
+
+export function isObjectExpression(node: TSESTree.Expression): node is TSESTree.ObjectExpression {
+  return node?.type === AST_NODE_TYPES.ObjectExpression
```
gndelia, line 111: I found this `enum` which contains all the types we've been hardcoding in this utils function. 

image


Perhaps we could do a small technical ticket to refactor these.

lib/rules/prefer-screen-queries.ts
```diff
@@ -50,9 +52,14 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
 
     return {
       VariableDeclarator(node) {
-        const isWithinFunction = isCallExpression(node.init) && isIdentifier(node.init.callee) && node.init.callee.name === 'within';
+        if (!isCallExpression(node.init) || !isIdentifier(node.init.callee)) {
+          return
+        }
+        const isWithinFunction =  node.init.callee.name === 'within';
+        // TODO add the custom render option #198
+        const usesRenderOptions = node.init.callee.name === 'render' && usesContainerOrBaseElement(node.init);
```
gndelia, line 60: I left this comment as a reminder we need to address custom renders here 😅

lib/rules/prefer-screen-queries.ts
```diff
@@ -110,3 +119,8 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
     };
   },
 });
+
+function usesContainerOrBaseElement(node: TSESTree.CallExpression) {
+  const secondArgument = node.arguments[1]
+  return isObjectExpression(secondArgument) && secondArgument.properties.some((property) => isProperty(property) && isIdentifier(property.key) && ALLOWED_RENDER_PROPERTIES_FOR_DESTRUCTURING.includes(property.key.name))
+}
```
renatoagds, line 126: Point of discussion: This should not be declared in top of the file? I don't know if we're following some pattern, but generally I prefer to put all utilities before default export

lib/node-utils.ts
timdeschryver: Yes! I like that idea!

docs/rules/prefer-screen-queries.md
```diff
@@ -60,6 +60,11 @@ const { rerender, unmount, asFragment } = render(<Foo />);
 rerender(<Foo />);
 asFragment();
 unmount();
+
+// using baseElement
+const { getByText } = render(<Foo />), { baseElement: treeA }
```
timdeschryver, line 65:
```suggestion
const { getByText } = render(<Foo />, { baseElement: treeA })
```

docs/rules/prefer-screen-queries.md
```diff
@@ -60,6 +60,11 @@ const { rerender, unmount, asFragment } = render(<Foo />);
...
+// using container
+const { getAllByText } = render(<Foo />), { baseElement: treeA }
```
timdeschryver, line 67:
```suggestion
const { getAllByText } = render(<Foo />, { baseElement: treeA })
```

lib/rules/prefer-screen-queries.ts
timdeschryver: I thought we had a linter rule for this.

Belco90:
> One thing I noticed from this rule is that it does not check `screen` is imported from RTL, nor anything about custom renders. Not sure if we should all of that for the v4 in #198 or if checking imports should be done in a separated ticket for the current version and then enhanced with custom renders for the v4... I'll assume it will be easier to do everything directly in v4 to avoid doing weird rebases but I want to know your opinion folks.
> 
> For this PR I just stick with the feature requested

That's right. We will address the checking `screen` imported on #198. As mentioned in that ticket, after using the mechanism in a couple of rules should be easy to apply it to remaining ones.

lib/node-utils.ts
Belco90: Sounds good!

lib/rules/prefer-screen-queries.ts
Belco90: We don't have something to for this as far as I remember. In some rules is at the top, in others at the bottom.

renatoagds: > We don't have something to for this as far as I remember. In some rules is at the top, in others at the bottom.

Can we add this rule? or is not necessary in this project?

Belco90: I'm not sure which ESLint rule cover this, but I guess it's better to address it on a separate ticket as it will implies fixing all reported issues (which is not directly related to this rule).

timdeschryver: Hmmm I really thought it was included in the recommended config, and that I moved all the utilities to the top because I had linting errors 🤔

lib/node-utils.ts
gndelia: #203

docs/rules/prefer-screen-queries.md
Belco90, line 67: I guess by comment above this should be `container` instead of `baseElement`

tests/lib/rules/prefer-screen-queries.test.ts
```diff
@@ -6,6 +6,9 @@ const ruleTester = createRuleTester();
 
 ruleTester.run(RULE_NAME, rule, {
   valid: [
+    {
+      code: `const baz = () => 'foo'`
```
Belco90, line 10: I don't get this valid case to be honest

gndelia: for the sake of code coverage 🤷  I needed a variable declaration that wasn't the return of a function call (`const foo = baz()`) nor a destructuring assignation (`const { foo } = baz()`)

Belco90: Ah ok, no problem then.

docs/rules/prefer-screen-queries.md
gndelia: fixed

lib/rules/prefer-screen-queries.ts
gndelia: just in case I moved above 😅

PR #208, opened 2020-08-04
fix(await-async-query): false positives for await-async-query

### LINKED ISSUE #122: False positives on `await-async-query` and `no-await-sync-query`

thomlom:
Closes #122.

## Changes

- Move `isAwaited` and `isPromiseResolved` to `node-utils.ts` that was used in three different rules (`await-async-utils`, `await-async-query`, `await-fire-event`)
- Prevent the `await-async-query` rule to fire if nothing is imported from one of the testing library modules as declared in `LIBRARY_MODULES` in `utils.ts`.
- Refactor the test cases in `await-async-query` to get closer to a real test case

Note that this is a partial improvement to the rule. Indeed, if one is using (and importing) a custom render function in his tests, he won't benefit from the rule as this function would not be imported from a testing library module such as `@testing-library/react` or `@testing-library/vue`.

A fix to that would be to allow the user to input what custom render function he uses, which is [one of the breaking changes coming in V4](https://github.com/testing-library/eslint-plugin-testing-library/issues/198). I think we will be able to further improve this rule after the V4.

What do you think?

### DIFF
--- a/lib/rules/await-async-query.ts
+++ b/lib/rules/await-async-query.ts
@@ -1,11 +1,12 @@
 import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils';
 import { getDocsUrl, LIBRARY_MODULES } from '../utils';
 import {
-  isVariableDeclarator,
-  hasThenProperty,
   isCallExpression,
   isIdentifier,
   isMemberExpression,
+  isAwaited,
+  isPromiseResolved,
+  getVariableReferences,
 } from '../node-utils';
 import { ReportDescriptor } from '@typescript-eslint/experimental-utils/dist/ts-eslint';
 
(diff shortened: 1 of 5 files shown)


tests/lib/rules/await-async-query.test.ts
```diff
@@ -7,199 +8,194 @@ import {
...
+interface TestCode {
+  code: string;
+  isAsync?: boolean;
+}
+
+function createTestCode({ code, isAsync = true }: TestCode) {
+  return `
+    import { render } from '@testing-library/react'
+    test("An example test",${isAsync ? ' async ' : ' '}() => {
+      ${code}
```
thomlom, line 20: This function allows to get closer to a real test. I'm happy to move it to `test-utils` and refactor some tests (for example in `await-async-utils`) if needed 🙂

lib/rules/await-async-query.ts
```diff
@@ -1,41 +1,21 @@
 import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils';
-import { getDocsUrl } from '../utils';
+import { getDocsUrl, LIBRARY_MODULES } from '../utils';
```
thomlom, line 2: I think there's an improvement to do on the `LIBRARY_MODULES` naming. Indeed, it doesn't make sense to make the rule fire on the async queries for `@testing-library/cypress`. Right now it works because it's not included in the `LIBRARY_MODULES` constants, however it's still a library module.

tests/lib/rules/await-async-query.test.ts
Belco90: This is interesting. I'd prefer to wait until v4 is released to avoid messing up the code and having to resolve lot of conflicts, but it's definitely a nice to have after it! It would need some improvements tho to be able to customize the import etc.

PR #209, opened 2020-08-04
feat(no-render-in-setup): adds no-render-in-setup rule, closes #207

### LINKED ISSUE #207: feature request: new rule; no-render-in-setup
imo, it would be nice to have a rule that doesn't allow the usage of `render` in setup functions (before, beforeEach, etc)

Would also be nice to have the ability to configure custom render functions, we have one for renderWithContext for example.

I've seen people do code like this:

```js
beforeEach(() => {
  render(<MyComponent />);
});

it("Should have foo", () => {
  expect(screen.getByText("foo")).toBeInTheDocument();
});

it("Should have bar", () => {
  expect(screen.getByText("bar")).toBeInTheDocument();
});

it("Should have baz", () => {
  expect(screen.getByText("baz")).toBeInTheDocument();
});
```

imo this should just be a single test.  why render it N times when nothing is changing in the render?

alessbell:
Added a `no-render-in-setup` rule with tests to address https://github.com/testing-library/eslint-plugin-testing-library/issues/207.

The rule looks for uses of `render` (or a custom render function who name can be passed in via config options) inside of two "before hooks": `beforeEach` and `beforeAll`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,9 @@
 [![Tweet][tweet-badge]][tweet-url]
 
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
+
 [![All Contributors](https://img.shields.io/badge/all_contributors-29-orange.svg?style=flat-square)](#contributors-)
+
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
 
 ## Installation
@@ -141,6 +143,7 @@ To enable this configuration use the `extends` property in your
 | [no-debug](docs/rules/no-debug.md)                                     | Disallow the use of `debug`                                                | ![angular-badge][] ![react-badge][] ![vue-badge][]                        |                    |
 | [no-dom-import](docs/rules/no-dom-import.md)                           | Disallow importing from DOM Testing Library                                | ![angular-badge][] ![react-badge][] ![vue-badge][]                        | ![fixable-badge][] |
 | [no-manual-cleanup](docs/rules/no-manual-cleanup.md)                   | Disallow the use of `cleanup`                                              |                                                                           |                    |
+| [no-render-in-setup](docs/rules/no-render-in-setup.md)                 | Disallow the use of `render` in setup functions                            |                                                                           |                    |
 | [no-wait-for-empty-callback](docs/rules/no-wait-for-empty-callback.md) | Disallow empty callbacks for `waitFor` and `waitForElementToBeRemoved`     |                                                                           |                    |
 | [prefer-explicit-assert](docs/rules/prefer-explicit-assert.md)         | Suggest using explicit assertions rather than just `getBy*` queries        |                                                                           |                    |
 | [prefer-find-by](docs/rules/prefer-find-by.md)                         | Suggest using `findBy*` methods instead of the `waitFor` + `getBy` queries | ![recommended-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] | ![fixable-badge][] |
@@ -219,6 +222,7 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
 
 <!-- markdownlint-enable -->
 <!-- prettier-ignore-end -->
+
 <!-- ALL-CONTRIBUTORS-LIST:END -->
 
 This project follows the [all-contributors](https://github.com/all-contributors/all-contributors) specification. Contributions of any kind welcome!
--- a/docs/rules/no-render-in-setup.md
+++ b/docs/rules/no-render-in-setup.md
@@ -1,0 +1,42 @@
+# Disallow the use of `render` in setup functions (no-render-in-setup)
+
+## Rule Details
+
+This rule disallows the usage of `render` in setup functions (`beforeEach` or `beforeAll`) in favor of a single test with multiple assertions.
+
+Examples of **incorrect** code for this rule:
+
+```js
+beforeEach(() => {
+  render(<MyComponent />);
+});
+
+it('Should have foo', () => {
+  expect(screen.getByText('foo')).toBeInTheDocument();
+});
+
+it('Should have bar', () => {
+  expect(screen.getByText('bar')).toBeInTheDocument();
+});
+
+it('Should have baz', () => {
+  expect(screen.getByText('baz')).toBeInTheDocument();
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+it('Should have foo, bar and baz', () => {
+  render(<MyComponent />);
+  expect(screen.getByText('foo')).toBeInTheDocument();
+  expect(screen.getByText('bar')).toBeInTheDocument();
+  expect(screen.getByText('baz')).toBeInTheDocument();
+});
+```
+
+If you use [custom render functions](https://testing-library.com/docs/example-react-redux) then you can set a config option in your `.eslintrc` to look for these.
+
+```
+   "testing-library/no-render-in-setup": ["error", {"renderFunctions":["renderWithRedux", "renderWithRouter"]}],
+```
--- a/docs/rules/no-render-in-setup.md
+++ b/docs/rules/no-render-in-setup.md
@@ -41,7 +41,7 @@ If you use [custom render functions](https://testing-library.com/docs/example-re
    "testing-library/no-render-in-setup": ["error", {"renderFunctions": ["renderWithRedux", "renderWithRouter"]}],
 ```
 
-If you would would like to allow the use of `render` (or custom render function) in _either_ `beforeAll` or `beforeEach`, this can be configured using the option `allowTestingFrameworkSetupHook`. This may be useful if you have configured your tests to [skip auto cleanup](https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup). `allowTestingFrameworkSetupHook` is an enum that accepts either `"beforeAll"` or `"beforeEach"`.
+If you would like to allow the use of `render` (or custom render function) in _either_ `beforeAll` or `beforeEach`, this can be configured using the option `allowTestingFrameworkSetupHook`. This may be useful if you have configured your tests to [skip auto cleanup](https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup). `allowTestingFrameworkSetupHook` is an enum that accepts either `"beforeAll"` or `"beforeEach"`.
 
 ```
    "testing-library/no-render-in-setup": ["error", {"allowTestingFrameworkSetupHook": "beforeAll"}],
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -23,7 +23,9 @@
 [![Tweet][tweet-badge]][tweet-url]
 
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
+
```
alessbell, line 26: These empty lines were added in the precommit hook

lib/rules/no-render-in-setup.ts
```diff
@@ -0,0 +1,74 @@
...
+  create(context, [{ renderFunctions }]) {
+    return {
+      CallExpression(node) {
+        const beforeHook = findClosestBeforeHook(node);
+        if (isRenderFunction(node, renderFunctions) && beforeHook) {
+          context.report({
+            node,
+            messageId: 'noRenderInSetup',
+            data: {
+              name: beforeHook.name,
```
alessbell, line 67: I wasn't sure which name made the most sense to pass to `data` here

gndelia:
Thanks for your first PR ❤️ 🚀 

For the styling changes, we've seen something similar in other PRs, so I just opened a ticket to review that #210  so it's ok - don't worry about it.

For the code in general, it looks great to me.

Perhaps my only suggestion is about the error message (both in the rule and in the docs).

They said

> (docs) This rule disallows the usage of `render` in setup functions (`beforeEach` or `beforeAll`) in favor of a single test with multiple assertions.
> (rule error) 'Combine assertions into a single test instead of re-rendering the component.'

I used to work in a project where we would use `render` in `beforeEach` just to save common code, and then each (multiple) test would have its own path starting from the render (plus assertions). When we dropped that convention, we went to move the render into each test, rather than combine them into only a large one. Those tests were unrelated and large enough to combine them.  
Because of that, I'd drop the `combine assertions into a single test` part and put something about moving the `render` closer to the assertions, or moving it into each test. Another option could just say why moving the `render` out of the `beforeEach`/`beforeAll` and leave open to the dev where to put it.

It's an open suggestion so I want to also hear everyone else's thoughts and yours too. 

Other than that, everything looks good to me so we can probably merge this soon.

alessbell:
Thanks @gndelia! I totally agree, and I've had a similar experience refactoring tests as the one you've described.

I'm also now wondering if this rule should only be applied to `beforeEach` if the intent is to prevent re-renders, since rendering in `beforeAll` could be used to organize assertions into different tests, which seems more like a stylistic choice. E.g.:

> (docs) This rule disallows the usage of `render` in the `beforeEach` setup function to prevent unnecessary re-renders.
> (rule error) 'Move `render` out of `beforeEach` and closer to test assertions to prevent unnecessary re-renders.'

If the goal is moving render closer to assertions more generally:

> (docs) This rule disallows the usage of `render` in setup functions (`beforeEach` or `beforeAll`) in favor of moving `render` closer to test assertions.
> (rule error) 'Move `render` out of `{{name}}` and into individual tests.'

Curious to hear others' thoughts!

README.md
Belco90: Yeah, this is changing the whole time 😓 I think we need to include a prettier:ignore comment to handle this properly. Sorry about weirdness this may have caused!

lib/rules/no-manual-cleanup.ts
```diff
@@ -21,7 +21,7 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
   meta: {
     type: 'problem',
     docs: {
-      description: ' Disallow the use of `cleanup`',
+      description: 'Disallow the use of `cleanup`',
```
Belco90, line 24: ✨

lib/rules/no-render-in-setup.ts
Belco90: I think it makes sense but I don't see it used in the message itself!

lib/utils.ts
```diff
@@ -63,6 +63,8 @@ const ASYNC_UTILS = [
   'waitForDomChange',
 ];
 
+const BEFORE_HOOKS = ['beforeEach', 'beforeAll'];
```
Belco90, line 66: I think it would be better to put a more verbose name for this one, something like `TESTING_FRAMEWORK_SETUP_HOOKS` to stick to [setup/teardown concepts](https://jestjs.io/docs/en/setup-teardown) and to make clear the hooks are related to the testing framework (and not something like React hooks or anything else).

tests/lib/rules/no-render-in-setup.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    jsx: true,
+  },
+});
+
+ruleTester.run(RULE_NAME, rule, {
+  valid: [
+    {
+      code: `
+        it('Test', () => {
+          render(<Component/>)
```
Belco90, line 16: Could you add a couple of additional valid test checking that the rule doesn't report a `render` used within `beforeEach` and `beforeAll` if that render method doesn't come from testing library? Something like

```javascript
import { render } from '@somewhere/else';

beforeEach(() => {
  render(<Component />);
});
```

This has been checked already in existing rules. Let me know if you need some help with this!

Belco90:
@gndelia @alessbell I agree on updating the error message to make it more generic and not so attached to the assert combination.

Good point about if it makes sense to report when used within `beforeAll`. I think it makes sense for the rule to report in that case, but maybe with a different error message?

gndelia:
I think I remember there were some issues with using `beforeAll` with RTL, because there's an implicit `afterEach` cleanup in every test (by default) that prevents from using `render` in a `beforeAll` successfully. But I am not sure if that's configurable, if it can be deactivated or what.

Assuming it can be used properly, I believe it makes sense to report it. 
What about allowing a config to disable one or the other? The rule by default would trigger on both, but the user would be able to configure which one to disable (disabling both makes no sense - disable the rule in that scenario)
Not sure if it makes sense though. What are your thoughts @Belco90 @alessbell ?

alessbell:
Thanks for the reviews @gndelia @Belco90!

I've attempted to address most of your comments, with the exception of [this one](https://github.com/testing-library/eslint-plugin-testing-library/pull/209#discussion_r465891668) regarding ignoring render methods that don't come from testing library. I will add that in a follow-on commit shortly :)

I've made the docs/error language more general, and added an enum config option `allowTestingFrameworkSetupHook` which accepts either `beforeAll` or `beforeEach` to be allowlisted. Good call @gndelia - RTL cites several ways to [skip clean-up](https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup), so allowing render in `beforeAll` may make sense for some while still disallowing it in `beforeEach`. Let me know what you think!

Belco90:
@alessbell awesome! The rule implementation is looking really good. No rush with the import check one. Thanks for your work!

docs/rules/no-render-in-setup.md
```diff
@@ -0,0 +1,48 @@
...
+});
+```
+
+If you use [custom render functions](https://testing-library.com/docs/example-react-redux) then you can set a config option in your `.eslintrc` to look for these.
+
+```
+   "testing-library/no-render-in-setup": ["error", {"renderFunctions": ["renderWithRedux", "renderWithRouter"]}],
+```
+
+If you would would like to allow the use of `render` (or custom render function) in _either_ `beforeAll` or `beforeEach`, this can be configured using the option `allowTestingFrameworkSetupHook`. This may be useful if you have configured your tests to [skip auto cleanup](https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup). `allowTestingFrameworkSetupHook` is an enum that accepts either `"beforeAll"` or `"beforeEach"`.
```
timdeschryver, line 44:
```suggestion
If you would like to allow the use of `render` (or custom render function) in _either_ `beforeAll` or `beforeEach`, this can be configured using the option `allowTestingFrameworkSetupHook`. This may be useful if you have configured your tests to [skip auto cleanup](https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup). `allowTestingFrameworkSetupHook` is an enum that accepts either `"beforeAll"` or `"beforeEach"`.
```

docs/rules/no-render-in-setup.md
```diff
@@ -0,0 +1,48 @@
...
+});
+
+it('Should have bar', () => {
+  expect(screen.getByText('bar')).toBeInTheDocument();
+});
+
+it('Should have baz', () => {
+  expect(screen.getByText('baz')).toBeInTheDocument();
+});
+```
```
gndelia, line 25: maybe we should add the `beforeAll` in the examples as it's mentioned too.

docs/rules/no-render-in-setup.md
```diff
@@ -0,0 +1,48 @@
...
+
+```
+   "testing-library/no-render-in-setup": ["error", {"renderFunctions": ["renderWithRedux", "renderWithRouter"]}],
+```
+
+If you would like to allow the use of `render` (or custom render function) in _either_ `beforeAll` or `beforeEach`, this can be configured using the option `allowTestingFrameworkSetupHook`. This may be useful if you have configured your tests to [skip auto cleanup](https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup). `allowTestingFrameworkSetupHook` is an enum that accepts either `"beforeAll"` or `"beforeEach"`.
+
+```
+   "testing-library/no-render-in-setup": ["error", {"allowTestingFrameworkSetupHook": "beforeAll"}],
+```
```
gndelia, line 48: 🎉

docs/rules/no-render-in-setup.md
alessbell: Good call! Added 👍

alessbell:
Thanks @gndelia! I believe I just have one more fix to make, and it's related to the commented out test that is still failing. Per @Belco90's advice, I will need to adjust the rule so that e.g.:

```js
import { render } from 'imNoTestingLibrary';

beforeAll(() => {
  render(<Component/>)
})`,
```

does **not** trigger the error. I believe that's all that is left, and I should be able to add it soon :) Thanks for all the comments!

alessbell:
Updated @gndelia @Belco90 :) LMK if there's anything else I should add/fix, have a great weekend!

lib/node-utils.ts
```diff
@@ -50,6 +50,22 @@ export function isVariableDeclarator(
   return node && node.type === AST_NODE_TYPES.VariableDeclarator;
 }
 
+export function isRenderFunction(
```
Belco90, line 53: I thought this one was already moved to `node-utils.ts` but that was done in v4 of course 😓

Belco90:
@all-contributors please add @alessbell for code, tests and doc.

PR #211, opened 2020-08-10
feat: add support for multiple test id attributes

skovy:
This PR proposes adding support for multiple test id attributes (via `testIdAttribute`) to the `consistent-data-testid` rule. 

### Movitation

When making shared components, some prefer to expose a `testId` prop ([example in Atlastkit](https://atlaskit.atlassian.com/packages/design-system/button)) in addition to using the `data-testid` prop. Ideally, this linting rule could handle more than one test id attribute to handle cases like this.

### Changes

The configuration was updated to allow either a `string` (existing API) or now an array of strings.  This was done so the rule continues to work with existing configurations. I also considered adding regex support which could also achieve this and happy to update with that approach if preferred.

### DIFF
--- a/docs/rules/consistent-data-testid.md
+++ b/docs/rules/consistent-data-testid.md
@@ -24,10 +24,10 @@ const baz = props => <div>...</div>;
 
 ## Options
 
-| Option            | Required | Default       | Details                                                                                                                                                                                                                                                       | Example                                |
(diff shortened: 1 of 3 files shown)


Belco90:
@all-contributors please add @skovy for code, test and doc.

PR #215, opened 2020-08-15
chore: updating dev deps. check formatting on travis ci

### LINKED ISSUE #210: Check formatting and eslint in CI + review pre-commit hook
There have been a few PRs that had style modifications in files or Readme examples of code that were no actual changes as part of the PR that was being opened. 

We are not enforcing our defined styling enough (I think it's only run in a `pre-commit` hook) which can be skipped. This ticket is to

- [ ] Review the pre-commit hook to ensure it is actually working, and (except for manual skip by the dev) that it always runs when committing new code

- [ ] Update our CI process to verify the code to be merged is following the eslint and prettier rules, preventing new code from being merged if it does not follow our style guide

gndelia:
closes #210 

I'll be doing some commits with `--no-verify` to force CI failure (to check it's working)

I also updated 4 dev dependencies

- lint-staged
- husky
- is-ci-cli (not required for the PR, but it's not a breaking change and given I was updating the CI conf...)
- prettier => in this case I did not go to the last up-to-date version, but to the latest stable without breaking changes - in this case 1.19.1. That's because in 1.19 they included support for parsing optional chaining `(foo?.baz)` which we use in TS files.
I could update prettier to the latest but I thought it'd be better to do it in a separated PR as I think they changed some defaults and it may do the PR larger than it's required.

The breaking changes of these ☝️  is just dropping old node support, which is not required for linting  our code


gndelia:
This is ready to review
many changes in `ts` files are added because prettier was no run in a couple of them 🤔

gndelia:
closing in favor of applying these changes in a PR targeting v4

PR #216, opened 2020-08-16
chore: lint and format on pre-commit and ci

### LINKED ISSUE #210: Check formatting and eslint in CI + review pre-commit hook

gndelia:
closes #210 

I also updated 4 dev dependencies

- lint-staged
- husky
- is-ci-cli (not required for the PR, but it's not a breaking change and given I was updating the CI conf...)
- prettier => in this case I did not go to the last up-to-date version, but to the latest stable without breaking changes - in this case 1.19.1. That's because in 1.19 they included support for parsing optional chaining `(foo?.baz)` which we use in TS files.
I could update prettier to the latest but I thought it'd be better to do it in a separated PR as I think they changed some defaults and it may do the PR larger than it's required.

The breaking changes of these ☝️  is just dropping old node support, which is not required for linting  our code

it's the same as #215 but targeting v4 instead of master 😅 

the changes of this PR in ts files are formatting stuff (automatically applied by prettier) or eslint fixes


.travis.yml
```diff
@@ -17,6 +17,12 @@ before_script:
 
 jobs:
   include:
+    - stage: test
```
Belco90, line 20: This one should be validation stage right?

Belco90:
Agree on upgrading prettier in a different task.

.travis.yml
gndelia: I thought the stages where only test, build and deploy, that's why I went with this one. If validation is a valid one, it makes sense. I don't like travis ci docs 😛

gndelia: Updated in https://github.com/testing-library/eslint-plugin-testing-library/pull/216/commits/b126b8d4b41fa86a7280305f13331cae9c15b7b2

Belco90: I'd say stages can have any name: https://config.travis-ci.com/ref/stages the only thing is to be consistent around the CI config.

Belco90:
Now we have corresponding stage for Validation ✨ 
CleanShot 2020-08-17 at 10 55 03

gndelia:
> Now we have corresponding stage for Validation ✨
> CleanShot 2020-08-17 at 10 55 03

🚀

Issue #218, opened 2020-08-19
Enforce consistent assertions 


skovy:
The [`prefer-explicit-assert`](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/prefer-explicit-assert.md) rule enforces an explicit assertion but allows a large number of possible assertions. For example:

```javascript
expect(getByText('foo')).toBeDefined();
expect(getByText('foo')).toBeTruthy();
expect(getByText('foo')).not.toBeNull();
expect(getByText('foo')).toBeInTheDocument();
```

Functionally, these are all the same (since `getBy*` throws) but in a large codebase, it's possible these can all be used interchangeably. It would be nice to enforce a consistent assertion on `getBy*` queries. Even though these behave the same with `getBy*` (since it throws or always returns something) it's misleading mixing them since [`toBeDefined`](https://jestjs.io/docs/en/expect#tobedefined) is explicitly only checking that it's _not_ `undefined`, but [`toBeTruthy`](https://jestjs.io/docs/en/expect#tobetruthy) is checking that it's _not_ six values: `false`, `0`, `''`, `null`, `undefined`, and `NaN`.

One way this could be added is by adding a new configuration option:

```json
"testing-library/prefer-explicit-assert": ["error", {"assertion": "toBeInTheDocument"}],
```

Or, it could be added as a new rule (eg: `prefer-consistent-assert`).

Happy to put together a PR that makes these changes, but wanted to get feedback if this aligns with the intent of these linting rules and also the preferred implementation?

Belco90:
Hello Spencer! Thanks for your suggestions, this is a really good idea actually.

When I originally implemented `prefer-explicit-assert` I wanted to allow different assertions checking truthy. You made a good point about adding the possibility to customize which of those assertions is considered a valid explicit one, which ideally always should be `toBeInTheDocument`.

I'm happy to add config option to the rule as you described, I think it's simple and makes sense to be part of this rule.

skovy:
@Belco90 thanks for the quick feedback and review!

I'm not realizing I entirely forgot about a valid set of use cases to _not_ use `toBeInTheDocument`, for example:

```js
expect(screen.getByText("foo")).toBeDisabled();
expect(screen.getByText("foo")).toHaveAttribute("href", "/foo");
// etc.
```

[...]

PR #220, opened 2020-08-20
feat: add 'assertion' config option to prefer-explicit-assert

skovy:
Resolves https://github.com/testing-library/eslint-plugin-testing-library/issues/218.

This Pull Request adds support for the `assertion` rule configuration option to the `prefer-explicit-assert` rule. See the linked issue for more details on the rationale. The general idea is to allow enforcing a consistent assertion (eg: `toBeInDocument`) when requiring `getBy*` queries to be wrapped in an expectation.

### DIFF
--- a/lib/rules/prefer-explicit-assert.ts
+++ b/lib/rules/prefer-explicit-assert.ts
@@ -3,10 +3,13 @@ import { getDocsUrl, ALL_QUERIES_METHODS } from '../utils';
 import { isMemberExpression } from '../node-utils';
 
 export const RULE_NAME = 'prefer-explicit-assert';
-export type MessageIds = 'preferExplicitAssert';
+export type MessageIds =
+  | 'preferExplicitAssert'
+  | 'preferExplicitAssertAssertion';
 type Options = [
   {
-    customQueryNames: string[];
+    assertion?: string;
+    customQueryNames?: string[];
   }
 ];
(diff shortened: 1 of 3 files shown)


lib/rules/prefer-explicit-assert.ts
```diff
@@ -34,12 +37,18 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
       preferExplicitAssert:
         'Wrap stand-alone `getBy*` query with `expect` function for better explicit assertion',
+      preferExplicitAssertAssertion:
+        '`getBy*` queries must be asserted with `{{assertion}}`',
     },
     fixable: null,
     schema: [
       {
         type: 'object',
+        additionalProperties: false,
```
skovy, line 47: I originally didn't have to add the `assertion` type because the schema wasn't validating excess properties. Technically this could be considered a breaking change if any folks have an invalid configuration, but this is now consistent with other rules (eg: [`consistent-data-testid.`](https://github.com/testing-library/eslint-plugin-testing-library/blob/609c6dbe9e97eb36993ca1d5c05ad4bc2f66e405/lib/rules/consistent-data-testid.ts#L33)). Thought it was worthwhile to make this change to also catch typos (eg: `customQueryName`)?

Issue #214, opened 2020-08-14
FR Rule: Prevent snapshot inside wait


SBoudrias:
I wonder if there'd be a way to make sure there's no snapshot done inside wait conditions.

Example invalid usage:

```js
await wait(() => {
  expect(foo).toMatchSnapshot();
});
```

Valid:
```js
expect(foo).toMatchSnapshot();
```

This is between jest and and testing-library, but the depreciated `wait()` functions allow tests like this to be written and lead to flakey cases.

Belco90:
Hey Simon, sorry for the late response!

This sounds like an interesting rule, but I don't see the problem 100% clear. Could you elaborate more around the issues this can cause please? Thanks.

SBoudrias:
Yes of course!

The problem is that `wait()` runs in a timer loop. So it'll retry every n amount of time.

If there's a snapshot inside the wait condition, jest will generate one snapshot per loop.

The problem then is the amount of loop ran until the condition is met will vary between different computers (or CI machines.) This leads to tests that will regenerate a lot of snapshots until the condition is match when devs run those tests locally updating the snapshots; e.g devs cannot run `jest -u` locally or it'll generate a lot of invalid snapshots who'll fail during CI.

This would be fine if snapshots were generated manually; but auto-update causes conflicts.

Belco90:
I see. Yeah, didn't think about the auto update. This can definitely be a nice rule, something like `no-wait-snapshot`. Are you interested into give it a try?

PR #223, opened 2020-08-26
feat: 🎸 new no-wait-for-snapshot rule

Gpx:
✅ Closes: #214

The rule is checking for all async utils not only `wait`. If that's not the preferred behavior I can change it back. It's my first contribution to this project so any feedback is welcome 😄

### DIFF
--- a/docs/rules/no-wait-for-snapshot.test.md
+++ b/docs/rules/no-wait-for-snapshot.test.md
@@ -1,0 +1,56 @@
+# Ensures no snapshot is generated inside of a `wait` call' (no-wait-for-snapshot)
+
+Ensure that no calls to `toMatchSnapshot` or `toMatchInlineSnapshot` are made from within a `waitFor` method (or any of the other async utility methods).
+
+## Rule Details
+
+The `waitFor()` method runs in a timer loop. So it'll retry every n amount of time.
+If a snapshot is generated inside the wait condition, jest will generate one snapshot per loop.
+
+The problem then is the amount of loop ran until the condition is met will vary between different computers (or CI machines). This leads to tests that will regenerate a lot of snapshots until the condition is matched when devs run those tests locally updating the snapshots; e.g devs cannot run `jest -u` locally or it'll generate a lot of invalid snapshots who'll fail during CI.
+
+Note that this lint rule prevents from generating a snapshot from within any of the [async utility methods](https://testing-library.com/docs/dom-testing-library/api-async).
+
+Examples of **incorrect** code for this rule:
+
+```js
+const foo = async () => {
+  // ...
+  await waitFor(() => expect(container).toMatchSnapshot());
+  // ...
+};
+
+const bar = async () => {
+  // ...
+  await waitFor(() => expect(container).toMatchInlineSnapshot());
+  // ...
+};
+
+const baz = async () => {
+  // ...
+  await wait(() => {
+    expect(container).toMatchSnapshot();
+  });
+  // ...
+};
+```
+
+Examples of **correct** code for this rule:
+
+```js
+const foo = () => {
+  // ...
+  expect(container).toMatchSnapshot();
+  // ...
+};
+
+const bar = () => {
+  // ...
+  expect(container).toMatchInlineSnapshot();
+  // ...
+};
+```
+
+## Further Reading
+
+- [Async Utilities](https://testing-library.com/docs/dom-testing-library/api-async)
--- a/lib/node-utils.ts
+++ b/lib/node-utils.ts
@@ -84,6 +84,8 @@ export function findClosestCallExpressionNode(
     return node;
   }
 
+  if(!node.parent) return null
+
   return findClosestCallExpressionNode(node.parent);
 }
 
--- a/lib/rules/no-wait-for-snapshot.ts
+++ b/lib/rules/no-wait-for-snapshot.ts
@@ -1,0 +1,132 @@
+import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils';
+import { getDocsUrl, ASYNC_UTILS, LIBRARY_MODULES } from '../utils';
+import {
+  findClosestCallExpressionNode,
+  isMemberExpression,
+} from '../node-utils';
+
+export const RULE_NAME = 'no-wait-for-snapshot';
+export type MessageIds = 'noWaitForSnapshot';
+type Options = [];
+
+const ASYNC_UTILS_REGEXP = new RegExp(`^(${ASYNC_UTILS.join('|')})$`);
+const SNAPSHOT_REGEXP = /^(toMatchSnapshot|toMatchInlineSnapshot)$/;
+
+export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'problem',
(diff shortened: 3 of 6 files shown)


docs/rules/no-wait-snapshot.test.md
```diff
@@ -0,0 +1,42 @@
+# Ensures no snapshot is generated inside of a `wait` call' (no-wait-snapshot)
+
+Ensure that no calls to `toMatchSnapshot` are made from within a `wait` method
```
Belco90, line 3: Can you make reference to `waitFor` rather than `wait`, and add a note at the end of Rule Details to clarify all util methods this rule reports?

Belco90 (changes requested):
Hey Giorgio, thanks for your first PR! Looking good, I just requested some rewording on the rule doc and should be fine.

We are usually asking to check if corresponding methods are imported from RTL, but I think it's fine with this one until we release v4.

docs/rules/no-wait-snapshot.test.md
Gpx: Sure thing

Gpx: Do you want to change the name of the rule too? Something like `no-waitfor-snapshot`?

Belco90: That's a good idea! Let's go for `no-wait-for-snapshot` for consistency with `no-wait-for-empty-callback`

lib/node-utils.ts
```diff
@@ -84,6 +84,8 @@ export function findClosestCallExpressionNode(
     return node;
   }
 
+  if(!node.parent) return null
```
Gpx, line 87: I've added this to avoid errors in case `node.parent` is `null`

docs/rules/no-wait-for-snapshot.test.md
```diff
@@ -0,0 +1,44 @@
+# Ensures no snapshot is generated inside of a `wait` call' (no-wait-for-snapshot)
+
+Ensure that no calls to `toMatchSnapshot` are made from within a `waitFor` method (or any of the other async utility methods).
+
+## Rule Details
+
+The `waitFor()` method runs in a timer loop. So it'll retry every n amount of time.
+If a snapshot is generated inside the wait condition, jest will generate one snapshot per loop.
+
+The problem then is the amount of loop ran until the condition is met will vary between different computers (or CI machines.) This leads to tests that will regenerate a lot of snapshots until the condition is match when devs run those tests locally updating the snapshots; e.g devs cannot run `jest -u` locally or it'll generate a lot of invalid snapshots who'll fail during CI.
```
gndelia, line 10: minor spelling suggestions: 
```suggestion
The problem then is the amount of loop ran until the condition is met will vary between different computers (or CI machines). This leads to tests that will regenerate a lot of snapshots until the condition is matched when devs run those tests locally updating the snapshots; e.g devs cannot run `jest -u` locally or it'll generate a lot of invalid snapshots who'll fail during CI.
```

gndelia:
Thanks for your PR !! 🚀 ❤️ 

@Belco90  one doubt I had: should we target new rules directly to v4, in order to simplify conflict resolution once we want to go out with that version? Especially for the changes we're considering to add in all rules

lib/rules/no-wait-for-snapshot.ts
```diff
@@ -0,0 +1,115 @@
...
+    type: 'problem',
+    docs: {
+      description:
+        'Ensures no snapshot is generated inside of a `waitFor` call',
+      category: 'Best Practices',
+      recommended: 'warn',
+    },
+    messages: {
+      noWaitForSnapshot:
+        "A snapshot can't be generated inside of a `waitFor` call",
```
gndelia, line 23: I think it'd be better to here use the async utility message that caused the error, rather than the hardcoded `waitFor`.

Otherwise, we'd face scenarios like the following:

```ts
await wait(() => {
    expect(container).toMatchSnapshot();
  });
// output error message: 'Ensures no snapshot is generated inside of a `waitFor` call'
// expected output error message: 'Ensures no snapshot is generated inside of a `wait` call'
```

lib/rules/no-wait-for-snapshot.ts
```diff
@@ -0,0 +1,115 @@
...
+          name: memberExpressionName,
+        });
+      },
+      [`Identifier[name='toMatchSnapshot']`](node: TSESTree.Identifier) {
+        snapshotUsage.push(node);
+      },
+      'Program:exit'() {
+        const testingLibraryUtilUsage = asyncUtilsUsage.filter(usage => {
+          if (usage.node.type === 'MemberExpression') {
+            const object = usage.node.object as TSESTree.Identifier;
```
gndelia, line 80: There's a type guard utility function [here](https://github.com/testing-library/eslint-plugin-testing-library/blob/636273a23d9bda48a9d2945abf51df9d1ee93c59/lib/node-utils.ts#L14) that will allow you to drop the  cast

```suggestion
          if (isMemberExpression(usage.node)) {
           // object is automatically a TSESTree.Identifier
            const object = usage.node.object;
```

lib/rules/no-wait-for-snapshot.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        function isWithinAsyncUtil(
+          asyncUtilUsage: {
+            node: TSESTree.Identifier | TSESTree.MemberExpression;
+            name: string;
+          },
+          node: TSESTree.Node
+        ) {
+          let callExpression = findClosestCallExpressionNode(node);
+          while (callExpression != null) {
```
gndelia, line 96: I'd have hoped the linter configuration enforced `!==` over `!=`

I tend to avoid `!=` completely but I'm not sure here 😅 

Thoughts @Belco90 ?

Gpx: I think the linter allows for `!=` because I'm checking against `null`. If you use `!=` it's implicitly checking for `undefined` too

lib/rules/no-wait-for-snapshot.ts
Gpx: Good call

Gpx:
The rule is now checking for `toMatchInlineSnapshot` too and in the error message it returns which async method was used

Belco90:
> @Belco90 one doubt I had: should we target new rules directly to v4, in order to simplify conflict resolution once we want to go out with that version? Especially for the changes we're considering to add in all rules

It would simplify things on v4, but I don't want to block nice additions just to wait for v4. It will take more work on v4 but that's fine.

lib/rules/no-wait-for-snapshot.ts
Belco90: I think in this case is fine to compare against nil, and I understand that's the actual usage right?

Belco90 (approved):
Awesome! Nice improvements. I'll wait for @gndelia approval for merging it.

Belco90:
@all-contributors please add @Gpx for code, test and docs.

arvigeus:
The docs file is wrong: `docs/rules/no-wait-for-snapshot.test.md`

I don't think it should have `test` suffix. The The README also points to `docs/rules/no-wait-for-snapshot.md`

Belco90:
@arvigeus fixed in #225 !

PR #228, opened 2020-09-10
chore: remove package lock

nickmccurdy:


Belco90:
What is the motivation for this PR?

gndelia:
I believe [here's the motivation](https://github.com/testing-library/pptr-testing-library/pull/47#issuecomment-689914532)

I'm not sure if I agree either but can't do more than that 🤷‍♂️

Belco90:
I guess that's the reason, but not sure either if I agree.

MichaelDeBoey:
`create-react-app` (and a lot of other big packages) also ignores lock files: https://github.com/facebook/create-react-app/pull/7789

PR #230, opened 2020-09-19
ci: update travis config

MichaelDeBoey:


Belco90:
Can you give some context about this change please?

MichaelDeBoey:
@Belco90 I did this change in order to make sure we are also testing against the latest Node 10 and the minimum Node 12.

CI will try to run the `release` job on my fork if I would push something to master (or set my fork equal to the original repo), so that's why I added an extra condition

Belco90 (approved):

PR #234, opened 2020-09-20
chore: update dependencies + run prettier on codebase

MichaelDeBoey:
BREAKING CHANGE: Requires Node@^10.22.1 || 12.x
BREAKING CHANGE: Requires ESLint@^7.5.x

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+  singleQuote: true,
+};
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,0 @@
-{
-  "trailingComma": "es5",
-  "singleQuote": true
-}
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,11 @@ env:
   global:
     - FORCE_COLOR=true
   matrix:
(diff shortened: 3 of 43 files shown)


Belco90:
Tons of dev dependencies bumped. Are we sure we are not affected by any of their breaking changes? They'll be probably removing node v8 support and similar stuff, but still.

MichaelDeBoey:
@Belco90 I checked them all out and only Node versions are really breaking in our case

Belco90 (approved):
Let's merge this one. Could you clarify why the minimum node version went from 10.12.0 to 10.22.1 please?

Belco90:
@all-contributors please add @MichaelDeBoey for maintenance

MichaelDeBoey:
@Belco90 It's because of the `@commitlint/cli` dependency

Belco90:
That's just a dev dependency, it's not a breaking change then.

MichaelDeBoey:
@Belco90 It is, but `>=10.22.1` is needed in order to being able to work on this package locally

Belco90:
Sure, but won't be a breaking change for users using this package. So we need to remember to exclude that breaking change message to arrive to the final release.

PR #236, opened 2020-10-06
fix: consider promise.all() in await-async-utils

### LINKED ISSUE #227: [await-async-utils] false positive promise all
Hello,

Take the following example:

```
await Promise.all([
    waitFor(() =>
      expect(getByText('first-item')).toBeTruthy()
    ),
    waitFor(() =>
      expect(getByText('second-item')).toBeTruthy()
    ),
  ]);
```

This is flagged up as an error: "Promise returned from `waitFor` must be handled", which I believe is a false positive, as the Promise is indeed being handled by the Promise.all.

Thanks

gndelia:
closes #227 

After a while, trying to program a Lil' bit outside of my regular work again. Here I'm just fixing the referenced ticket. I also added a bunch of helpers to `node-utils` instead of checking the node's type directly in the rule

### DIFF
--- a/lib/rules/await-async-utils.ts
+++ b/lib/rules/await-async-utils.ts
@@ -5,6 +5,12 @@ import {
   isAwaited,
   isPromiseResolved,
   getVariableReferences,
+  isMemberExpression,
+  isImportSpecifier,
+  isImportNamespaceSpecifier,
+  isCallExpression,
+  isArrayExpression,
+  isIdentifier
 } from '../node-utils';
 
 export const RULE_NAME = 'await-async-utils';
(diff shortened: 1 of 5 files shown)


lib/rules/await-async-utils.ts
```diff
@@ -13,6 +19,17 @@ type Options = [];
...
+function isPromiseAll(node: TSESTree.CallExpression) {
+  return isMemberExpression(node.callee) && isIdentifier(node.callee.object) && node.callee.object.name === 'Promise' && isIdentifier(node.callee.property) && node.callee.property.name === 'all'
+}
+
+// verifies the node is part of an array used in a CallExpression
+function isInPromiseAll(node: TSESTree.Node) {
+  const parent = node.parent
+  return isCallExpression(parent) && isArrayExpression(parent.parent) && isCallExpression(parent.parent.parent) && isPromiseAll(parent.parent.parent)
+}
+
```
gndelia, line 32: I am not sure if these are worth moving elsewhere

gndelia:
I thought the 100% coverage was for the rules, not for the utils 🤔 🧐 

should I create a `node-utils.test.ts`?  I don't think there's a realistic scenario to get coverage for `isArrayExpression` usage from the rule (I would have to call `Promise.all()` without args or without an array [?]).

Similar goes to `isImportNamespaceSpecifier`

Belco90:
> I thought the 100% coverage was for the rules, not for the utils 🤔 🧐
> 
> should I create a `node-utils.test.ts`? I don't think there's a realistic scenario to get coverage for `isArrayExpression` usage from the rule (I would have to call `Promise.all()` without args or without an array [?]).
> 
> Similar goes to `isImportNamespaceSpecifier`
> 
> Should we exclude files that are not under `lib/rules` of the 100% threshold?

I thought so! I was planning to add specific tests for those in v4, so I'd wait for add specific tests for them. I guess we can modify the threshold meanwhile.

gndelia:
threshold updated for `node-utils`

lib/rules/await-async-utils.ts
Belco90: Yeah, they could be useful. We can move them when necessary tho.

Belco90 (approved):

PR #237, opened 2020-10-09
Move custom render functions option to ESLint plugin config: part 1

Belco90:
Relates to #198 

First part which includes:
- Generic mechanism to detect Testing Library utils
- Use the Testing Library detection mechanism in `no-node-access` rule

This mechanism is the reason why this is taking me so long. I couldn't find a way of _enhancing_ a rule's `create` for being able to inject rule listeners + utils. I finally got it after doing some research which basically was asking on ESLint discord + check how other ESLint plugins are doing this (my main reference has been `eslint-plugin-react`).

So this is what I have for now. Next steps will be applying the new detection to current rules **without modifying the current behavior**. Rules with option for custom render method are the only one which will slightly change, since that will be setup from shared config (i.e. the test must be updated to pass the config in the new way but the result must be the same). I think it's better to do that in a first round and then do a 2nd round for actually implementing the improved behavior on TL detection: check if follows TL patterns rather than if importing TL modules, check if follows TL queries patterns instead of if it's one of defined TL queries and so on.

@timdeschryver I miserably failed trying to typing this new detector. My idea was to _inherit/extends_ somehow current `create` type to inject the `helpers` type. There are several reasons I couldn't accomplish this: I don't know exactly how to define that high-order function I created, not sure how/where to define types for helper object, and no idea how can I extend original `create` type since it needs to get 2 types annotated (TOptions and TMessageIds). This is way over my head, I haven't done typings so complex! Not asking you to do it 100% tho, but at least if you could give me some guidelines it would be really appreciated.


lib/testing-library-detection.ts
```diff
@@ -0,0 +1,52 @@
...
+        );
+      },
+    };
+
+    // update given rule to inject Testing Library detection
+    const ruleInstructions = rule(context, options, helpers);
+    const enhancedRuleInstructions = Object.assign({}, ruleInstructions);
+
+    Object.keys(detectionInstructions).forEach((instruction) => {
+      enhancedRuleInstructions[instruction] = (node: any) => {
```
Belco90, line 33: I just realized I lost some types I added correctly like `TSESTree.Node` here. I was so frustrated I reverted everything 😂

lib/testing-library-detection.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    return enhancedRuleInstructions;
+  };
+}
+
+function detect(rule: any) {
+  return testingLibraryDetection(rule);
+}
+
+export default detect;
```
gndelia, line 52: I'm a bit lost, but given default exports just take the name you want, we could just export `testingLibraryDetection`

if you wanted to enforce the name `detect`, then `detect` should be a named export. Unless to expect to add more code here 🤔

lib/testing-library-detection.ts
```diff
@@ -0,0 +1,52 @@
...
+    const detectionInstructions = {
+      ImportDeclaration(node: TSESTree.ImportDeclaration) {
+        isImportingTestingLibrary = /testing-library/g.test(
+          node.source.value as string
+        );
+      },
+    };
+
+    // update given rule to inject Testing Library detection
+    const ruleInstructions = rule(context, options, helpers);
```
gndelia, line 29: Is this correct? If I understand correctly, this comes from the parameter, which in line 48 is received

but in `lib/rules/no-node-access.ts` you define

```ts
create: detect((context: any, _: any, helpers: any) => {
// code 
}
```

should that ☝️  be 

```ts
create: detect((context: any, _: any, options: any, helpers: any) => {
// code 
}
```

?

gndelia:
@Belco90  I think I can help with the types. I tested it locally and it compiles, though I cannot test it in runtime as I don't have the scenarios you have in mind 🤣 . I was able to reuse only part of the  Eslint types, and the generic part is really annoying to repeat (they are not exported, but copied everywhere in their definitions), so that's why I created the export types to be reused when using `detect`.

Here's `testing-library-detection.ts` with the types

```ts
import { TSESTree } from '@typescript-eslint/experimental-utils';
import { RuleContext, RuleModule } from '@typescript-eslint/experimental-utils/dist/ts-eslint'
// TYPES DEFINED
export type Context<TMessageIds extends string, TOptions extends readonly unknown[]> = Readonly<RuleContext<TMessageIds, TOptions>>
export type EnhancedRule<TMessageIds extends string, TOptions extends readonly unknown[]> = (context: Context<TMessageIds, TOptions>, options: Readonly<TOptions>, helpers: any) => ReturnType<RuleModule<TMessageIds, TOptions>['create']>
export type Helper = { getIsImportingTestingLibrary: () => boolean }
// TYPES DEFINED - then I replaced the proper types in each usage


/**
 * Enhances a given rule with helpers to detect Testing Library utils.
 */
function testingLibraryDetection<TMessageIds extends string, TOptions extends readonly unknown[]>(rule: EnhancedRule<TMessageIds, TOptions>) {
  return (context: Context<TMessageIds, TOptions>, options: Readonly<TOptions>) => {
    let isImportingTestingLibrary = false;

    // TODO: init here options based on shared ESLint config

    // helpers for Testing Library detection
    const helpers = {
      getIsImportingTestingLibrary() {
        return isImportingTestingLibrary;
      },
    };

    // instructions for Testing Library detection
    const detectionInstructions = {
      ImportDeclaration(node: TSESTree.ImportDeclaration) {
        isImportingTestingLibrary = /testing-library/g.test(
          node.source.value as string
        );
      },
    };

    // update given rule to inject Testing Library detection
    const ruleInstructions = rule(context, options, helpers);
    const enhancedRuleInstructions = Object.assign({}, ruleInstructions);

    Object.keys(detectionInstructions).forEach((i) => {
      const instruction = i as keyof typeof detectionInstructions
      enhancedRuleInstructions[instruction] = (node: TSESTree.ImportDeclaration) => {
        if (instruction in detectionInstructions) {
          detectionInstructions[instruction](node);
        }

        if (ruleInstructions[instruction]) {
          return ruleInstructions[instruction](node);
        }
      };
    });

    return enhancedRuleInstructions;
  };
}

function detect<TMessageIds extends string, TOptions extends readonly unknown[]>(rule: EnhancedRule<TMessageIds, TOptions>) {
  return testingLibraryDetection(rule);
}

export default detect;

```

and then, in your example in `no-node-access`, you would add the following import
```ts
import detect, { Context, Helper } from '../testing-library-detection';
// code

  create: detect((context: Context<MessageIds, Options>, _: Readonly<Options>, helper: Helper) => {
    function showErrorForNodeAccess(node: TSESTree.MemberExpression) {
      isIdentifier(node.property) &&
        ALL_RETURNING_NODES.includes(node.property.name) &&
        helper.getIsImportingTestingLibrary() &&
        context.report({
          node: node,
          loc: node.property.loc.start,
          messageId: 'noNodeAccess',
        });
    }

    return {
      ['ExpressionStatement MemberExpression']: showErrorForNodeAccess,
      ['VariableDeclarator MemberExpression']: showErrorForNodeAccess,
    };
  }),
```


let me know if it works!

lib/testing-library-detection.ts
gndelia, line 29: would `options` be where the eslint configuration (for instance: the custom import) is?

timdeschryver (commented):
The problem is that the types from `create` are inferred from the `RuleCreator`.
While @gndelia's solution works, we would still have to manually type the arguments of `create`, for example

```ts
//                       👇                               👇
create: detect((context: Context<MessageIds, Options>, _: Readonly<Options>, helper: Helper) => {
```

If we can live with this, I think the suggested solution is fine.

If we want those types to be inferred, I think the only solution would be to create our own `RuleCreator`. [See typescript-eslint's `RuleCreator`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/experimental-utils/src/eslint-utils/RuleCreator.ts#L45). I think we should be able to do that, while adding the helper method but I haven't tested it.

As a side note, while exploring the source of of eslint-typescript I saw that you could re-use/extend from another rule which I didn't know was possible:

- simple: https://github.com/typescript-eslint/typescript-eslint/blob/03886d75b5eba0a06d28bd7bbbd30c6a1b2e901c/packages/eslint-plugin/src/rules/brace-style.ts#L36
- complex: https://github.com/typescript-eslint/typescript-eslint/blob/03886d75b5eba0a06d28bd7bbbd30c6a1b2e901c/packages/eslint-plugin/src/rules/indent.ts#L177

Belco90:
@gndelia Your solution seems fine but I'd like to avoid typing `create` params as @timdeschryver said!

I think creating our own rule creator is even better so `create` is already our enhanced version.

Didn't know about extending rules either @timdeschryver! That's really useful.

I'll update my pr on Monday to move the detection to our own rule creator. Thanks guys!

lib/testing-library-detection.ts
Belco90: No, these options are always the rule options. If you want to access shared ones, you'll need to get them through `context`. The idea is not having to check shared options manually within rules, as the TL detection will check which  are those options to decide if a method is from Testing Library or not.

lib/testing-library-detection.ts
Belco90: This is a stupid function not necessary anymore. I first implemented this by calling `testingLibraryDetection` from this `detect` one using `bind` to pass the rule. Then I realized I could do the same just with a high-order function, but I forgot about removing this one! I'll get rid of it.

Belco90:
Ok I gave it a try and I think I got something. I'll cleanup my code tomorrow and add the types, so I'll update this PR after that. It's working for me but I'm seeing couple of weird things.

Btw, the base rule thing it's quite interesting. We could use that pattern to build some of our rules based on others (like await-async-X). I'm not gonna do anything related to that yet tho, or I'll will finish this by 2032 😂 

Thanks for your help!

Belco90:
@gndelia @timdeschryver I think I got it now! I created a new `createTestingLibraryRule` which wraps `ESLintUtils.RuleCreator` to 1) automatically set docs url getter and 2) enhance `create` with the 3rd param for detection helpers.

Now the types are properly set, although there are some duplication which I don't know if I could avoid somehow.

lib/create-testing-library-rule.ts
```diff
@@ -0,0 +1,37 @@
...
+import { getDocsUrl } from './utils';
+import {
+  detectTestingLibraryUtils,
+  DetectionHelpers,
+} from './detect-testing-library-utils';
+
+type CreateRuleMetaDocs = Omit<TSESLint.RuleMetaDataDocs, 'url'>;
+type CreateRuleMeta<TMessageIds extends string> = {
+  docs: CreateRuleMetaDocs;
+} & Omit<TSESLint.RuleMetaData<TMessageIds>, 'docs'>;
```
Belco90, line 11: This is copied from `@typescript-eslint/experimental-utils`, I don't know if I could inferred that somehow.

lib/create-testing-library-rule.ts
```diff
@@ -0,0 +1,37 @@
...
+    ) => TRuleListener;
+  }>
+) {
+  const { create, ...remainingConfig } = config;
+
+  return ESLintUtils.RuleCreator(getDocsUrl)({
+    ...remainingConfig,
+    create: detectTestingLibraryUtils<TOptions, TMessageIds, TRuleListener>(
+      create
+    ),
```
Belco90, line 35: Where all the magic happens now. It looks **so** simple after all the time I spent on it that I feel kinda stupid.

lib/rules/no-node-access.ts
```diff
@@ -24,19 +25,11 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
-  create(context) {
-    let isImportingTestingLibrary = false;
-
-    function checkTestingEnvironment(node: TSESTree.ImportDeclaration) {
-      isImportingTestingLibrary = /testing-library/g.test(
-        node.source.value as string
-      );
-    }
-
+  create: (context, _, helpers) => {
```
Belco90, line 28: This gets automatic types as I wanted originally 💯

lib/rules/no-node-access.ts
```diff
@@ -1,12 +1,13 @@
...
+import { ALL_RETURNING_NODES } from '../utils';
 import { isIdentifier } from '../node-utils';
+import { createTestingLibraryRule } from '../create-testing-library-rule';
 
 export const RULE_NAME = 'no-node-access';
 export type MessageIds = 'noNodeAccess';
 type Options = [];
 
-export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
+export default createTestingLibraryRule<Options, MessageIds>({
```
Belco90, line 10: Just creating rules like this everything is hooked and helpers are passed!

lib/create-testing-library-rule.ts
timdeschryver: I think this is fine, we should perhaps add the same comment in the code (that it's copied from there).
It might be useful in the future 😅

lib/create-testing-library-rule.ts
timdeschryver: Nice!
It actually looks easier and cleaner than expected ... that's easy to say when it's written.

lib/detect-testing-library-utils.ts
```diff
@@ -0,0 +1,65 @@
...
+    detectionHelpers: Readonly<DetectionHelpers>
+  ) => TRuleListener
+) {
+  return (
+    context: Readonly<TSESLint.RuleContext<TMessageIds, TOptions>>,
+    optionsWithDefault: Readonly<TOptions>
+  ): TRuleListener => {
+    let isImportingTestingLibrary = false;
+
+    // TODO: init here options based on shared ESLint config
```
timdeschryver, line 27: 👀 Just in case you missed this.

timdeschryver (commented):
Awesome work 🤩

lib/detect-testing-library-utils.ts
Belco90: This is fine, it's an actual TODO for next PR where I'll init first shared config.

lib/create-testing-library-rule.ts
Belco90: Good idea, I'm gonna add it now.

lib/create-testing-library-rule.ts
gndelia: this looks great! Great job!

PR #238, opened 2020-10-13
fix(prefer-explicit-assert): only enforce assertion when presence/absence matcher

skovy:
### Problem

This is a follow-up improvement to issue https://github.com/testing-library/eslint-plugin-testing-library/issues/218.

The original implementation was flawed.  With the way this was implemented, any `getBy*` query will trigger a linting error if it's not using `toBeInTheDocument` (assuming the config defined the `assertion` to use as `toBeInTheDocument`). However, there are still valid cases to use a different assertion. For example, if you want to check if a button is disabled (`toBeDisabled`) or maybe check if an anchor has the appropriate `href` attribute (`toHaveAttribute`).

With the way the rule was implemented, both of these examples would trigger as violations (since they're not using `toBeInTheDocument`) but they're both still valid uses of `getBy*` and thus are false positives.
### Solution

Now, this rule only triggers if the matchers used are `toBeInTheDocument`, `toBeTruthy`, `toBeDefined`, `not.toBeNull`, or `not.toBeFalsy` _and_ doesn't match the configured `assertion`. These matches already existed as a constant for another rule (`prefer-presence-queries`), so I moved them to `utils` so they could be reused. This rule also had somewhat related logic for handling negated matchers. This logic was adapted for this rule.

### DIFF
--- a/tests/lib/rules/prefer-explicit-assert.test.ts
+++ b/tests/lib/rules/prefer-explicit-assert.test.ts
@@ -74,6 +74,14 @@ ruleTester.run(RULE_NAME, rule, {
         },
       ],
     },
+    {
+      code: `expect(getByText('foo')).toBeEnabled()`,
+      options: [
+        {
+          assertion: 'toBeInTheDocument',
+        },
+      ],
+    },
   ],
 
   invalid: [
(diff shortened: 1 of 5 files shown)


tests/lib/rules/prefer-explicit-assert.test.ts
```diff
@@ -144,14 +152,44 @@ ruleTester.run(RULE_NAME, rule, {
       code: `expect(getByText('foo')).toBeDefined()`,
       options: [
         {
-          assertion: 'toBeInDocument',
+          assertion: 'toBeInTheDocument',
```
Belco90, line 155: Nice one 💯

Belco90 (dismissed):
Looking really good! Thanks also for taking the time to refactor and extract those utils, really appreciated.

Belco90:
@skovy Should rule's doc be updated to reflect this change?

skovy:
I think the current docs are accurate, but I can update with more details about the specific logic and thinking around this rule.

skovy:
@Belco90 updated the docs, let me know if that was roughly what you had in mind or something else?

Belco90 (approved):
Nice, that's more than enough. Thanks!

Issue #222, opened 2020-08-24
[await-async-query] Support custom queries and wrappers


CreativeTechGuy:
As documented [here](https://testing-library.com/docs/react-testing-library/setup#add-custom-queries), you can create custom queries which extend the functionality of the basic query types. The rules of the eslint plugin should apply the same rules to these custom queries as they also apply.

I've identified the two lines that are the root cause of this issue:

* [This line](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/lib/rules/await-async-query.ts#L17) causes the `await-async-query` rule to be hard-coded to only specifically named queries and thus won't support any custom ones.
* [This line](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/lib/utils.ts#L16-L23) specifies the only libraries which are valid to import from. This prevents you to wrap testing-library and re-export it's functionality under a new name (as documented [here](https://testing-library.com/docs/dom-testing-library/api-helpers#custom-queries)).

I'm happy to submit a PR to fix this but I don't know how to best approach it.

* Should these restrictions be more laxed so that anything which matches the general pattern is included? This would be easier on the developer because there's less to configure and it "just works" but could have some false positives.
* Or should an additional configuration option be added to specify the names of imports and queries which should be added to the rule? This would be more flexible but require the developer to manually configure it whenever they add/change something.

I'm inclined to go with the first option as generally you want your tooling to be smart enough to just work. Any thoughts on this approach or any other recommended approaches?

*Oh and let me know if this problem affects any other rules that I haven't run into.*

CreativeTechGuy:
Found a few more occurrences of this issue:

* [no-await-sync-query](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/lib/rules/no-await-sync-query.ts#L8)
* [import sources](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/lib/utils.ts#L16-L23) affect many other (maybe all) rules so this should probably be a top level option.
* [no-manual-cleanup](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/lib/rules/no-manual-cleanup.ts#L17) has a separate hard-coded list of sources
[...]

PR #239, opened 2020-10-21
Move rules settings to ESLint shared config: part 2 - check imports

Belco90:
Relates to #198 

Second part which includes:
- shared config to be able to set module where Testing Library utils can be imported from
- mechanism to check if Testing Library is considered as imported depending on shared config
- aggressive reporting by default

So the plugin is gonna assume that Testing Library is imported:
- if no "testing-library/module" set, then it's always considered as imported
- else, then it's considered as imported if "testing-library" import or custom module set within "testing-library/module" setting found

Why? To follow the philosophy discussed [here](https://github.com/testing-library/eslint-plugin-testing-library/issues/222#issuecomment-679592434) so by default the plugin is really aggressive to avoid missing code that should be reported. Then, if the plugin is reporting things out of the scope for the user, they can config that scope properly through shared settings.

First thing that can be configured is the module from where Testing Library is considered as imported. So if the user has its own utils reexported from `@testing-library/framework` in `test-utils` module, they can set that up so the plugin will consider just `@testing-library/framework` and `test-utils` as modules where TL utils can be imported from.

This mechanism is implemented globally through `detectTestingLibraryUtils` so rules don't need to take care of this manually since the rule creator will prevent reporting when necessary.

Next step: shared config and mechanism for test file name patterns. You'll find more comments throughout this PR.

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -77,9 +77,14 @@ export function detectTestingLibraryUtils<
     };
 
     // Instructions for Testing Library detection.
-    // `ImportDeclaration` must be first in order to know if Testing Library
-    // is imported ASAP.
     const detectionInstructions: TSESLint.RuleListener = {
+      /**
+       * This ImportDeclaration rule listener will check if Testing Library related
+       * modules are loaded. Since imports happen first thing in a file, it's
+       * safe to use `isImportingTestingLibraryModule` and `isImportingCustomModule`
+       * since they will have corresponding value already updated when reporting other
+       * parts of the file.
+       */
       ImportDeclaration(node: TSESTree.ImportDeclaration) {
         if (!isImportingTestingLibraryModule) {
           // check only if testing library import not found yet so we avoid
@@ -102,14 +107,12 @@ export function detectTestingLibraryUtils<
     const ruleInstructions = ruleCreate(context, optionsWithDefault, helpers);
     const enhancedRuleInstructions: TSESLint.RuleListener = {};
 
-    // The order here is important too: detection instructions must come before
-    // than rule instructions to:
-    //  - detect Testing Library things before the rule is applied
-    //  - be able to prevent the rule about to be applied if necessary
     const allKeys = new Set(
       Object.keys(detectionInstructions).concat(Object.keys(ruleInstructions))
     );
 
+    // Iterate over ALL instructions keys so we can override original rule instructions
+    // to prevent their execution if conditions to report errors are not met.
     allKeys.forEach((instruction) => {
       enhancedRuleInstructions[instruction] = (node) => {
         if (instruction in detectionInstructions) {
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -1,12 +1,31 @@
 import { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
 
+export type TestingLibrarySettings = {
+  'testing-library/module'?: string;
+};
+
+export type TestingLibraryContext<
+  TOptions extends readonly unknown[],
+  TMessageIds extends string
+> = Readonly<
+  TSESLint.RuleContext<TMessageIds, TOptions> & {
+    settings: TestingLibrarySettings;
+  }
+>;
+
+export type EnhancedRuleCreate<
+  TOptions extends readonly unknown[],
+  TMessageIds extends string,
+  TRuleListener extends TSESLint.RuleListener = TSESLint.RuleListener
+> = (
+  context: TestingLibraryContext<TOptions, TMessageIds>,
+  optionsWithDefault: Readonly<TOptions>,
+  detectionHelpers: Readonly<DetectionHelpers>
+) => TRuleListener;
+
 export type DetectionHelpers = {
   getIsTestingLibraryImported: () => boolean;
   canReportErrors: () => boolean;
-};
-
-export type TestingLibrarySettings = {
-  'testing-library/module'?: string;
 };
 
 /**
@@ -16,23 +35,9 @@ export function detectTestingLibraryUtils<
   TOptions extends readonly unknown[],
   TMessageIds extends string,
   TRuleListener extends TSESLint.RuleListener = TSESLint.RuleListener
->(
-  ruleCreate: (
-    context: Readonly<
-      TSESLint.RuleContext<TMessageIds, TOptions> & {
-        settings: TestingLibrarySettings;
-      }
-    >,
-    optionsWithDefault: Readonly<TOptions>,
-    detectionHelpers: Readonly<DetectionHelpers>
-  ) => TRuleListener
-) {
+>(ruleCreate: EnhancedRuleCreate<TOptions, TMessageIds, TRuleListener>) {
   return (
-    context: Readonly<
-      TSESLint.RuleContext<TMessageIds, TOptions> & {
-        settings: TestingLibrarySettings;
-      }
-    >,
+    context: TestingLibraryContext<TOptions, TMessageIds>,
     optionsWithDefault: Readonly<TOptions>
   ): TSESLint.RuleListener => {
     let isImportingTestingLibraryModule = false;
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -1,7 +1,8 @@
 import { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
 
 export type DetectionHelpers = {
-  getIsImportingTestingLibrary: () => boolean;
+  getIsTestingLibraryImported: () => boolean;
+  canReportErrors: () => boolean;
 };
 
 /**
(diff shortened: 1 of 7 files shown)


lib/detect-testing-library-utils.ts
```diff
@@ -13,48 +18,84 @@ export function detectTestingLibraryUtils<
   TRuleListener extends TSESLint.RuleListener = TSESLint.RuleListener
 >(
   ruleCreate: (
-    context: Readonly<TSESLint.RuleContext<TMessageIds, TOptions>>,
+    context: Readonly<
+      TSESLint.RuleContext<TMessageIds, TOptions> & {
+        settings: TestingLibrarySettings;
```
Belco90, line 23: I can't think of a better way of typing the settings!

lib/detect-testing-library-utils.ts
```diff
@@ -13,48 +18,84 @@ export function detectTestingLibraryUtils<
...
+      }
+    >,
     optionsWithDefault: Readonly<TOptions>
-  ): TRuleListener => {
-    let isImportingTestingLibrary = false;
+  ): TSESLint.RuleListener => {
```
Belco90, line 37: I've replaced returned type from `TRuleListener` to `TSESLint.RuleListener` since `detectTestingLibraryUtils` modify the received rules listener to add or remove some so it can end being a different set of rule listener.

lib/detect-testing-library-utils.ts
```diff
@@ -13,48 +18,84 @@ export function detectTestingLibraryUtils<
...
-    // TODO: init here options based on shared ESLint config
+    // Init options based on shared ESLint settings
+    const customModule = context.settings['testing-library/module'];
 
-    // helpers for Testing Library detection
+    // Helpers for Testing Library detection.
     const helpers: DetectionHelpers = {
-      getIsImportingTestingLibrary() {
-        return isImportingTestingLibrary;
+      getIsTestingLibraryImported() {
```
Belco90, line 46: I should probably add a jsdoc here explaining the motivation behind this behavior that I wrote in this PR description, right?

lib/detect-testing-library-utils.ts
```diff
@@ -13,48 +18,84 @@ export function detectTestingLibraryUtils<
...
+      },
+      canReportErrors() {
+        return this.getIsTestingLibraryImported();
       },
     };
 
-    // instructions for Testing Library detection
+    // Instructions for Testing Library detection.
+    // `ImportDeclaration` must be first in order to know if Testing Library
+    // is imported ASAP.
```
Belco90, line 60: I'm still surprised this work. I can elaborate more on this but basically:

1. `detectTestingLibraryUtils` inject a detection rule listener for "ImportDeclaration" which will try to find a Testing Library import
2. before running checks from the original rule listener, `detectTestingLibraryUtils` will make sure Testing Library is imported to prevent original rule listener to be executed

I think this actually works not because the order of the rule listener definitions here, but because imports always happen first in the code linted. I just realized about this rereading my own code and try different orders here and in the concat below. I think I'm gonna remove this comment then to avoid any misunderstanding.

lib/detect-testing-library-utils.ts
```diff
@@ -13,48 +18,84 @@ export function detectTestingLibraryUtils<
...
 
     // update given rule to inject Testing Library detection
     const ruleInstructions = ruleCreate(context, optionsWithDefault, helpers);
-    const enhancedRuleInstructions = Object.assign({}, ruleInstructions);
+    const enhancedRuleInstructions: TSESLint.RuleListener = {};
+
+    // The order here is important too: detection instructions must come before
+    // than rule instructions to:
+    //  - detect Testing Library things before the rule is applied
+    //  - be able to prevent the rule about to be applied if necessary
```
Belco90, line 87: Second part of my previous comment. As I said, I don't finally think this order matters.

lib/detect-testing-library-utils.ts
```diff
@@ -13,48 +18,84 @@ export function detectTestingLibraryUtils<
...
+    const allKeys = new Set(
+      Object.keys(detectionInstructions).concat(Object.keys(ruleInstructions))
+    );
 
-    Object.keys(detectionInstructions).forEach((instruction) => {
-      (enhancedRuleInstructions as TSESLint.RuleListener)[instruction] = (
-        node
-      ) => {
+    allKeys.forEach((instruction) => {
```
Belco90, line 92: Now I iterate through **all** keys to be able to prevent original rule instructions.

lib/detect-testing-library-utils.ts
```diff
@@ -13,48 +18,84 @@ export function detectTestingLibraryUtils<
...
+      enhancedRuleInstructions[instruction] = (node) => {
         if (instruction in detectionInstructions) {
           detectionInstructions[instruction](node);
         }
 
-        if (ruleInstructions[instruction]) {
+        if (helpers.canReportErrors() && ruleInstructions[instruction]) {
```
Belco90, line 98: Here is where the plugin prevents original rule instructions to be executed if conditions are not met!

lib/rules/no-node-access.ts
```diff
@@ -25,11 +25,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
   },
   defaultOptions: [],
 
-  create: (context, _, helpers) => {
+  create(context) {
     function showErrorForNodeAccess(node: TSESTree.MemberExpression) {
       isIdentifier(node.property) &&
         ALL_RETURNING_NODES.includes(node.property.name) &&
-        helpers.getIsImportingTestingLibrary() &&
```
Belco90, line 32: This is not necessary anymore! ✨

tests/create-testing-library-rule.test.ts
```diff
@@ -0,0 +1,93 @@
+import { createRuleTester } from './lib/test-utils';
```
Belco90, line 1: I added this test file to check both `createTestingLibraryRule` + `detectTestingLibraryUtils` together, covering as much edge cases as possible.

tests/fake-rule.ts
```diff
@@ -0,0 +1,42 @@
+/**
```
Belco90, line 1: Fake rule trying to report stuff asap to be able to run the previous test file.

tests/lib/rules/no-node-access.test.ts
```diff
@@ -51,22 +51,38 @@ ruleTester.run(RULE_NAME, rule, {
         within(signinModal).getByPlaceholderText('Username');
       `,
     },
-    {
+    /*{
```
Belco90, line 54: I'll reenable this test in the PR part 3 since there is where I'm gonna include the test file name pattern to report or not rules, so this will be solved then.

tests/lib/rules/no-node-access.test.ts
```diff
@@ -51,22 +51,38 @@ ruleTester.run(RULE_NAME, rule, {
...
+      },
     },
   ],
   invalid: [
+    {
+      code: `
+      // case: without importing TL (aggressive reporting)
+      const closestButton = document.getElementById('submit-btn')
+      expect(closestButton).toBeInTheDocument();
+      `,
```
Belco90, line 83: This was a valid case before but it isn't anymore. I'll find more like this in other rules.

lib/detect-testing-library-utils.ts
```diff
@@ -13,48 +18,84 @@ export function detectTestingLibraryUtils<
...
-        return isImportingTestingLibrary;
+      getIsTestingLibraryImported() {
+        if (!customModule) {
+          return true;
+        }
+
+        return isImportingTestingLibraryModule || isImportingCustomModule;
+      },
+      canReportErrors() {
+        return this.getIsTestingLibraryImported();
```
gndelia, line 54: it's been a long time since I've seen a `this` 😄

tests/fake-rule.ts
```diff
@@ -0,0 +1,42 @@
...
+          node,
+          messageId: 'fakeError',
+        });
+      }
+    };
+
+    return {
+      'CallExpression Identifier': reportRenderIdentifier,
+    };
+  },
```
gndelia, line 41: Can we add to this rule a bunch of `Import` keys? This way we could test that the `Import` key we check does not override any `Import` the user could've defined in their rules. For instance, an example I could think of is a rule that verifies that some deprecated methods are not imported (like the `prefer-wait-for` rule)

lib/detect-testing-library-utils.ts
gndelia: I'm a bit lost what's the difference between the method `canReportErrors` and `getIsTestingLibraryImported` - given the first one just calls the second

gndelia: in addition to that, if I understand correctly, we would just execute the rules if TL is imported, otherwise, nothing gets executed - is that correct?

Belco90: I'm anticipating next PR with `canReportErrors` so it warps all necessary checks to know if rules should report errors or not. That's just knowing if Testing Library is imported from now, but in the next PR I'll add the shared option to customize the file names pattern which should be checked by regexp, so the user can set to just report from this plugin on files named `*.test.js`. This is interesting when you have different kind of tests files (.test.js for Testing Library and .spec.js for Cypress) or just to reduce the scope of the plugin. So in the next PR `canReportErrors` will check more things and will be in charge of wrap all necessary checks in the future if others needed.

> in addition to that, if I understand correctly, we would just execute the rules if TL is imported, otherwise, nothing gets executed - is that correct?

Yes, but by default TL is always considered as imported due to the "aggressive reporting" mode.

tests/fake-rule.ts
Belco90: Sure, I can do that.

gndelia (approved):
Looking great 🚀

Belco90:
Thanks for your review, merging now! Can't wait to have the rest of the shared config, it's gonna make rules implementation much easier.

My idea is to create the PR for the option to customize the test files name pattern next week, reenabling the commented test out in this PR and updating the login for `canReportErrors`. Then I'll update couple of rules to use the new rule creator.

After that I'll add the last shared config for defining custom `render` methods, which will be the most complex one.

Issue #219, opened 2020-08-19
feature request: no-await-fire-event


benmonro:
We accidentally enabled await-fire-event in our react project and discovered that awaiting fireEvent can actually cause tests to be slow and flaky as a result (removing the await fixes it). so was thinking that an opposite rule or even a config option to the await-fire-event to work in opposite mode would be useful

Belco90:
I guess this could be wrapped into a `no-await-sync-utils` to do the opposite than `await-async-utils`. I don't know if there are any other sync util than `fireEvent` that could be included there, but with this approach we can include any new sync util here. What do you think?

benmonro:
Works for me. My only thought would be that you wouldn't want the rules to conflict with each other for vue users but I'm not a cute user so that's not going to affect me. 🙂

Belco90:
I haven't thought about that conflict. Maybe we need to create this rule with a config options for disabling specific methods to be check so Vue users can enable this rule but disabling `fireEvent`.

benmonro:
That would work.

PR #240, opened 2020-10-22
feat(rule): add no-await-sync-events rule

J-Huang:
Feature request:
https://github.com/testing-library/eslint-plugin-testing-library/issues/219

All functions in "fireEvent" and "userEvent" are synchronous, meaning it's wrong to add "await" operator for those function calls, with an exception of userEvent.type.

This new rule "no-await-sync-events" disallows adding "await" for those function calls in order to avoid errors.


Belco90:
Thanks! I'll try to review it during the weekend. Although I don't know if there is something pending since it's a draft PR.

J-Huang:
@Belco90 it's in draft because would like to have @benmonro take a look first since he was the owner of the feature request. Thanks.

benmonro (dismissed):
LGTM thanks @J-Huang

docs/rules/no-await-sync-events.md
```diff
@@ -0,0 +1,53 @@
+# Disallow unnecessary `await` for sync events (no-await-sync-events)
+
+Ensure that sync events are not awaited unnecessarily.
+
+## Rule Details
+
+Functions in the event object provided by Testing Library, including
+fireEvent and userEvent, do NOT return Promise, with an exception of
+`userEvent.type`. Some examples are:
```
Belco90, line 9: This is technically true, but you don't need to always wait for it. From [user-event doc](https://github.com/testing-library/user-event#typeelement-text-options):

> To be clear, userEvent.type always returns a promise, but you only need to await the promise it returns if you're using the delay option. Otherwise everything runs synchronously and you can ignore the promise.

So should this rule report about `userEvent.type` if no delay option passed? I think so

lib/utils.ts
```diff
@@ -63,6 +63,11 @@ const ASYNC_UTILS = [
   'waitForDomChange',
 ];
 
+const ASYNC_EVENTS = [
```
Belco90, line 66: This are SYNC_EVENTS actually, right?

tests/lib/rules/no-await-sync-events.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    case 'userEvent':
+      eventFunctions = eventFunctions.concat(userEventFunctions.map((f: string): string => `${event}.${f}`));
+      break;
+    default:
+      eventFunctions.push(`${event}.anyFunc`);
+  }
+});
+
+ruleTester.run(RULE_NAME, rule, {
+  valid: [
```
Belco90, line 120: You need to include more tests for checking `userEvent` and `fireEvent` non-related to Testing Library are not reported so they are valid with `await` operator.

There is an ongoing refactor for v4 of the plugin which actually will check that for all rules out of the box, so I don't know if you prefer to wait for that version to be released.

tests/lib/rules/no-await-sync-events.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        await userEvent.type('foo')
+      }
+      `,
+    },
+  ],
+
+  invalid: [
+    // sync events with await operator are not valid
+    ...eventFunctions.map(func => ({
+      code: `() => {
```
Belco90, line 146: Can you write more realistic code here? So you end up with something like:

```javascript
import { fireEvent } from '@testing-library/framework'
import userEvent from '@testing-library/user-event'
import MyComponent from './MyComponent'

it('should report sync event awaited', async () => {
  render(<MyComponent />)
  
  await ${func}('foo')
  await findByQuery('some element')
})
```

and then check in the errors which line the error should appear.

Belco90 (changes requested):
The rule implementation itself seems fine, but we need to improve few things:
- handle `userEvent.type` special cases better
- make tests more confident
- add the rule to README

docs/rules/no-await-sync-events.md
gndelia: I agree with that ☝️  we should only await if the delay option is used. It would be great if we can cover that scenario with the rule

lib/utils.ts
gndelia: I wouldn't say there are events either - there are helpers or utilities.. Events would be "tab", "input", "change", "click" and so on... 🤔

tests/lib/rules/no-await-sync-events.test.ts
J-Huang: @Belco90 
Do you mean if userEvent or fireEvent are user named objects, it should not apply the rule? Or, if extra functions are added to those objects, for those cases, await may be valid?
In order to ensure that, it has to check if the property name in userEvent or fireEvent is one of the expected event names, such as click, type, tab and so on. Am I misunderstanding the question?

lib/utils.ts
J-Huang: initially thought it should be no-await-sync-utils, but feel not so correct...I can rename it to utils if it makes more sense.

docs/rules/no-await-sync-events.md
J-Huang: Agree. 
the rule is, only if userEvent.type with delay option, await is valid. For all the other cases, it disallows await.

tests/lib/rules/no-await-sync-events.test.ts
J-Huang: updated

tests/lib/rules/no-await-sync-events.test.ts
Belco90: The former, so we don't report const declared by the user with the same name. As mentioned, the internal refactor for v4 will pass some helpers to all rules to check this generically, so I'm not sure if you prefer to wait for that.

docs/rules/no-await-sync-events.md
```diff
@@ -45,6 +53,12 @@ const bar = () => {
   userEvent.tab();
   // ...
 };
+
+const baz = () => {
+  // ...
+  await userEvent.type(textInput, 'abc', {delay: 1000});
```
Belco90, line 59: Can you add an extra correct example for `userEvent.type` without await and delay please?

Belco90 (changes requested):
Just an extra correct example and I think it would be ready, apart from what we want to do with checking if reported objects are from Testing Library modules or not.

docs/rules/no-await-sync-events.md
J-Huang: updated

tests/lib/rules/no-await-sync-events.test.ts
J-Huang: Thanks for the clarification.
will refactor the code by then if needed.

J-Huang:
Not sure why CI stops. My main computer died, so I just made a commit through github web page UI to my forked repo.

README.md
```diff
@@ -137,6 +137,7 @@ To enable this configuration use the `extends` property in your
 | [await-async-utils](docs/rules/await-async-utils.md)                   | Enforce async utils to be awaited properly                                 | ![recommended-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |                    |
 | [await-fire-event](docs/rules/await-fire-event.md)                     | Enforce async fire event methods to be awaited                             | ![vue-badge][]                                                            |                    |
 | [consistent-data-testid](docs/rules/consistent-data-testid.md)         | Ensure `data-testid` values match a provided regex.                        |                                                                           |                    |
+| [no-await-sync-events](docs/rules/no-await-sync-events.md)             | Disallow unnecessary `await` for sync events                               | ![recommended-badge][] ![angular-badge][] ![react-badge][]                |                    |
```
Belco90, line 140: This new rule is not included in any preset by default. Can you remove recommended, angular and react badges please?

J-Huang: updated

Belco90 (approved):

Belco90:
@all-contributors please add @J-Huang for code, test and docs

Issue #241, opened 2020-10-22
prefer-screen-queries doesn't work with `within`


benmonro:
ran into a false negative that I thought i'd share.  consider the following code: 

```jsx
import { screen, within } from "@testing-library/react";

const container = within(screen.getByRole("foo"));

expect(container.getByText("bar"));
expect(container.getByText("baz"));
```

these are being reported as invalid, but they are valid. can this be supported in the rule?

gndelia:
According to the [docs](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/prefer-screen-queries.md) these are valid and supported 🤔 

I don't see an exact test matching this scenario, I will try to take a look in the incoming days

gndelia:
Hey, @benmonro  I created this ☝️  PR with your scenario and, unless I'm missing something, it's passing. Maybe you need to upgrade to a newer version? Which version are you using of the plugin?

In addition to that, could you verify if the code is exactly that one, or if the rule that fails is this one or another? Thanks!

benmonro:
yeah doesn't seem to be an issue anymore, if it pops up again i'll reopen

PR #243, opened 2020-10-24
test: adding test for possible false positive

gndelia:
adding a test to see if I can reproduce #241

### DIFF
--- a/tests/lib/rules/prefer-screen-queries.test.ts
+++ b/tests/lib/rules/prefer-screen-queries.test.ts
@@ -125,6 +125,16 @@ ruleTester.run(RULE_NAME, rule, {
         render(foo, { baseElement: treeA }).${queryMethod}()
       `,
     })),
+    {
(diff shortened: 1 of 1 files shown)


Belco90:
Can you remove the draft state from this one so we merge it?

gndelia:
not sure if it makes sense to merge it. I opened in order to test the exact scenario (expression by expression 😋 ) mentioned in the bug, but I was unable to reproduce it. There's another test with the same scenario, but without checking the import; we would add that as we migrate the PRs in v4, so that would make this PR useless as the exact scenario will become covered (And more!)

If there's no response soon in the ticket, I guess we could just close it

Belco90:
Ok, I see. Let's wait for v4 then.

Belco90:
Ok, I see. Let's wait for v4 then.

gndelia:
closing as the issue was closed

PR #244, opened 2020-10-28
Move rules settings to ESLint shared config: part 3 - check reported file name

Belco90:
Relates to #198 

Third part which includes:
- checking file name in testing library detection so we only report desired files
- new setting to customize file name pattern for desired reported files. It defaults to `.(test|spec).(t|j)sx?` (same as jest)
- new RuleTester for Testing Library which makes sure valid and invalid test cases have default config (just file name for now)
- common parserOptions for tests updated

With this PR we already have 2/3 necessary settings for v4: module where testing library utils can be imported and file name pattern for desired reported files. The final one will be the setting for customizing other methods than `render` to be considered as Testing Library render (e.g. `renderWithRedux`, `setUp`).

We are missing also the "aggressive reporting" mode for Testing Library queries so everything starting by `findBy*`, `getBy*` and `queryBy*` are considered queries, not just different combinations from Testing Library.

So, after this one is merged, next steps are:

1. refactor rules not reporting anything related to `render` or queries (that's in my analysis attached on #198 )
2. implement the setting for customize `render` method and use it in one existing rule
3. implement "aggressive reporting" for queries and use it in one existing rule
4. refactor remaining rules

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -2,6 +2,7 @@ import { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
 
 export type TestingLibrarySettings = {
   'testing-library/module'?: string;
+  'testing-library/file-name'?: string;
 };
 
 export type TestingLibraryContext<
@@ -25,8 +26,11 @@ export type EnhancedRuleCreate<
 
 export type DetectionHelpers = {
   getIsTestingLibraryImported: () => boolean;
+  getIsValidFileName: () => boolean;
   canReportErrors: () => boolean;
 };
+
+const DEFAULT_FILE_NAME_PATTERN = '^.*\\.(test|spec)\\.[jt]sx?$';
 
 /**
  * Enhances a given rule `create` with helpers to detect Testing Library utils.
@@ -45,6 +49,9 @@ export function detectTestingLibraryUtils<
 
     // Init options based on shared ESLint settings
     const customModule = context.settings['testing-library/module'];
+    const fileNamePattern =
+      context.settings['testing-library/file-name'] ??
+      DEFAULT_FILE_NAME_PATTERN;
 
     // Helpers for Testing Library detection.
     const helpers: DetectionHelpers = {
@@ -69,10 +76,20 @@ export function detectTestingLibraryUtils<
       },
 
       /**
+       * Gets if file name being analyzed is valid or not.
+       *
+       * This is based on "testing-library/file-name" setting.
+       */
+      getIsValidFileName() {
+        const fileName = context.getFilename();
+        return !!fileName.match(fileNamePattern);
+      },
+
+      /**
        * Wraps all conditions that must be met to report rules.
        */
       canReportErrors() {
-        return this.getIsTestingLibraryImported();
+        return this.getIsTestingLibraryImported() && this.getIsValidFileName();
       },
     };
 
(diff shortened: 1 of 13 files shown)


tests/lib/rules/consistent-data-testid.test.ts
```diff
@@ -1,11 +1,7 @@
 import { createRuleTester } from '../test-utils';
 import rule, { RULE_NAME } from '../../../lib/rules/consistent-data-testid';
 
-const ruleTester = createRuleTester({
-  ecmaFeatures: {
```
Belco90, line 5: We don't need this anymore, set by default on `createRuleTester`.

tests/lib/rules/no-node-access.test.ts
```diff
@@ -51,18 +47,16 @@ ruleTester.run(RULE_NAME, rule, {
...
-      //  to track where the nodes are coming from. This one wasn't reported before
-      //  just because this code is not importing TL module, but that's a really
-      //  brittle check. Instead, this one shouldn't be reported since `children`
-      //  it's just a property not related to a node
+    {
       code: `
         const Component = props => {
           return <div>{props.children}</div>
         }
       `,
```
Belco90, line 55: This test is reenabled now, but just because the file name doesn't match so there is nothing reported here. As mentioned in my original comment there, the rule implementation is really brittle and should make sure node methods are coming from an actual node and not a var or prop. That's out of the scope of this refactor so I'm not gonna fix it for now (it shouldn't happen that often with file-name setting tho).

tests/lib/test-utils.ts
```diff
@@ -1,14 +1,49 @@
 import { resolve } from 'path';
 import { TSESLint } from '@typescript-eslint/experimental-utils';
 
+const DEFAULT_TEST_CASE_CONFIG = {
+  filename: 'MyComponent.test.js',
+};
+
+class TestingLibraryRuleTester extends TSESLint.RuleTester {
```
Belco90, line 8: Our custom RuleTester just injects default test case config for each valid/invalid test case so we don't need to set `filename` config to a valid value to be reported.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
```
Belco90, line 3: When creating `TestingLibraryRuleTester` I ran into [this issue](https://stackoverflow.com/questions/50203369/class-constructor-cannot-be-invoked-without-new-typescript-with-commonjs/50203532#50203532) so I had to update the target to "es6".

lib/detect-testing-library-utils.ts
```diff
@@ -45,6 +49,9 @@ export function detectTestingLibraryUtils<
 
     // Init options based on shared ESLint settings
     const customModule = context.settings['testing-library/module'];
+    const fileNamePattern =
+      context.settings['testing-library/file-name'] ??
```
gndelia, line 53: should the setting be `'testing-library/file-name-pattern'`? Or adding the word pattern somewhere 😄  so the intention is clearer

gndelia (approved):
Looks great to me! I left a small suggestion but not a blocker. You've got a great job ongoing for the v4 🚀

lib/detect-testing-library-utils.ts
Belco90: You are absolutely right. I'll rename it in #247 which is the next PR where I already renamed this from `file-name` to `filename`.

PR #247, opened 2020-11-01
Move rules settings to ESLint shared config: part 4 - refactor no-dom-import rule

Belco90:
Relates to #198 

4th part which includes:
- `no-dom-import` rule refactored to sue new rule creator
- more test cases
- new detection helpers for import nodes

Since `no-dom-import` had the best implementation to check Testing Library imports, I've moved those within detection helpers so:
- the plugin now reports based on `import` or `require` (the latter was missing)
- there are new detection helpers to get the node where the imports where found

As result, now this rule is much clear and simple, and detection helper is more flexible when finding Testing Library imports.

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -2,6 +2,7 @@ import { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
 
 export type TestingLibrarySettings = {
   'testing-library/module'?: string;
+  'testing-library/file-name'?: string;
 };
 
 export type TestingLibraryContext<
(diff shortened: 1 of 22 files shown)


lib/detect-testing-library-utils.ts
```diff
@@ -1,8 +1,9 @@
 import { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
+import { isLiteral } from './node-utils';
 
 export type TestingLibrarySettings = {
   'testing-library/module'?: string;
-  'testing-library/file-name'?: string;
+  'testing-library/filename-pattern'?: string;
```
Belco90, line 6: Renamed to `filename-pattern` as discussed in previous PR.

timdeschryver (approved):

PR #248, opened 2020-11-01
Move rules settings to ESLint shared config: part 5 - refactor consistent-data-testid rule

Belco90:
Relates to #198

5th part which includes:
- refactor `consistent-data-testid` rule
- add new option to Testing Library rule creator to skip reporting validation checks


Belco90:
Actually I think this is not necessary at all. This rule apply to other files than testing ones, so it should not be created with `createTestingLibraryRule`. I'll cancel this PR and just add a comment about why this one doesn't use the new rule creator.

Belco90:
Changes reverted. This rule doesn't use `createTestingLibraryRule` now per reasons added in the comment.

gndelia (approved):
I guess there's nothing to review 😄

PR #249, opened 2020-11-03
Move rules settings to ESLint shared config: part 5 - refactor no-manual-cleanup rule

Belco90:
Relates to #198 

6th part which includes:
-  refactor `no-manual-cleanup` rule to use custom rule creator + detection helpers
- extract some detection helpers for getting imported module name related to Testing Library (either official or custom modules)
- improve rule to detect custom module from settings

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -1,9 +1,5 @@
 import { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
-import {
-  getImportModuleName,
-  isLiteral,
-  ModuleImportation,
-} from './node-utils';
+import { getImportModuleName, isLiteral, ImportModuleNode } from './node-utils';
 
 export type TestingLibrarySettings = {
   'testing-library/module'?: string;
@@ -30,8 +26,8 @@ export type EnhancedRuleCreate<
 ) => TRuleListener;
 
 export type DetectionHelpers = {
-  getTestingLibraryImportNode: () => ModuleImportation | null;
-  getCustomModuleImportNode: () => ModuleImportation | null;
+  getTestingLibraryImportNode: () => ImportModuleNode | null;
+  getCustomModuleImportNode: () => ImportModuleNode | null;
   getTestingLibraryImportName: () => string | undefined;
   getCustomModuleImportName: () => string | undefined;
   getIsTestingLibraryImported: () => boolean;
@@ -53,8 +49,8 @@ export function detectTestingLibraryUtils<
     context: TestingLibraryContext<TOptions, TMessageIds>,
     optionsWithDefault: Readonly<TOptions>
   ): TSESLint.RuleListener => {
-    let importedTestingLibraryNode: ModuleImportation | null = null;
-    let importedCustomModuleNode: ModuleImportation | null = null;
+    let importedTestingLibraryNode: ImportModuleNode | null = null;
+    let importedCustomModuleNode: ImportModuleNode | null = null;
 
     // Init options based on shared ESLint settings
     const customModule = context.settings['testing-library/module'];
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -30,8 +30,8 @@ export type EnhancedRuleCreate<
 ) => TRuleListener;
 
 export type DetectionHelpers = {
-  getTestingLibraryImportNode: () => ModuleImportation;
-  getCustomModuleImportNode: () => ModuleImportation;
+  getTestingLibraryImportNode: () => ModuleImportation | null;
+  getCustomModuleImportNode: () => ModuleImportation | null;
   getTestingLibraryImportName: () => string | undefined;
   getCustomModuleImportName: () => string | undefined;
   getIsTestingLibraryImported: () => boolean;
@@ -53,8 +53,8 @@ export function detectTestingLibraryUtils<
     context: TestingLibraryContext<TOptions, TMessageIds>,
     optionsWithDefault: Readonly<TOptions>
   ): TSESLint.RuleListener => {
-    let importedTestingLibraryNode: ModuleImportation = null;
-    let importedCustomModuleNode: ModuleImportation = null;
+    let importedTestingLibraryNode: ModuleImportation | null = null;
+    let importedCustomModuleNode: ModuleImportation | null = null;
 
     // Init options based on shared ESLint settings
     const customModule = context.settings['testing-library/module'];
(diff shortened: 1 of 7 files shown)


lib/detect-testing-library-utils.ts
```diff
@@ -69,6 +70,12 @@ export function detectTestingLibraryUtils<
       getCustomModuleImportNode() {
         return importedCustomModuleNode;
       },
+      getTestingLibraryImportName() {
+        return getImportModuleName(importedTestingLibraryNode);
+      },
+      getCustomModuleImportName() {
+        return getImportModuleName(importedCustomModuleNode);
+      },
```
Belco90, line 78: Two new helpers to get the name of Testing Library or Custom Module imports 💪

lib/rules/no-dom-import.ts
```diff
@@ -82,36 +78,22 @@ export default createTestingLibraryRule<Options, MessageIds>({
 
     return {
       'Program:exit'() {
+        const importName = helpers.getTestingLibraryImportName();
```
Belco90, line 81: Reusing the new helper in this rule already refactored.

lib/rules/no-manual-cleanup.ts
```diff
@@ -122,24 +99,29 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
         }
+      }
+    }
 
-        if (defaultRequireFromTestingLibrary) {
-          const references = context
-            .getDeclaredVariables(defaultRequireFromTestingLibrary.parent)[0]
-            .references.slice(1);
+    return {
+      'Program:exit'() {
```
Belco90, line 110: This rule is so much simpler now: it just needs to focus on checks around the node imports, but doesn't need to handle anything to find them.

tests/fake-rule.ts
```diff
@@ -55,14 +55,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
       ImportDeclaration: checkImportDeclaration,
       'Program:exit'() {
         const importNode = helpers.getCustomModuleImportNode();
+        const importName = helpers.getCustomModuleImportName();
```
Belco90, line 58: Reusing the helper in the fake rule too!

lib/node-utils.ts
```diff
@@ -220,3 +224,29 @@ export function isRenderVariableDeclarator(
 
   return false;
 }
+
+// TODO: extract into types file?
+export type ModuleImportation =
```
gndelia, line 229: should it be

```suggestion
export type ModuleImport =
```

?

Importation sounds like spanglish (?)

lib/node-utils.ts
```diff
@@ -220,3 +224,29 @@ export function isRenderVariableDeclarator(
...
+  | TSESTree.ImportDeclaration
+  | TSESTree.CallExpression;
+
+export function getImportModuleName(
+  node: ModuleImportation | undefined | null
+): string | undefined {
+  // import node of shape: import { foo } from 'bar'
+  if (
+    node?.type === AST_NODE_TYPES.ImportDeclaration &&
```
gndelia, line 238: we've usually centered this checks on `isCallExpression` in `node-utils`

gndelia: ups sorry, this comment was supposed to be in line 246. For this check there's no function, but you could add it (there are all type guards)

lib/node-utils.ts
```diff
@@ -17,6 +17,7 @@ export function isNewExpression(
   return node && node.type === 'NewExpression';
 }
 
+// TODO: remove this one and use ASTUtils one instead
```
gndelia, line 20: 🧐

gndelia: oh man, I saw your `ASTUtils` now... ignore my previous comments. It's awesome they export that 🚀

lib/node-utils.ts
Belco90: Good suggestion. I renamed that type so many times that I ended up with a shitty name 😂

lib/node-utils.ts
Belco90: Actually I was thinking about adding a generic one, as I'm tired of having one function per check 😂 

Shouldn't be possible to implement some method like this?

```typescript
function getIsNodeType(
  node: TSESTree.Node,
  expectedType: keyof typeof AST_NODE_TYPES
): boolean {
  return node?.type === AST_NODE_TYPES[expectedType];
}
```

ccing our TS expert @timdeschryver 😅

gndelia: the problem with that is that you lose the type inference I think, the `x is TheNodeTypeYouConfirm`  from the [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html)

Belco90: I guess I could extract `keyof tyepof AST_NODE_TYPES` into `T` and update return type to `node is T` but to be honest no clue if I'm trying to go too far here. Probably won't work so I'll extract it into its own method in a new PR where I'll add tests for utils.

lib/node-utils.ts
```diff
@@ -150,6 +154,12 @@ export function isArrayExpression(
   return node?.type === AST_NODE_TYPES.ArrayExpression;
 }
 
+export function isImportDeclaration(
```
Belco90, line 157: Added util for knowing if a node is ImportDeclaration 👍

lib/node-utils.ts
Belco90: Didn't realized we already had `isCallExpression`. Using this one now, and `isImportDeclaration` added too.

lib/rules/no-dom-import.ts
```diff
@@ -40,11 +37,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
 
   create(context, [framework], helpers) {
     function report(
-      node: TSESTree.ImportDeclaration | TSESTree.Identifier,
+      node: TSESTree.ImportDeclaration | TSESTree.CallExpression,
       moduleName: string
     ) {
       if (framework) {
-        const isRequire = isIdentifier(node) && node.name === 'require';
```
gndelia, line 47: small doubt here: I guess some of the rules had support for require, but that only applied to a few. Are we dropping the support for `require` in all of them?
I am not opposed to it, I'm wondering if we should document it somewhere? (not sure if the docs for this rule and others showed it as part of the valid examples)

gndelia: yup in this case the `require` examples are in the docs https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/no-dom-import.md

gndelia: oh ups nevermind, I saw the helpers consider require as well, great! 🚀  you're one step ahead

gndelia (approved):
🚀 🎖️

lib/rules/no-dom-import.ts
Belco90: Sorry, I should have pointed this with a comment on this PR. But yest, I basically extracted the imported module checks from this rule (which were the best ones) into detection helpers. So now all refactores rules automatically getting the imports from either `import` or `require` syntax ✨

PR #251, opened 2020-11-05
feat: add new settings for prefer-user-event

gndelia:
refactor draft to include all the changes from #198 

this should make all the current scenarios and tests to pass, using the new helpers instead. What's pending:

- add tests with custom import libraries (the new ability from v4)
- add tests scenarios with `const { fireEvent } = require('module')` (the original implementation was not considering it, but now should be easier)
- add tests scenarios with `const rtl= require('module'); rtl.fireEvent()` (the original implementation was not considering it, but now should be easier)

I gotta say the rule's code is cleaner now 🎉 🚀

### DIFF
--- a/lib/node-utils.ts
+++ b/lib/node-utils.ts
@@ -253,22 +253,3 @@ export function getImportModuleName(
     return node.arguments[0].value;
   }
 }
-
-export function getSpecifierFromImport(
-  node: ImportModuleNode,
-  specifierName: string
-) {
-  if (isImportDeclaration(node)) {
-    const namedExport = node.specifiers.find(
-      (node) => isImportSpecifier(node) && node.imported.name === specifierName
-    );
-    // it is "import { foo } from 'baz'""
-    if (namedExport) {
-      return namedExport;
-    }
-    // it could be "import * as rtl from 'baz'"
-    return node.specifiers.find((n) => isImportNamespaceSpecifier(n));
-  } else {
-    // TODO make it work for require
-  }
-}
--- a/lib/node-utils.ts
+++ b/lib/node-utils.ts
@@ -253,3 +253,22 @@ export function getImportModuleName(
     return node.arguments[0].value;
   }
 }
+
+export function getSpecifierFromImport(
+  node: ImportModuleNode,
+  specifierName: string
+) {
+  if (isImportDeclaration(node)) {
+    const namedExport = node.specifiers.find(
+      (node) => isImportSpecifier(node) && node.imported.name === specifierName
+    );
+    // it is "import { foo } from 'baz'""
+    if (namedExport) {
+      return namedExport;
+    }
+    // it could be "import * as rtl from 'baz'"
+    return node.specifiers.find((n) => isImportNamespaceSpecifier(n));
+  } else {
+    // TODO make it work for require
+  }
+}
--- a/lib/rules/prefer-user-event.ts
+++ b/lib/rules/prefer-user-event.ts
@@ -1,9 +1,9 @@
-import { ESLintUtils, TSESTree } from '@typescript-eslint/experimental-utils';
-import { getDocsUrl, hasTestingLibraryImportModule } from '../utils';
+import { TSESTree } from '@typescript-eslint/experimental-utils';
+import { createTestingLibraryRule } from '../create-testing-library-rule';
 import {
-  isImportSpecifier,
   isIdentifier,
   isMemberExpression,
+  getSpecifierFromImport,
 } from '../node-utils';
 
 export const RULE_NAME = 'prefer-user-event';
@@ -65,7 +65,7 @@ function buildErrorMessage(fireEventMethod: string) {
 
 const fireEventMappedMethods = Object.keys(MappingToUserEvent);
 
-export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
+export default createTestingLibraryRule<Options, MessageIds>({
   name: RULE_NAME,
   meta: {
     type: 'suggestion',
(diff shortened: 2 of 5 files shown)


lib/node-utils.ts
```diff
@@ -253,3 +253,22 @@ export function getImportModuleName(
...
+    // it is "import { foo } from 'baz'""
+    if (namedExport) {
+      return namedExport;
+    }
+    // it could be "import * as rtl from 'baz'"
+    return node.specifiers.find((n) => isImportNamespaceSpecifier(n));
+  } else {
+    // TODO make it work for require
+  }
+}
```
gndelia, line 274: if we move this function into the detect-import-testing-library utils, instead of passing `node` I could get it directly from the helper. Not sure which option is best.

I still have to make it work for both `require` scenarios mentioned

lib/node-utils.ts
```diff
@@ -253,3 +253,22 @@ export function getImportModuleName(
...
+
+export function getSpecifierFromImport(
+  node: ImportModuleNode,
+  specifierName: string
+) {
+  if (isImportDeclaration(node)) {
+    const namedExport = node.specifiers.find(
+      (node) => isImportSpecifier(node) && node.imported.name === specifierName
+    );
+    // it is "import { foo } from 'baz'""
```
Belco90, line 265: What about `import { foo as qux } from 'bar'`?

lib/rules/prefer-user-event.ts
```diff
@@ -90,59 +90,38 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
-          return;
-        }
-        const fireEventImport = node.specifiers.find(
-          (node) =>
-            isImportSpecifier(node) && node.imported.name === 'fireEvent'
-        );
-        hasNamedImportedFireEvent = !!fireEventImport;
-        if (!hasNamedImportedFireEvent) {
+      ['CallExpression > MemberExpression'](node: TSESTree.MemberExpression) {
+        if (!helpers.getIsTestingLibraryImported()) {
```
Belco90, line 99: You don't have to do this anymore. Detection module is taking care of this and prevent the rules listeners you declare here from being executed if not testing library imported or filename not valid. You can find that at the very bottom of `detectTestingLibraryUtils`:

```javascript
if (helpers.canReportErrors() && ruleInstructions[instruction]) {
  return ruleInstructions[instruction](node);
}
```

lib/node-utils.ts
Belco90: I think it's a good idea to have this within detection module. I actually [mentioned this helper](https://github.com/testing-library/eslint-plugin-testing-library/issues/198#issuecomment-669927224) when analyzing how to organize the helpers for detection module. There are difference scenarios as you have already figured in your draft, but there are more related to other Testing Library utils, specially for `render`.

Anyway, I think we should include this one within detection already, covering all different imports. Then we can improve it when necessary when refactoring other rules. I'd name it something like `findTestingLibraryUtilNode`

lib/rules/prefer-user-event.ts
```diff
@@ -90,59 +90,38 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
+          isIdentifier(node.object) &&
+          node.object.name === fireEventAliasOrWildcard;
+
         const fireEventFromWildcardUsed =
           isMemberExpression(node.object) &&
           isIdentifier(node.object.object) &&
-          node.object.object.name === wildcardImportName &&
+          node.object.object.name === fireEventAliasOrWildcard &&
           isIdentifier(node.object.property) &&
           node.object.property.name === 'fireEvent';
```
Belco90, line 121: Shouldn't this one be `fireEventAliasOrWildcard` too?

lib/rules/prefer-user-event.ts
gndelia: yeah I remembered about it when I went to sleep, but the laptop was already turned off 😆  but thanks for commenting it 👍

lib/node-utils.ts
gndelia: it is covered (the tests cover this scenario), but the comment is not reflecting it - I will update it

lib/rules/prefer-user-event.ts
gndelia: no, this scenario validates is for the case when rtl was imported with `import * as rtl from 'module'`.
Because of that, `fireEvent` is never aliased, as you can do `rtl.fireEvent.change()` or whatever alias you placed to the whole module.

I'm not covering

```ts
import * as rtl from 'module'
const aliasedFireEvent = rtl.fireEvent
aliasedFireEvent.change()
```

because that's too much 😅 

So `node.object.object.name === fireEventAliasOrWildcard` checks for the `rtl` or whatever alias you've used (the tests use `dom` for instance), and the line from this comment should always be `fireEvent`

Belco90: Ah of course! Thanks for the clarification.

lib/node-utils.ts
gndelia: done

lib/rules/prefer-user-event.ts
gndelia: removed!

lib/node-utils.ts
gndelia: I moved it into the detection helper, and added the scenario for `require`

gndelia:
I added a second commit, if CI passes I think this one is ready (unless I'm missing something or feedback is required).

- I added the function in the detection helpers to extract a "specifier" from the `import` or `require`
- I improved the docs of this PR's rule with examples supporting `require` (they weren't there before)
- I added tests for `require` and supporting a custom import for all previous scenarios

tests/lib/rules/prefer-user-event.test.ts
```diff
@@ -130,5 +191,53 @@ ruleTester.run(RULE_NAME, rule, {
         errors: [{ messageId: 'preferUserEvent' }],
       })
     ),
+    ...createScenarioWithImport<InvalidTestCase<MessageIds, Options>>(
+      (libraryModule: string, fireEventMethod: string) => ({
+        code: `
+        const { fireEvent } = require('${libraryModule}')
+        fireEvent.${fireEventMethod}(foo)
+      `,
+        errors: [{ messageId: 'preferUserEvent' }],
```
Belco90, line 200: It would be nice to check line and column of the error.

Belco90 (changes requested):
Looking really good! We just need to increase the coverage up to 100% and I'd suggest to check column and line of errors as mentioned.

tests/lib/rules/prefer-user-event.test.ts
gndelia: added in https://github.com/testing-library/eslint-plugin-testing-library/pull/251/commits/06cb9cf45c0a78f963461dcaeb3230840ed657e0

Belco90 (approved):
Looking good to me! I think we should put some specific tests for `findImportedUtilSpecifier` in the fake rule, but it's fine for now.

PR #252, opened 2020-11-05
Move rules settings to ESLint shared config: refactor prefer-presence-queries rule

Belco90:
Relates to #198 

This refactor for `prefer-presence-queries` includes:
- improve existing tests to check column where the error is reported
- add new detection helpers to determine if a node corresponds to `getBy*`, `queryBy*` or sync queries
- extract detection helpers to determine if a node corresponds to presence/absence assertion

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,6 +1,6 @@
 {
   "*.{js,ts}": [
-    "eslint --fix",
+    "eslint --max-warnings 0 --fix",
     "prettier --write",
     "jest --findRelatedTests"
   ],
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -1,4 +1,4 @@
-import { TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
+import { ASTUtils, TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
 import {
   getImportModuleName,
   isLiteral,
@@ -6,7 +6,6 @@ import {
   isImportDeclaration,
   isImportNamespaceSpecifier,
   isImportSpecifier,
-  isIdentifier,
   isProperty,
 } from './node-utils';
 
@@ -41,6 +40,9 @@ export type DetectionHelpers = {
   getCustomModuleImportName: () => string | undefined;
   getIsTestingLibraryImported: () => boolean;
   getIsValidFilename: () => boolean;
+  isGetByQuery: (node: TSESTree.Identifier) => boolean;
+  isQueryByQuery: (node: TSESTree.Identifier) => boolean;
+  isSyncQuery: (node: TSESTree.Identifier) => boolean;
   canReportErrors: () => boolean;
   findImportedUtilSpecifier: (
     specifierName: string
@@ -85,7 +87,8 @@ export function detectTestingLibraryUtils<
         return getImportModuleName(importedCustomModuleNode);
       },
       /**
-       * Gets if Testing Library is considered as imported or not.
+       * Determines whether Testing Library utils are imported or not for
+       * current file being analyzed.
        *
        * By default, it is ALWAYS considered as imported. This is what we call
        * "aggressive reporting" so we don't miss TL utils reexported from
@@ -105,9 +108,8 @@ export function detectTestingLibraryUtils<
       },
 
       /**
-       * Gets if filename being analyzed is valid or not.
-       *
-       * This is based on "testing-library/filename-pattern" setting.
+       * Determines whether filename is valid or not for current file
+       * being analyzed based on "testing-library/filename-pattern" setting.
        */
       getIsValidFilename() {
         const fileName = context.getFilename();
@@ -115,7 +117,28 @@ export function detectTestingLibraryUtils<
       },
 
       /**
-       * Wraps all conditions that must be met to report rules.
+       * Determines whether a given node is `getBy*` or `getAllBy*` query variant or not.
+       */
+      isGetByQuery(node) {
+        return !!node.name.match(/^get(All)?By.+$/);
+      },
+
+      /**
+       * Determines whether a given node is `queryBy*` or `queryAllBy*` query variant or not.
+       */
+      isQueryByQuery(node) {
+        return !!node.name.match(/^query(All)?By.+$/);
+      },
+
+      /**
+       * Determines whether a given node is sync query.
+       */
+      isSyncQuery(node) {
+        return this.isGetByQuery(node) || this.isQueryByQuery(node);
+      },
+
+      /**
+       * Determines if file inspected meets all conditions to be reported by rules or not.
        */
       canReportErrors() {
         return (
@@ -144,7 +167,7 @@ export function detectTestingLibraryUtils<
           return node.specifiers.find((n) => isImportNamespaceSpecifier(n));
         } else {
           const requireNode = node.parent as TSESTree.VariableDeclarator;
-          if (isIdentifier(requireNode.id)) {
+          if (ASTUtils.isIdentifier(requireNode.id)) {
             // this is const rtl = require('foo')
             return requireNode.id;
           }
@@ -153,7 +176,7 @@ export function detectTestingLibraryUtils<
           const property = destructuring.properties.find(
             (n) =>
               isProperty(n) &&
-              isIdentifier(n.key) &&
+              ASTUtils.isIdentifier(n.key) &&
               n.key.name === specifierName
           );
           return (property as TSESTree.Property).key as TSESTree.Identifier;
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -1,6 +1,7 @@
 import { ASTUtils, TSESLint, TSESTree } from '@typescript-eslint/experimental-utils';
 import {
   getImportModuleName,
+  getAssertNodeInfo,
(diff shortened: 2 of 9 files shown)


.lintstagedrc
```diff
@@ -1,6 +1,6 @@
 {
   "*.{js,ts}": [
-    "eslint --fix",
+    "eslint --max-warnings 0 --fix",
```
Belco90, line 3: We are setting max-warnings to 0 in the CI so I think it's better to have same conditions in precommit hook.

lib/detect-testing-library-utils.ts
```diff
@@ -32,6 +38,11 @@ export type DetectionHelpers = {
   getCustomModuleImportName: () => string | undefined;
   getIsTestingLibraryImported: () => boolean;
   getIsValidFilename: () => boolean;
+  isGetByQuery: (node: TSESTree.Identifier) => boolean;
+  isQueryByQuery: (node: TSESTree.Identifier) => boolean;
+  isSyncQuery: (node: TSESTree.Identifier) => boolean;
+  isPresenceAssert: (node: TSESTree.MemberExpression) => boolean;
+  isAbsenceAssert: (node: TSESTree.MemberExpression) => boolean;
```
Belco90, line 45: I named these functions with prefix `is`. I prefer to use prefix `is`, `has`, `can` for vars, and `getIs`, `getHas` , `getCan` for methods returning booleans. However, `@typescript-eslint/experimental-utils` is using the former for all its utils so I think to follow same standard ours should be named in the same way.

This implies I'll have to rename:
- `getIsTestingLibraryImported` to `isTestingLibraryImported`
- `getIsValidFilename` to `isValidFilename`

I'll do this in a separate PR tho.

lib/detect-testing-library-utils.ts
```diff
@@ -93,17 +105,75 @@ export function detectTestingLibraryUtils<
...
         const fileName = context.getFilename();
         return !!fileName.match(filenamePattern);
       },
 
       /**
-       * Wraps all conditions that must be met to report rules.
+       * Determines whether a given node is `getBy*` or `getAllBy*` query variant or not.
+       */
+      isGetByQuery(node) {
+        return !!node.name.match(/^get(All)?By.+$/);
```
Belco90, line 120: Aggressive query reporting enabled! It doesn't matter if it's a Testing Library built-in query or a custom one: if it matches `getBy*` query pattern, it's considered like so. Same for `queryBy*` queries, and `findBy*` in the future.

lib/node-utils.ts
```diff
@@ -253,3 +254,42 @@ export function getImportModuleName(
...
 }
+
+type AssertNodeInfo = {
+  matcher: string | null;
+  isNegated: boolean;
+};
+/**
+ * Extracts matcher info from MemberExpression node representing an assert.
+ */
+export function getAssertNodeInfo(
```
Belco90, line 265: This was extracted from the rule itself to get info about the assert node. Could be useful in other rules if assert statement impacts the rule reporting.

lib/rules/prefer-presence-queries.ts
```diff
@@ -33,62 +16,47 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
       recommended: 'error',
     },
     messages: {
-      presenceQuery:
+      wrongPresenceQuery:
         'Use `getBy*` queries rather than `queryBy*` for checking element is present',
-      absenceQuery:
+      wrongAbsenceQuery:
         'Use `queryBy*` queries rather than `getBy*` for checking element is NOT present',
-      expectQueryBy:
```
Belco90, line 40: This one wasn't used.

lib/utils.ts
```diff
@@ -24,6 +24,7 @@ const LIBRARY_MODULES = [
   '@testing-library/svelte',
 ];
 
+// TODO: should be deleted after all rules are migrated to v4
 const hasTestingLibraryImportModule = (
```
Belco90, line 28: This method **must** disappear after refactoring all rules since this check it's part of detection module now.

tests/create-testing-library-rule.test.ts
```diff
@@ -89,6 +89,102 @@ ruleTester.run(RULE_NAME, rule, {
       import { foo } from 'custom-module-forced-report'
     `,
     },
+
+    // Test Cases for all settings mixed
```
Belco90, line 93: Adding tons of test cases here for new detection helpers. Even if they are covered by rules themselves, I find this really useful to:

1. document its intended usage
2. have specific test for each detection helper in case the real rules are modified
3. implement detection helpers isolated first and then include them within real rules
4. check all necessary conditions are met to report errors (i.e. `canReportErrors`) so each rule doesn't have to repeat these tests (although I've repeated some in few rules, just to give some context within rule tests themselves)

tests/lib/rules/prefer-presence-queries.test.ts
```diff
@@ -14,81 +14,329 @@ const queryAllByQueries = ALL_QUERIES_METHODS.map(
...
+const getInvalidAssertion = ({
+  query,
+  matcher,
+  messageId,
+  shouldUseScreen = false,
+}: AssertionFnParams) => {
+  const finalQuery = shouldUseScreen ? `screen.${query}` : query;
+  return {
+    code: `expect(${finalQuery}('Hello'))${matcher}`,
+    errors: [{ messageId, line: 1, column: shouldUseScreen ? 15 : 8 }],
```
Belco90, line 44: Tests in this file are mostly the same, but I refactor the assertion makers to be able to check the column within the error.

tests/lib/rules/prefer-presence-queries.test.ts
```diff
@@ -98,96 +346,343 @@ ruleTester.run(RULE_NAME, rule, {
...
       code: 'const el = queryByText("button")',
     },
-    {
-      code:
-        'expect(getByNonTestingLibraryQuery("button")).not.toBeInTheDocument()',
-    },
-    {
-      code:
-        'expect(queryByNonTestingLibraryQuery("button")).toBeInTheDocument()',
-    },
```
Belco90, line 108: This are invalid now since aggressive query reporting has been added!

tests/lib/rules/prefer-presence-queries.test.ts
```diff
@@ -98,96 +346,343 @@ ruleTester.run(RULE_NAME, rule, {
...
         expect(el).toBeInTheDocument()
       }`,
     },
-    // some weird examples after here to check guard against parent nodes
-    {
-      code: 'expect(getByText("button")).not()',
-    },
-    {
-      code: 'expect(queryByText("button")).not()',
-    },
```
Belco90, line 121: These two have been moved to fake rule tests.

lib/detect-testing-library-utils.ts
```diff
@@ -93,17 +105,75 @@ export function detectTestingLibraryUtils<
...
+       */
+      isQueryByQuery(node) {
+        return !!node.name.match(/^query(All)?By.+$/);
+      },
+
+      /**
+       * Determines whether a given node is sync query or not.
+       */
+      isSyncQuery(node) {
+        return this.isGetByQuery(node) || this.isQueryByQuery(node);
```
gndelia, line 134: perhaps we should use `helpers` instead of `this` to avoid losing the type inference (as discussed in discord)

Belco90: Indeed. I'll address that in a separate PR when open PRs are merged.

timdeschryver: Good catch 👍

lib/node-utils.ts
```diff
@@ -253,3 +254,42 @@ export function getImportModuleName(
...
+    return emptyInfo;
+  }
+
+  if (node.object.callee.name !== 'expect') {
+    return emptyInfo;
+  }
+
+  let matcher = ASTUtils.getPropertyName(node);
+  let isNegated = false;
+  if (matcher === 'not') {
```
timdeschryver, line 283: Nit, I do find this more readable because it isn't reassigned


```suggestion
  let isNegated = matcher === 'not';
  if (isNegated ) {
```

lib/utils.ts
timdeschryver: Should we mark it as `@deprecated`, because you'll get a warning if you want to use it?
Could be helpful for contributors.

lib/node-utils.ts
Belco90: That's better, I'm changing it!

Belco90:
@gndelia @timdeschryver Anything else I'm missing here? 😅

lib/detect-testing-library-utils.ts
```diff
@@ -105,17 +116,75 @@ export function detectTestingLibraryUtils<
...
+        const { matcher, isNegated } = getAssertNodeInfo(node);
+
+        if (!matcher) {
+          return false;
+        }
+
+        return isNegated
+          ? PRESENCE_MATCHERS.includes(matcher)
+          : ABSENCE_MATCHERS.includes(matcher);
+      },
```
gndelia, line 184: might be a stupid question, but should we just do `!isPresenceAssert(node)` ?

gndelia (approved):
LGTM

lib/utils.ts
Belco90: It will disappear when v4 is finished so I don't think we even need to deprecate it

PR #254, opened 2020-11-10
refactor: detection helpers tweaks

Belco90:
Relates to #198 

This PR includes two refactors related to detection helpers:
- declare detection helpers as standalone functions rather than object properties (so we avoid missing types if called with `this`)
- rename boolean functions to follow proper convention for boolean statements (prefixed by `is`, `has`, `should` or `can`)

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -44,8 +44,8 @@ export type DetectionHelpers = {
   getCustomModuleImportNode: () => ImportModuleNode | null;
   getTestingLibraryImportName: () => string | undefined;
   getCustomModuleImportName: () => string | undefined;
-  getIsTestingLibraryImported: () => boolean;
-  getIsValidFilename: () => boolean;
+  isTestingLibraryImported: () => boolean;
+  isValidFilename: () => boolean;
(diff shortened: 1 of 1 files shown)


lib/detect-testing-library-utils.ts
```diff
@@ -81,153 +81,169 @@ export function detectTestingLibraryUtils<
...
-       * However, there is a setting to customize the module where TL utils can
-       * be imported from: "testing-library/module". If this setting is enabled,
-       * then this method will return `true` ONLY IF a testing-library package
-       * or custom module are imported.
-       */
-      getIsTestingLibraryImported() {
-        if (!customModule) {
-          return true;
-        }
+    const getTestingLibraryImportNode: DetectionHelpers['getTestingLibraryImportNode'] = () => {
```
Belco90, line 84: I'm defining `DetectionHelpers` outside so then I can both pass it to `EnhancedRuleCreate` and reuse it for functions definitions. Ideally I'd like to do the opposite: type each function definition, then extract the type of final `helper` object and pass that to `EnhancedRuleCreate` but I don't know if there is a way of doing that since the type should be extracted to an outer scope somehow. This works tho.

gndelia (approved):

PR #255, opened 2020-11-14
refactor: prefer-wait-for with the new settings

gndelia:
#198 migrating prefer-wait-for

I had to reimplement the rule because the implementation was tightly coupled with the imports

- I added support for `require`
- I improved the test coverage, I took all the existing scenarios and I added testing with `require`, with `settings` and all known imports for testing-library
- I improved the docs with the `require` support

### DIFF
--- a/lib/rules/prefer-wait-for.ts
+++ b/lib/rules/prefer-wait-for.ts
@@ -6,11 +6,11 @@ import {
   isIdentifier,
   findClosestCallExpressionNode,
   isCallExpression,
-  isImportDeclaration,
   isImportNamespaceSpecifier,
-  isVariableDeclarator,
   isObjectPattern,
   isProperty,
+  isMemberFromMethodCallFromTestingLibrary,
+  isIdentifierInCallExpressionFromTestingLibrary,
 } from '../node-utils';
 
 export const RULE_NAME = 'prefer-wait-for';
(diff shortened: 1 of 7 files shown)


lib/rules/prefer-wait-for.ts
```diff
@@ -112,46 +150,100 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
+          isVariableDeclarator(testingLibraryNode.parent) &&
+          isIdentifier(testingLibraryNode.parent.id) &&
+          node.object.name === testingLibraryNode.parent.id.name;
+        if (
+          !callerIsTestingLibraryFromImport &&
+          !callerIsTestingLibraryFromRequire
+        ) {
+          // the method does not match from the imported elements from TL (even from custom)
+          return;
+        }
```
Belco90, line 183: Should we extract this into its own detection helper? A function which determines if a given node was directly imported from testing library (either default or custom module). It looks like we are gonna end up repeating this bit across different rules from now on.

lib/rules/prefer-wait-for.ts
```diff
@@ -112,46 +150,100 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
+              isIdentifier(p.key) &&
+              isIdentifier(p.value) &&
+              p.value.name === node.name &&
+              DEPRECATED_METHODS.includes(p.key.name)
+          );
+        if (
+          !callerIsTestingLibraryFromRequire &&
+          !callerIsTestingLibraryFromImport
+        ) {
+          return;
```
Belco90, line 217: Very similar to the previous one. I think this and the previous code should be merged within a detection helper.

Belco90 (commented):
Amazing work! You covered so many cases with tests, that's really nice! I think you could have skip some tho, as the main detection helpers are already tested within fake rule (custom module name, TL not imported, different file name, etc) but anyway it's nice to have all that here.

I think we can extract some of your code as a generic way to check if a node was imported from testing library (either default or custom module). What do you think about it?

gndelia:
@Belco90  generalized both method (I think) in this commit https://github.com/testing-library/eslint-plugin-testing-library/pull/255/commits/97d89d28f3efc5c8611318eedf28e3c13ef0ad01

let's see how useful are for future rules 😅

Belco90:
@gnapse sorry, I probably wasn't to specific about it. This util should be placed within detection wrapper and passed through `helpers`. See how you have to pass `helpers` itself to those methods because they actually should belong to the detection mechanism itself, since they need to check testing library imports to determine whether the node is considered as imported or not.

My idea then was: having a single `isNodeImportedFromTestingLibrary` or something similar, where you can pass a node of type `Identifier` or `MemberExpression`. Then that method will take care of checking if imported/required with `isMemberFromMethodCallFromTestingLibrary` or `isIdentifierInCallExpressionFromTestingLibrary` (these two are not directly exposed, they just live inside detection mechanism, or even detection file). Finally, those 2 extra utils `isMemberFromCallExpressionComingFromImport` and `isMemberFromCallExpressionComingFromRequire` are fine within `node-utils` file since are not tied to detection helpers themselves.

Does that make sense?

gndelia:
you tagged the wrong person @Belco90  🤣 

I can review it yeah. I also had doubts on where to place it, my fear was that by the implementation it looked closer to node_utiles, but

Belco90:
> you tagged the wrong person @Belco90 🤣
> 
> I can review it yeah. I also had doubts on where to place it, my fear was that by the implementation it looked closer to node_utiles, but

Damn, sorry @gnapse I mentioned you by mistake (thanks GitHub mentions autocomplete).

gndelia:
done @Belco90  - check it out!

Belco90 (approved):
Awesome! Looks good now, sorry to bother you with the last change but the simple we leave the API of detection helpers the better. Now you need 100% coverage on detect utils module, you should be able to achieve that by adding some stupid test to fake rule.

Belco90 (approved):

PR #256, opened 2020-11-17
refactor(ast-utils): migrate custom node-utils to ASTUtils

### LINKED ISSUE #253: Drop some custom type guards in favor of ASTUtils
As seen in some PRs created by @Belco90 , there's a utility `ASTUtils` in ` '@typescript-eslint/experimental-utils'` which can replace some, but not all, of the type guards we have in our [node-utils.ts file](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/lib/node-utils.ts)

This ticket is to review which type guards (`isIdentifier`, `isVariableDeclarator`, etc) are repeated, and drop those in favour of the ones in `ASTUtils`. The ones that are not available there should continue to exist with our custom implementation.

Edit: This PR should target the v4 branch

thomlom:
Closes #253 

- [x] Migrate `isIdentifier` 
- [x] Migrate `isAwaitExpression`

It seems there are no [further utils](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/experimental-utils/src/ast-utils/predicates.ts) than the two above. I guess they will add it over time, so we should keep an eye on it 🙂

### DIFF
--- a/lib/node-utils.ts
+++ b/lib/node-utils.ts
@@ -15,13 +15,13 @@ export function isCallExpression(
 export function isNewExpression(
   node: TSESTree.Node
 ): node is TSESTree.NewExpression {
-  return node && node.type === 'NewExpression';
+  return node?.type === 'NewExpression';
 }
 
 export function isMemberExpression(
   node: TSESTree.Node
 ): node is TSESTree.MemberExpression {
-  return node && node.type === AST_NODE_TYPES.MemberExpression;
+  return node?.type === AST_NODE_TYPES.MemberExpression;
 }
 
 export function isLiteral(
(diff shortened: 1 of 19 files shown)


Belco90 (approved):
Hey Thomas, nice to have you back! Thanks for this refactor. I think in the future we should export our own `ASTUtils` extending the `experimental-utils` one with our utils, but this is enough for now.

Belco90:
Oops, code coverage for node-utils file is under 90% bar. Could you modify it to be 85% as we are gonna increase that file coverage in future PRs please? After that, we can merge this.

Belco90:
Actually I can do that myself directly on v4, don't worry!

PR #257, opened 2020-11-17
fix(prefer-explicit-assert): handle getBy* without expect

skovy:
This PR fixes an issue with the `prefer-explicit-assert` `assertion` configuration option. If `assertion` is defined, it currently assumes all `getBy*` queries are wrapped in an `expect`. However, even with the `perfer-explicit-assert` there are valid uses that aren't wrapped in an `expect`:

```js
fireEvent.click(getByText('bar'));
const el = getByText('foo');
```

### DIFF
--- a/lib/rules/prefer-explicit-assert.ts
+++ b/lib/rules/prefer-explicit-assert.ts
@@ -95,6 +95,7 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
             });
           } else if (assertion) {
             const expectCallNode = findClosestCallNode(node, 'expect');
+            if (!expectCallNode) return;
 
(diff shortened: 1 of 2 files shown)


Belco90 (approved):
Hi again! You are becoming an expert for this rule 😅

skovy:
Haha yea, working through the issues that are surfaced while running in a real codebase 😅 

I appreciate your feedback and reviews, thanks!

julienw:
I'm a bit surprised: we definitely had such legitimate uses in our project, but this wasn't flagged by the rule even without this patch. I'm puzzled :-)

Belco90:
@julienw You edited before I answered properly! Indeed, this only applies to `assertion` option which is not enabled by default.

PR #260, opened 2020-11-24
Move rules settings to ESLint shared config: refactor await-async-query and no-await-sync-query

Belco90:
Relates to #198 

This refactor for `await-async-query` and `no-await-sync-query` rules includes:
- using custom rule creator + detection helpers for both rules
- enabling aggressive query reporting
- covering new case for `await-async-query` to report wrapper functions returning promise from async query
- detection helpers for findBy and async queries
- making checks more consistent across unhandled promises
- adding several new utils for functions and async nodes


docs/rules/await-async-query.md
```diff
@@ -1,4 +1,4 @@
-# Enforce async queries to have proper `await` (await-async-query)
+# Enforce promise from async queries to be handled (await-async-query)
```
Belco90, line 1: Just updating the description and examples to be more accurate. We probably need to do the same on the others `await-async-*` rules.

Belco90 (commented):
Oops, forgot to submit my comments

docs/rules/await-async-query.md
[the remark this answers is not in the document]
gndelia: extremely minor comment, I think it should be plural

```suggestion
# Enforce promises from async queries to be handled (await-async-query)
```

because `queries` is plural

lib/node-utils.ts
```diff
@@ -180,6 +266,86 @@ export function getVariableReferences(
...
+  functionNode:
+    | TSESTree.FunctionDeclaration
+    | TSESTree.FunctionExpression
+    | TSESTree.ArrowFunctionExpression
+): TSESTree.Node | null {
+  if (isBlockStatement(functionNode.body)) {
+    // regular function or arrow function with block
+    const returnStatementNode = functionNode.body.body.find((statement) =>
+      isReturnStatement(statement)
+    ) as TSESTree.ReturnStatement | null;
```
gndelia, line 305: if not found, `find` will return `undefined` instead of `null`

```suggestion
    const returnStatementNode = functionNode.body.body.find((statement) =>
      isReturnStatement(statement)
    ) as TSESTree.ReturnStatement | undefined;
```

gndelia (approved):
I left a minor suggestion but looks great 🚀

lib/node-utils.ts
Belco90: Nice one. Done.

docs/rules/await-async-query.md
```diff
@@ -1,75 +1,86 @@
...
+This rule aims to prevent users from forgetting to handle the returned
 promise from those async queries to be fulfilled, which could lead to
-errors in the tests. The promises can be handled by using either `await`
-operator or `then` method.
+errors in the tests. The promise will be considered as handled when:
+
+- using `await` operator
+- chaining `then` method
+- chaining `resolves` or `rejects` from jest
+- is returned from a function (in this case, that particular function will be analyzed by this rule too)
```
timdeschryver, line 21:
```suggestion
- using the `await` operator
- chaining the `then` method
- chaining `resolves` or `rejects` from jest
- it's returned from a function (in this case, that particular function will be analyzed by this rule too)
```

lib/detect-testing-library-utils.ts
```diff
@@ -145,13 +147,27 @@ export function detectTestingLibraryUtils<
       return !!node.name.match(/^query(All)?By.+$/);
     };
 
+    /**
+     * Determines whether a given node is `findBy*` or `findAllBy*` query variant or not.
+     */
+    const isFindByQuery: DetectionHelpers['isFindByQuery'] = (node) => {
+      return !!node.name.match(/^find(All)?By.+$/);
```
timdeschryver, line 154: This is a tiny bit faster 😅

```suggestion
      return /^find(All)?By.+$/.test(node.name);
```

lib/node-utils.ts
```diff
@@ -169,6 +212,49 @@ export function isPromiseResolved(node: TSESTree.Node): boolean {
...
 }
 
+/**
+ * Determines whether an Identifier related to a promise is considered as handled.
+ *
+ * It will be considered as handled if:
+ * - it belongs to `await` expression
+ * - it's chained with `then` method
+ * - it's returned from a function
+ * - has `resolves` or `rejects`
```
timdeschryver, line 222:
```suggestion
 * It will be considered as handled if:
 * - it belongs to the `await` expression
 * - it's chained with the `then` method
 * - it's returned from a function
 * - has `resolves` or `rejects`
```

lib/node-utils.ts
```diff
@@ -180,6 +266,86 @@ export function getVariableReferences(
...
+
+    if (!returnStatementNode) {
+      return null;
+    }
+    return returnStatementNode.argument;
+  } else if (functionNode.expression) {
+    // arrow function with implicit return
+    return functionNode.body;
+  }
+}
```
timdeschryver, line 315: Weird that TS doesn't complain about this (because the return value of this function is `TSESTree.Node | null`).

```suggestion
  return null;
}
```

timdeschryver (approved):
LGTM, I did leave a few nits.

PR #265, opened 2020-12-06
Move rules settings to ESLint shared config: refactor await-fire-event

Belco90:
Relates to #198 

This refactor for `await-fire-event` includes:

- extracting helper for fireEvent detection from `prefer-user-event`
- covering new case to report wrapper functions returning promise from fireEvent methods (same as `await-async-query` and `await-async-utils`)
- covering new case to report consider promises from fireEvent methods as handled if wrapped within `Promise.all` or `Promise.allSettled` (same as `await-async-query` and `await-async-utils`)


Belco90:
@afontcu I added you as a reviewer just in case you could double check the tests and docs examples to confirm they are fine. This rule should behave in the same way as others `await-*` rules but since only applies to Vue I prefer to make sure it works as expected.

lib/detect-testing-library-utils.ts
```diff
@@ -88,6 +90,15 @@ export function detectTestingLibraryUtils<
...
       DEFAULT_FILENAME_PATTERN;
 
+    /**
+     * Determines whether aggressive reporting is enabled or not.
+     *
+     * Aggressive reporting is considered as enabled when:
+     * - custom module is not set (so we need to assume everything
+     *    matching TL utils is related to TL no matter where it was imported from)
+     */
+    const isAggressiveReportingEnabled = () => !customModule;
```
Belco90, line 100: I extracted this into this simple function for adding some syntactic sugar around "aggressive reporting" concept. I'll create a separate PR soon where I'll use this in `isNodeComingFromTestingLibrary`.

lib/detect-testing-library-utils.ts
```diff
@@ -176,6 +187,58 @@ export function detectTestingLibraryUtils<
       return ASYNC_UTILS.includes(node.name);
     };
 
+    /**
+     * Determines whether a given node is fireEvent method or not
+     */
+    const isFireEventMethod: DetectionHelpers['isFireEventMethod'] = (node) => {
```
Belco90, line 193: Extracted from `prefer-user-event`! I tweaked it a little bit to improve some edge cases.

lib/detect-testing-library-utils.ts
```diff
@@ -260,52 +324,24 @@ export function detectTestingLibraryUtils<
     };
     /**
      * Takes a MemberExpression or an Identifier and verifies if its name comes from the import in TL
-     * @param node a MemberExpression (in "foo.property" it would be property) or an Identifier (it should be provided from a CallExpression, for example "foo()")
+     * @param node a MemberExpression (in "foo.property" it would be property) or an Identifier
      */
     const isNodeComingFromTestingLibrary: DetectionHelpers['isNodeComingFromTestingLibrary'] = (
```
Belco90, line 329: I realized this one was doing basically the same as `findImportedUtilSpecifier` so I refactored it.

lib/node-utils.ts
```diff
@@ -27,7 +27,6 @@ const ValidLeftHandSideExpressions = [
   AST_NODE_TYPES.ObjectExpression,
   AST_NODE_TYPES.ObjectPattern,
   AST_NODE_TYPES.Super,
-  AST_NODE_TYPES.TemplateLiteral,
```
Belco90, line 30: Duplicated in the list

lib/node-utils.ts
```diff
@@ -132,7 +131,7 @@ export function findClosestCallExpressionNode(
     return null;
   }
 
-  return findClosestCallExpressionNode(node.parent);
+  return findClosestCallExpressionNode(node.parent, shouldRestrictInnerScope);
```
Belco90, line 134: Bug difficult to spot here! Took me a while to find it.

lib/rules/prefer-user-event.ts
```diff
@@ -88,50 +88,34 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
     const { allowedMethods } = options;
-    const sourceCode = context.getSourceCode();
 
     return {
-      ['CallExpression > MemberExpression'](node: TSESTree.MemberExpression) {
-        const util = helpers.findImportedUtilSpecifier('fireEvent');
-        if (!util) {
-          // testing library was imported, but fireEvent was not imported
+      'CallExpression Identifier'(node: TSESTree.Identifier) {
+        if (!helpers.isFireEventMethod(node)) {
```
Belco90, line 94: I extracted logic from this rule to `isFireEventMethod` so the rule is really simple now!

docs/rules/await-fire-event.md
```diff
@@ -1,12 +1,14 @@
...
 properly.
 
 ## Rule Details
 
-This rule aims to prevent users from forgetting to await `fireEvent`
-methods when they are async.
+This rule aims to prevent users from forgetting to handle promise returned from `fireEvent`
+methods.
+
+> ⚠️ `fireEvent` methods are async only on `@testing-library/vue` package
```
afontcu, line 11: Looks like svelte is also wrapping fireEvent with a promise: https://github.com/testing-library/svelte-testing-library/blob/master/src/pure.js#L94-L106

Belco90: I see, I didn't know it. I'll update this note to clarify fire event methods are async on Vue and Svelte packages, but only supported official by the plugin for the former.

lib/rules/await-fire-event.ts
```diff
@@ -1,51 +1,106 @@
...
+          if (!closestCallExpression) {
+            return;
+          }
+
+          const references = getVariableReferences(
+            context,
+            closestCallExpression.parent
+          );
+
+          if (!references || references.length === 0) {
```
timdeschryver, line 79: Nit, I believe `getVariableReferences` will return an empty array when there are no references

```suggestion
          if (references.length === 0) {
```

timdeschryver (approved):
To be honest, I got a bit lost here 😅 but that's on me

LGTM

Belco90:
@timdeschryver I can avoid adding you as reviewer in the next PRs if you want me to. I'd appreciate if you could review the next one I'm gonna create since you implemented the rule.

PR #270, opened 2020-12-11
refactor: migrate prefer-find-by to v4

gndelia:
This PR migrates the `prefer-find-by` rule to use the new helper, as  part of #198

Fortunately, the rule was not even checking the imports, so only a few changes were required.

I did update the tests a bit so they are more "realistic" than just code floating around.

### DIFF
--- a/tests/lib/rules/prefer-find-by.test.ts
+++ b/tests/lib/rules/prefer-find-by.test.ts
@@ -279,5 +279,59 @@ ruleTester.run(RULE_NAME, rule, {
         const submitButton = await findByRole('baz', { name: 'button' })
       `,
     },
+    // custom query triggers the error but there is no fix - so output is the same
+    ...WAIT_METHODS.map((waitMethod: string) => ({
+      code: `
+        import {${waitMethod},render} from '@testing-library/foo';
+        it('tests', async () => {
+          const { getByCustomQuery } = render()
+          const submitButton = await ${waitMethod}(() => getByCustomQuery('baz'))
+        })
+      `,
+      errors: [
+        {
+          messageId: 'preferFindBy',
+          data: {
+            queryVariant: 'findBy',
+            queryMethod: 'CustomQuery',
+            fullQuery: `${waitMethod}(() => getByCustomQuery('baz'))`,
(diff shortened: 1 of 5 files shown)


Belco90 (changes requested):
Sorry, but there things related to aggressive query reporting to refactor here to use detection utils instead of manual rule checks:

The rule is still using `SYNC_QUERIES_COMBINATIONS.includes` when this must be replaced by `helpers.isSyncQuery`. Otherwise, aggressive query reporting is not enabled for this rule (you can check this by adding an extra test using a custom query like `getByIcon`). This happens several several times within this rule, so you need to replace all its usages.

The idea is not just create the library with `createTestingLibraryRule` but also use all detection helpers which already cover rule manual checks + enable aggressive reporting stuff.

gndelia:
@Belco90  I've created the `isCustomQuery` helper method. I updated the tests for the fake rule to take it into consideration. As discussed in discord, I am now using `helpers.isSyncQuery` and not doing a fix if the query is a custom one.

tests/lib/rules/prefer-find-by.test.ts
```diff
@@ -216,5 +279,59 @@ ruleTester.run(RULE_NAME, rule, {
...
+          const submitButton = await ${waitMethod}(() => screen.getByCustomQuery('baz'))
+        })
+      `,
+      errors: [
+        {
+          messageId: 'preferFindBy',
+          data: {
+            queryVariant: 'findBy',
+            queryMethod: 'CustomQuery',
+            fullQuery: `${waitMethod}(() => screen.getByCustomQuery('baz'))`,
```
gndelia, line 324: my only doubt: are the custom queries available in `screen` ? I don't use them honestly 🤣  (on second thoughts, I can now think a few usages where I could use this at work 🤔 )

Belco90: I asked about this on Testing Library discord but got no answer. I think it is technically possible, since you would only need to define your custom queries + reexport `screen` extended with custom ones.

Belco90 (approved):

Issue #269, opened 2020-12-09
Disallow the use of `queryBy*` queries for assertions


thomaslombart:
I've seen people using `queryBy` as assertions in their tests:

```js
screen.queryByText("Account")
```

This is not the right way to use it because the test won't fail whether the text is there or not. 

Would it be possible to extend the `prefer-explicit-assert` rule in order to disallow the use of `queryBy` as assertions as well?

Belco90:
I think it would make sense to be part of `prefer-explicit-assert`.

Belco90:
This should be really easy to implement on v4, but it hasn't been refactored yet.

thomlom:
@Belco90 I'll refactor the rule and add the enhancement then 🙌

Belco90:
That would be awesome. I'm gonna put that rule on WIP assigned to you on #198 then.

Please double check that issue as that rule has a side note to remove its option, which is not necessary anymore. You can check `prefer-presence-queries` or `await-async-query` to get an idea about how we are reporting the queries now. You'll see that:
- we have detection helpers to know if an identifier is a particular Testing Library query
- we don't check anymore if coming from `render` method or Testing Library itself (aggressive query reporting)
- we don't differentiate between built-in and custom queries: if it matches a particular Testing Library query pattern, is reported (aggressive query reporting)

Let me know if you have any questions or need help to hook the rule with the new detection mechanism!

thomlom:
I was about to implement that enhancement, and there's something that bugged me. The goal of this enhancement is to disallow this kind of statement:

```
screen.queryByText("My button")
```

It can happen if someone is mistaken about `getBy` and `queryBy` queries. But if something ever happens, what should we suggest to the user? Maybe he doesn't like explicit assertions and wants to use `screen.getByText` in the end. 

Then, maybe that rule should be implemented in `prefer-presence-queries`. When the user uses `screen.queryByText`, it's suggested to use `screen.getByText`. And if the user has `prefer-explicit-assert` enabled, then the rule will fire.

What do you think, @Belco90?

Belco90:
Mmm not sure to be honest, but definitely something to be handled combining `prefer-presence-queries` and `prefer-explicit-assert`.

I think we should just suggest in `prefer-explicit-assert` to wrap `screen.queryByText("My button")` within some assertion, no matter which one. Then, if that assertion is checking element is present (e.g. `expect(screen.queryByText("My button")
[...]

PR #274, opened 2020-12-22
refactor(prefer-explicit-assert): use new utils and remove custom query option

thomlom:
## Changes

- Move `prefer-explicit-assert` to ESLint shared configuration
- Remove custom query option as aggressive reporting will be enabled by default
- Added more extensive test cases: when the user configures a custom module and cover all queries

## Notes

For #269, I'll add this change in another PR 🙂

### DIFF
--- a/tests/lib/rules/prefer-explicit-assert.test.ts
+++ b/tests/lib/rules/prefer-explicit-assert.test.ts
@@ -6,86 +6,96 @@ const ruleTester = createRuleTester();
 
 ruleTester.run(RULE_NAME, rule, {
   valid: [
-    {
-      code: `getByText`,
-    },
-    {
-      code: `const utils = render()
-      
-      utils.getByText
+    ...ALL_QUERIES_METHODS.map((queryMethod) => ({
+      code: `get${queryMethod}('Hello')`,
+      settings: {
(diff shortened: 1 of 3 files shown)


tests/lib/rules/prefer-explicit-assert.test.ts
```diff
@@ -6,86 +6,96 @@ const ruleTester = createRuleTester();
```
thomlom, line 6: Let me know if I added enough tests as I'm not totally confident with the latest changes and what it implies in the different rules 🙂

Belco90: The tests variants are more than fine! I would suggest to include a custom query method for testing this rule. You only need to create `const COMBINED_QUERIES_METHODS = [...ALL_QUERIES_METHODS, 'ByIcon']` and use this array of methods rather than `ALL_QUERIES_METHODS`

Belco90 (changes requested):
Looking good! I'm gonna request just to include a custom query method within the tests as described in my comment. Thanks!

Belco90 (approved):
I didn't realize you already addressed the requested change! Looks good, thank you so much for migrating this rule 🚀

PR #281, opened 2021-03-07
refactor: several tweaks

Belco90:
Relates to #198

Pull Request not related to any particular rule. The purpose of this PR is to fix some checks, extract some commons patterns and general improvements:

- Rename `testing-library/module` setting to `testing-library/utils-module`
- Move detection helpers types to simplify their definitions
- Add new `getReferenceNode` and `getPropertyIdentifierNode` node util methods
- Rename `getIdentifierNode` to `getDeepestIdentifierNode` to avoid any misunderstanding with its behavior
- Add `isTestingLibraryUtil` method to extract common logic from `isAsyncUtil` and `isRenderUtil`
- Check inside `isAsyncUtil` if coming from Testing Library or not
- Improve how `isNodeComingFromTestingLibrary` obtained node identifier so it detects chained properties correctly
- Refactor rules using `isAsyncUtil` to avoid checking if coming from Testing Library manually
- Add several tests to fake rule to cover new detection helpers better

I mentioned I wanted to use glob pattern for `testing-library/filename-pattern` setting instead of regexp, but it doesn't make sense since it's just the name of the file, not its path. It's still using a string regexp then, but I'll need to clarify in the docs that it needs double backslash to scape special chars.

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -5,8 +5,9 @@ import {
 } from '@typescript-eslint/experimental-utils';
 import {
   getAssertNodeInfo,
-  getIdentifierNode,
   getImportModuleName,
+  getPropertyIdentifierNode,
+  getReferenceNode,
   ImportModuleNode,
   isImportDeclaration,
   isImportNamespaceSpecifier,
@@ -60,9 +61,9 @@ type IsFindQueryVariantFn = (node: TSESTree.Identifier) => boolean;
 type IsSyncQueryFn = (node: TSESTree.Identifier) => boolean;
 type IsAsyncQueryFn = (node: TSESTree.Identifier) => boolean;
 type IsCustomQueryFn = (node: TSESTree.Identifier) => boolean;
-type IsAsyncUtilFn = (node: TSESTree.Node) => boolean;
+type IsAsyncUtilFn = (node: TSESTree.Identifier) => boolean;
 type IsFireEventMethodFn = (node: TSESTree.Identifier) => boolean;
-type IsRenderUtilFn = (node: TSESTree.Node) => boolean;
+type IsRenderUtilFn = (node: TSESTree.Identifier) => boolean;
 type IsPresenceAssertFn = (node: TSESTree.MemberExpression) => boolean;
 type IsAbsenceAssertFn = (node: TSESTree.MemberExpression) => boolean;
 type CanReportErrorsFn = () => boolean;
@@ -128,33 +129,19 @@ export function detectTestingLibraryUtils<
      * related to Testing Library or not.
      */
     function isTestingLibraryUtil(
-      node: TSESTree.Node,
+      node: TSESTree.Identifier,
       isUtilCallback: (identifierNode: TSESTree.Identifier) => boolean
     ): boolean {
-      const identifierNode = getIdentifierNode(node);
-
-      if (!identifierNode) {
+      if (!isUtilCallback(node)) {
         return false;
       }
 
-      if (!isUtilCallback(identifierNode)) {
-        return false;
-      }
-
-      const referenceNode = (function () {
-        if (isMemberExpression(node)) {
-          return node;
-        }
-
-        if (node.parent && isMemberExpression(node.parent)) {
-          return node.parent;
-        }
-        return identifierNode;
-      })();
+      const referenceNode = getReferenceNode(node);
+      const referenceNodeIdentifier = getPropertyIdentifierNode(referenceNode);
 
       return (
         isAggressiveModuleReportingEnabled() ||
-        isNodeComingFromTestingLibrary(referenceNode)
+        isNodeComingFromTestingLibrary(referenceNodeIdentifier)
       );
     }
 
@@ -468,7 +455,7 @@ export function detectTestingLibraryUtils<
       }
 
       if (!identifierName) {
-        return;
+        return false;
       }
 
       return !!findImportedUtilSpecifier(identifierName);
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -47,32 +47,54 @@ export type EnhancedRuleCreate<
   detectionHelpers: Readonly<DetectionHelpers>
 ) => TRuleListener;
 
-export type DetectionHelpers = {
-  getTestingLibraryImportNode: () => ImportModuleNode | null;
-  getCustomModuleImportNode: () => ImportModuleNode | null;
-  getTestingLibraryImportName: () => string | undefined;
-  getCustomModuleImportName: () => string | undefined;
-  isTestingLibraryImported: () => boolean;
-  isValidFilename: () => boolean;
-  isGetQueryVariant: (node: TSESTree.Identifier) => boolean;
-  isQueryQueryVariant: (node: TSESTree.Identifier) => boolean;
-  isFindQueryVariant: (node: TSESTree.Identifier) => boolean;
-  isSyncQuery: (node: TSESTree.Identifier) => boolean;
-  isAsyncQuery: (node: TSESTree.Identifier) => boolean;
-  isCustomQuery: (node: TSESTree.Identifier) => boolean;
-  isAsyncUtil: (node: TSESTree.Identifier) => boolean;
-  isFireEventMethod: (node: TSESTree.Identifier) => boolean;
-  isRenderUtil: (node: TSESTree.Node) => boolean;
-  isPresenceAssert: (node: TSESTree.MemberExpression) => boolean;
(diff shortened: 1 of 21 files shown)


lib/detect-testing-library-utils.ts
```diff
@@ -398,32 +436,23 @@ export function detectTestingLibraryUtils<
...
-      if (ASTUtils.isIdentifier(node)) {
-        identifierName = node.name;
-      } else if (ASTUtils.isIdentifier(node.object)) {
-        identifierName = node.object.name;
-      }
-
-      if (!identifierName) {
-        return;
-      }
+      const identifierName: string | undefined = getPropertyIdentifierNode(node)
```
Belco90, line 449: With this change it reports properly those nodes chained more than once.

lib/rules/await-async-utils.ts
```diff
@@ -46,16 +45,6 @@ export default createTestingLibraryRule<Options, MessageIds>({
     return {
       'CallExpression Identifier'(node: TSESTree.Identifier) {
         if (helpers.isAsyncUtil(node)) {
-          if (
```
Belco90, line 49: This is now checked directly within `isAsyncUtil`

lib/rules/no-wait-for-snapshot.ts
```diff
@@ -37,15 +37,14 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
           ASTUtils.isIdentifier(callExpression.callee) &&
-          helpers.isNodeComingFromTestingLibrary(callExpression.callee) &&
           helpers.isAsyncUtil(callExpression.callee)
         ) {
           return callExpression.callee;
         }
         if (
           isMemberExpression(callExpression.callee) &&
           ASTUtils.isIdentifier(callExpression.callee.property) &&
-          helpers.isNodeComingFromTestingLibrary(callExpression.callee)
```
Belco90, line 48: This is now checked directly within `isAsyncUtil`

lib/rules/render-result-naming-convention.ts
```diff
@@ -32,7 +32,13 @@ export default createTestingLibraryRule<Options, MessageIds>({
   create(context, _, helpers) {
     return {
       VariableDeclarator(node) {
-        if (!helpers.isRenderUtil(node.init)) {
```
Belco90, line 35: This was the wrong approach.

Before: it was passing the whole call expression to `isRenderUtil`, leaving this helper to find the deepest identifier to be reported. This was like that to simplify getting the reference node to determine if imported from Testing Library package or not.

Now: it expects to receive the identifier you actually want to report, and then `isRenderUtil` will make sure it gets corresponding reference node to check if it was imported from Testing Library package or not.

You can get a better understanding of this by checking `isTestingLibraryUtil` and new `getReferenceNode`/`getPropertyIdentifierNode` methods.

tests/create-testing-library-rule.test.ts
```diff
@@ -160,21 +147,39 @@ ruleTester.run(RULE_NAME, rule, {
       getSomeElement('button')
     `,
     },
+    {
+      code: `
+      // case: custom method not matching "getBy*" variant pattern using within
+      within(container).getSomeElement('button')
```
Belco90, line 153: Added extra tests to make sure queries reporting works fine with chained calls.

tests/create-testing-library-rule.test.ts
```diff
@@ -229,32 +264,69 @@ ruleTester.run(RULE_NAME, rule, {
...
+    {
+      settings: { 'testing-library/utils-module': 'test-utils' },
+      code: `
+      // case: aggressive module disabled and render coming from non-related module
+      import * as somethingElse from '@somewhere/else'
       import { render } from '@testing-library/react'
       
-      const utils = render()
+      // somethingElse.render is not coming from any module related to TL
+      const utils = somethingElse.render()
```
Belco90, line 329: This new test has detected few issues in our implementation! I'm glad I added it at the beginning of this tweaks branch. It will make sure from now on that we check from where are imported those methods in member expressions properly.

tests/fake-rule.ts
```diff
@@ -87,12 +99,6 @@ export default createTestingLibraryRule<Options, MessageIds>({
     return {
       'CallExpression Identifier': reportCallExpressionIdentifier,
       MemberExpression: reportMemberExpression,
-      'CallExpression > MemberExpression'(node: TSESTree.MemberExpression) {
```
Belco90, line 90: This was a silly report while implementing proper Testing Library utils detection. Now that are all implemented, it can be removed.

tests/lib/rules/await-async-utils.test.ts
```diff
@@ -103,18 +103,28 @@ ruleTester.run(RULE_NAME, rule, {
       `,
     })),
     ...ASYNC_UTILS.map((asyncUtil) => ({
+      settings: {
+        'testing-library/utils-module': 'test-utils',
+      },
       code: `
         import { ${asyncUtil} } from 'some-other-library';
-        test('util "${asyncUtil}" which is not related to testing library is valid', async () => {
+        test(
```
Belco90, line 111: After checking if coming from Testing Library directly inside `isAsyncUtil`, this test was invalid. I opted-out aggressive reporting so this test keeps passing, and moved the original one to invalid test cases.

tests/lib/rules/await-async-utils.test.ts
```diff
@@ -174,16 +184,6 @@ ruleTester.run(RULE_NAME, rule, {
         });
       `,
     },
-    ...ASYNC_UTILS.map((asyncUtil) => ({
```
Belco90, line 177: Duplicated.

tests/lib/rules/no-await-sync-query.test.ts
```diff
@@ -69,7 +69,7 @@ ruleTester.run(RULE_NAME, rule, {
     },
     // sync query awaited but not matching filename pattern is invalid but not reported
     {
-      settings: { 'testing-library/filename-pattern': '^.*\\.(nope)\\.js$' },
+      settings: { 'testing-library/filename-pattern': 'nope\\.js' },
```
Belco90, line 72: Simplifying the pattern name in the test for clarification.

tests/lib/rules/no-wait-for-snapshot.test.ts
```diff
@@ -51,73 +51,101 @@ ruleTester.run(RULE_NAME, rule, {
       `,
     })),
     ...ASYNC_UTILS.map((asyncUtil) => ({
+      settings: {
+        'testing-library/utils-module': 'test-utils',
+      },
       code: `
         import { ${asyncUtil} } from 'some-other-library';
-        test('snapshot calls within ${asyncUtil} are not valid', async () => {
+        test('aggressive reporting disabled - snapshot calls within ${asyncUtil} not related to Testing Library are valid', async () => {
```
Belco90, line 59: After checking if coming from Testing Library directly inside `isAsyncUtil`, this test was invalid. I opted-out aggressive reporting so this test keeps passing, and moved the original one to invalid test cases.

Belco90 (commented):
Sorry for the noise added by renaming `testing-library/module` setting 😓

lib/detect-testing-library-utils.ts
```diff
@@ -150,84 +195,92 @@ export function detectTestingLibraryUtils<
...
      * or custom module are imported.
      */
-    const isTestingLibraryImported: DetectionHelpers['isTestingLibraryImported'] = () => {
-      if (isAggressiveModuleReportingEnabled()) {
-        return true;
-      }
-
-      return !!importedTestingLibraryNode || !!importedCustomModuleNode;
+    const isTestingLibraryImported: IsTestingLibraryImportedFn = () => {
+      return (
```
thomlom, line 203: I personally find the old syntax more readable but it's nit-picking.

lib/detect-testing-library-utils.ts
```diff
@@ -150,84 +195,92 @@ export function detectTestingLibraryUtils<
...
     /**
-     * Determines whether a given node is async util or not.
+     * Determines whether a given node is a valid async util or not.
+     *
+     * A node will be interpreted as a valid async util based on two conditions:
+     * the name matches with some Testing Library async util, and the node is
+     * coming from Testing Library module.
+     *
+     * The latter depends on Aggressive
+     * module reporting: if enabled, then it doesn't matter from
```
thomlom, line 269: The line break seems weird, is it intended?

thomlom (approved):
Everything looks good to me. A second pair of eyes would be nice, though 😄

lib/detect-testing-library-utils.ts
Belco90: Not really. It probably go split in two lines after some rewording. I'll fix it.

PR #282, opened 2021-03-11
refactor(render-result-naming-convention): refine checks to decide if coming from Testing Library

Belco90:
Relates to #198

As mentioned in [the PR for migrating `render-result-naming-convention` rule to the new mechanism](https://github.com/testing-library/eslint-plugin-testing-library/pull/280#discussion_r585942050), this new PR aims to check nested/wrapped render calls.

After doing so, I came up with checking misleading renamed imports like:
```javascript
import { render as testingLibrary } from '@testing-library/react'
import { render } from 'somewhere-else'

// ...

// this render must not be reported when aggressive reporting is disabled
render()
```

It turned out we were reporting this case, so I've refined the "is coming from Testing Library" check to make sure the renamed import is what we are looking for. It was based on @gndelia's check for `fireEvent` but extracted into few utils and made more generic, without changing `isFireEventMethod` since it needs the actual import clause to do more checks.

After this refinement, I'm really confident about the whole detection mechanism. I don't expect to find major issues migrating the remaining rules, so I hope I can create several PRs in the coming days migrating most of them 💪

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -139,10 +139,25 @@ export function detectTestingLibraryUtils<
       const referenceNode = getReferenceNode(node);
       const referenceNodeIdentifier = getPropertyIdentifierNode(referenceNode);
 
-      return (
-        isAggressiveModuleReportingEnabled() ||
-        isNodeComingFromTestingLibrary(referenceNodeIdentifier)
+      if (isAggressiveModuleReportingEnabled()) {
+        return true;
+      }
+
+      // TODO: extract this into function, combined with logic from isFireEventMethod
+      // TODO: include some tests create-testing-library-rule
+      const importNode = findImportedUtilSpecifier(
+        referenceNodeIdentifier.name
       );
+
+      if (!importNode) {
+        return false;
+      }
+
+      if (ASTUtils.isIdentifier(importNode)) {
+        return importNode.name === referenceNodeIdentifier.name;
+      }
+
+      return importNode.local.name === referenceNodeIdentifier.name;
     }
 
     /**
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -8,6 +8,7 @@ import {
   getImportModuleName,
   getPropertyIdentifierNode,
   getReferenceNode,
+  hasImportMatch,
   ImportModuleNode,
   isImportDeclaration,
   isImportNamespaceSpecifier,
@@ -127,6 +128,12 @@ export function detectTestingLibraryUtils<
     /**
      * Small method to extract common checks to determine whether a node is
      * related to Testing Library or not.
+     *
+     * To determine whether a node is a valid Testing Library util, there are
+     * two conditions to match:
+     * - it's named in a particular way (decided by given callback)
+     * - it's imported from valid Testing Library module (depends on aggressive
+     *    reporting)
      */
     function isTestingLibraryUtil(
       node: TSESTree.Identifier,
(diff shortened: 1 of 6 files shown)


lib/detect-testing-library-utils.ts
```diff
@@ -439,17 +447,26 @@ export function detectTestingLibraryUtils<
...
+     * This method will try to find any import matching the given node name,
+     * and also make sure the name is a valid match in case it's been renamed.
      */
     const isNodeComingFromTestingLibrary: IsNodeComingFromTestingLibraryFn = (
       node
     ) => {
       const identifierName: string | undefined = getPropertyIdentifierNode(node)
         .name;
 
-      return !!findImportedUtilSpecifier(identifierName);
```
Belco90, line 452: If any import found, it was assuming the given node was related to the import. But that's not always the case. In the example from the PR description "render" import was found but not the one renamed, so adding this extra check fixes it!

lib/rules/render-result-naming-convention.ts
```diff
@@ -30,15 +35,33 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
         const initIdentifierNode = getDeepestIdentifierNode(node.init);
 
         if (!initIdentifierNode) {
           return;
         }
 
-        if (!helpers.isRenderUtil(initIdentifierNode)) {
+        if (
+          !helpers.isRenderUtil(initIdentifierNode) &&
+          !renderWrapperNames.includes(initIdentifierNode.name)
```
Belco90, line 63: Detecting nested/wrapped `render` usages was as simple as adding these few lines of code 😨

tests/create-testing-library-rule.test.ts
```diff
@@ -642,26 +670,6 @@ ruleTester.run(RULE_NAME, rule, {
...
-    {
-      settings: {
-        'testing-library/utils-module': 'test-utils',
-      },
-      code: `
-        // case: waitFor from object property shadowed name is checked correctly
-        import * as tl from 'test-utils'
-        const obj = { tl }
-        
-        obj.module.waitFor(() => {})
```
Belco90, line 654: I don't even know what I was trying to test here 😂

tests/lib/rules/prefer-wait-for.test.ts
```diff
@@ -218,6 +218,21 @@ ruleTester.run(RULE_NAME, rule, {
...
       `,
     },
+    {
+      settings: {
+        'testing-library/utils-module': 'test-utils',
+      },
+      code: `
+      // case: aggressive reporting disabled - method named same as invalid method
+      // but not coming from Testing Library is valid
+      import { wait as testingLibraryWait } from 'test-utils'
```
Belco90, line 228: Adding an extra test to this rule since was the only one using `isNodeComingFromTestingLibrary` directly.

tests/lib/rules/render-result-naming-convention.test.ts
```diff
@@ -134,6 +134,20 @@ ruleTester.run(RULE_NAME, rule, {
...
+        import { render } from '@somewhere/else'
+        
+        const setup = () => render(<SomeComponent />);
+
+        test('aggressive reporting disabled - should not report nested render not related to TL', () => {
+          const wrapper = setup();
+          const button = wrapper.getByText('some button');
+        });
+      `,
+    },
```
Belco90, line 150: This is the test which helped me to discover the improvement needed on imports checking.

tests/lib/rules/render-result-naming-convention.test.ts
```diff
@@ -343,5 +406,29 @@ ruleTester.run(RULE_NAME, rule, {
...
+        {
+          messageId: 'renderResultNamingConvention',
+          data: {
+            renderResultName: 'wrapper',
+          },
+          line: 9,
+          column: 17,
+        },
+      ],
+    },
```
Belco90, line 432: I can't believe yet this is reported correctly 💯

timdeschryver (approved):
👀 I might steal this later for another plugin, nicely done!

Belco90:
> 👀 I might steal this later for another plugin, nicely done!

Thank you! Let me know if you want me to show you any particular part of the new mechanism in detail!

Belco90:
I covered even more cases around renamed imports!

PR #283, opened 2021-03-13
fix: add no-await-sync-events exception for userEvent.keyboard

ph-fritsche:
Added exception for the new API proposed in https://github.com/testing-library/user-event/pull/581

Note that with the change `userEvent.type` (and `userEvent.keyboard`) will actually return no Promise if the `delay` is not `>0` so that the typing and the actual return value will match.

Further it might be worth discussing if this rule should be disabled for `.ts` and `.tsx` files as `ts(80007)` and `@typescript-eslint/await-thenable` already warn here and curating a manual list has no benefits when there already are warnings based on the type.

### DIFF
--- a/lib/rules/no-await-sync-events.ts
+++ b/lib/rules/no-await-sync-events.ts
@@ -44,7 +44,7 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
               property.key.name === 'delay'
           );
 
-        if (!(node.name === 'userEvent' && methodNode.name === 'type' && withDelay)) {
+        if (!(node.name === 'userEvent' && (methodNode.name === 'type' || methodNode.name === 'keyboard') && withDelay)) {
           context.report({
             node: methodNode,
             messageId: 'noAwaitSyncEvents',
--- a/lib/rules/no-await-sync-events.ts
+++ b/lib/rules/no-await-sync-events.ts
@@ -44,7 +44,7 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
               property.key.name === 'delay'
           );
 
-        if (!(node.name === 'userEvent' && (methodNode.name === 'type' || methodNode.name === 'keyboard') && withDelay)) {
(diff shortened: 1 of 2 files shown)


Belco90:
Hey @ph-fritsche, thanks for your contribution! We need to wait until https://github.com/testing-library/user-event/pull/581 gets merged.

Meanwhile, could you add some additional tests to check `userEvent.keyboard` is reported or not as intended?

About disabling the rule in TS environments: the plugin won't detect that for now, so it's up to the user to decide if they want to enable or disable the rule.

lib/rules/no-await-sync-events.ts
```diff
@@ -44,7 +44,7 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
               property.key.name === 'delay'
           );
 
-        if (!(node.name === 'userEvent' && methodNode.name === 'type' && withDelay)) {
+        if (!(node.name === 'userEvent' && (methodNode.name === 'type' || methodNode.name === 'keyboard') && withDelay)) {
```
MichaelDeBoey, line 47:
```suggestion
        if (!(node.name === 'userEvent' && ['type', 'keyboard'].includes(methodNode.name) && withDelay)) {
```

MichaelDeBoey (changes requested):

gndelia (commented):
should we update a test to include the `keyboard` method ? 🤔

Belco90:
I just realized this rule is still [pending to be refactored for v4](https://github.com/testing-library/eslint-plugin-testing-library/issues/198). @ph-fritsche I can include `userEvent.keyboard` there when I refactor it. What do you think?

ph-fritsche:
@Belco90 I've added the changes necessary. Should be ready to be merged now. :smiley:

Belco90:
LGTM, thanks for your PR and nice work adding this new feature to `userEvent` 🚀

Belco90 (approved):

Belco90:
@ph-fritsche sorry mate, I forgot to add you as a contributor!

@all-contributors please add @ph-fritsche for code

Belco90:
@all-contributors please add @ph-fritsche for code

PR #285, opened 2021-03-14
Move rules settings to ESLint shared config: refactor prefer-screen-queries rule

Belco90:
Relates to #198

This refactor for `prefer-screen-queries` includes:
- detect only queries coming from valid renders
- minor tweaks and fixes

I had an idea to simplify the whole rule implementation. However, since this one works fine, I'll leave it for now as I don't have that amount of time to spend on this.

### DIFF
--- a/lib/rules/prefer-screen-queries.ts
+++ b/lib/rules/prefer-screen-queries.ts
@@ -1,16 +1,13 @@
+import { ASTUtils, TSESTree } from '@typescript-eslint/experimental-utils';
+import { ALL_QUERIES_COMBINATIONS } from '../utils';
 import {
-  ESLintUtils,
-  TSESTree,
-  ASTUtils,
-} from '@typescript-eslint/experimental-utils';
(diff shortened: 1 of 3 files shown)


lib/rules/prefer-screen-queries.ts
```diff
@@ -77,63 +90,61 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
           return;
         }
+
+        const isComingFromValidRender = helpers.isRenderUtil(node.init.callee);
+
+        if (!isComingFromValidRender) {
+          // save the destructured query methods as safe since they are coming
+          // from render not related to TL
+          saveSafeDestructuredQueries(node);
+        }
```
Belco90, line 100: This is the only addition to previous implementation, so it reports only queries coming from valid renders.

gndelia (approved):
LGTM

timdeschryver (approved):
LGTM.
I was a bit confused with the `isComingFromValidRender ` variable name, but the comment that they don't come from TL helped.

PR #286, opened 2021-03-14
ci: migrate to GitHub Actions

### LINKED ISSUE #275: Move release flow from travis to github actions
Following [this](https://github.com/testing-library/dom-testing-library/pull/809#issue-515554670) PR by @kentcdodds, we're migrating all travis pipelines to github actions :)

An example for the action can be seen [here](https://github.com/testing-library/dom-testing-library/blob/master/.github/workflows/validate.yml).

Belco90:
Closes #275


Belco90:
No clue why it's failing when installing dependencies. Any idea @MatanBobi ?

Belco90:
It seemed related to Node v15 which installs npm 7. I've removed Node v15 for now since I don't see any obvious way of running npm 6 with it. I'll investigate this in the future.

MatanBobi:
> No clue why it's failing when installing dependencies only in the tests job. Any idea @MatanBobi ?

If I understand correctly, it looks like the version of `ts-jest` we're using isn't compatible with out `typescript` version (it only supports `typescript@">=3.4 <4.0"` so IMO we should upgrade `ts-jest`..
I guess this raised only now because npm 7 had some changes related to peer dependencies (I think it now actually tries to install them).

Belco90:
@MatanBobi It definitely must be that. Since we have updated all dependencies in v4, I'll take care of this there when this one gets merged.

Belco90:
I'm merging this since I need it to keep going with v4. I'll find out if the release job works fine now 😅

.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Use Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+        with:
+          useLockFile: false
```
gndelia, line 33: I think it is not necessary to use an extra GH action here, while a simple 

```yaml
name: Install dependencies
run: npm install
```

should do the trick

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          node-version: '14'
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+        with:
+          useLockFile: false
+
+      - name: Build package
+        run: npm run build
```
gndelia, line 38: should we build in `ci.yml` ? I think we should (without releasing ofc)

gndelia (commented):
I left some suggestions / questions, I'm sorry I was late to the party 😅

.github/workflows/ci.yml
Belco90: This npm-install action will cache stuff related to npm properly :)

.github/workflows/release.yml
Belco90: Mmmm not really sure about that, what would be the purpose for it? I'm gonna merge both workflows into a single file actually, since it's triggering the release twice.

.github/workflows/ci.yml
gndelia: ohh nice

.github/workflows/release.yml
gndelia: to ensure there are no compilation errors (Typings for instance) committed without noticing?

Belco90: Ah! I'm doing that on v4 so on "Code Validation" job I'll add format check and typescript check steps. I'll create a PR later with those updates for v4!

PR #289, opened 2021-03-16
ci: update pipeline with v4 changes

Belco90:
Adapt changes from #286 to v4


.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,117 @@
...
+        with:
+          node-version: '14'
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+        with:
+          useLockFile: false
+
+      - name: Check Types
+        run: npm run type-check
```
Belco90, line 42: @gndelia we are checking types here now

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,117 @@
...
+
+      - name: Lint code
+        run: npm run lint
+
+      - name: Check format
+        run: npm run format:check -- --max-warnings 0
```
Belco90, line 48: Format check step is back

Belco90:
I don't know why the tests with node v10 are failing (the error messages are really weird). It's passing fine in my local:

imagen

Belco90:
Ok, I was using ESLint v7.19 in my local. If I upgrade to v7.22 locally (same as the pipeline) then I get the errors. I'll take a look at this.

MichaelDeBoey:
Maybe we can just drop Node 10 as that's going to be EOL very soon (2021-04-30)?
https://nodejs.org/en/about/releases

gndelia:
dropping node10 is a breaking change, so v4 sounds like a good opportunity to do so, otherwise we'll have to wait for v5

Belco90:
Dropping node 10 is definitely a possibility but I’d like to investigate the issue first (it fails on v4 but not on main, and I don’t see any obvious reason).

MichaelDeBoey:
On the other hand: ESLint doesn't seem to drop Node 10  soon (as can be seen in https://github.com/eslint/eslint/issues/14023), so maybe we should keep the same dependencies as ESLint has?

MichaelDeBoey:
@Belco90 I haven't looked deeply into the problem or tried to solve it, but this one seems to be a good starting point in figuring out what's wrong I think: https://github.com/jsdom/jsdom/issues/2795

Belco90:
> On the other hand: ESLint doesn't seem to drop Node 10 soon (as can be seen in [eslint/eslint#14023](https://github.com/eslint/eslint/issues/14023)), so maybe we should keep the same dependencies as ESLint has?

Sounds like a good plan to me. Even if Node LTS support gets dropped in one month an a half, there are tons of projects which haven't been migrated to newer Node yet. So v10 will be around for a bit. I don't have any problem with releasing another major to drop node v10 support in the future.


> @Belco90 I haven't looked deeply into the problem or tried to solve it, but this one seems to be a good starting point in figuring out what's wrong I think: [jsdom/jsdom#2795](https://github.com/jsdom/jsdom/issues/2795)

Taking a look, thanks!

Belco90:
I'm not sure jsdom is a problem here, since we don't render anything to any DOM but to ESLint test render.

I've located the PR where this started failing: it's #282 from few days ago. But still it's really strange: we had 2 pipelines run in Travis CI due to some misconfig (one from branch push and one from PR update). One of them is fine and the other isn't. The only different I found is that the success one installed the dependencies from cache and the failure one installed the deps fresh without cache. Since our GitHub Actions is fresh, it's installing deps without cache, so I'd say the problem could be some dependency getting a new version installed which causes this?

I tried by downgrading the deps to the very version indicated in our package.json and it's still failing, so I'm gonna try to downgrade them one by one to versions fixed in `main` branch. Any other ideas?

Belco90:
I think I found it: if I downgrade both `jest` and `ts-jest` to v25 the tests are ok again. I don't see any obvious reason preventing us from using v26 of these two (actually there are some warnings about unexpected results when running v25).

Belco90:
So it's just `jest` v26. If I just downgrade that one to v25, everything works fine (always referring to node v10). I'm thinking now it actually could be the jsdom @MichaelDeBoey mentioned before? I don't see how, but it's one of the changes introduced in jest v26. I'll keep investigating.

MichaelDeBoey:
I think you could use `jest-environment-jsdom-sixteen` instead for the environment.
Have a look in the upgrade docs for Jest v26, I know they mention it somewhere

Belco90:
> I think you could use `jest-environment-jsdom-sixteen` instead for the environment.
> Have a look in the upgrade docs for Jest v26, I know they mention it somewhere

Well, not sixteen (which is the jsdom version buil-tin with jest v26) but fifteen. But it works! Thanks for your help @MichaelDeBoey 💯

MichaelDeBoey:
Glad I could be of assistence!

PR #293, opened 2021-03-20
Move rules settings to ESLint shared config: refactor no-debug rule 

Belco90:
Relates to #198 

This refactor for `no-debug` includes:

- improve test coverage and errors assertions

### DIFF
--- a/docs/rules/no-debug.md
+++ b/docs/rules/no-debug.md
@@ -1,6 +1,6 @@
 # Disallow the use of `debug` (no-debug)
 
-Just like `console.log` statements pollutes the browser's output, debug statements also pollutes the tests if one of your team mates forgot to remove it. `debug` statements should be used when you actually want to debug your tests but should not be pushed to the codebase.
+Just like `console.log` statements pollutes the browser's output, debug statements also pollutes the tests if one of your teammates forgot to remove it. `debug` statements should be used when you actually want to debug your tests but should not be pushed to the codebase.
 
 ## Rule Details
 
@@ -26,12 +26,6 @@ screen.debug();
 ```js
 const { screen } = require('@testing-library/react');
 screen.debug();
-```
-
-If you use [custom render functions](https://testing-library.com/docs/example-react-redux) then you can set a config option in your `.eslintrc` to look for these.
-
-```
-   "testing-library/no-debug": ["error", {"renderFunctions":["renderWithRedux", "renderWithRouter"]}],
 ```
 
 ## Further Reading
--- a/lib/node-utils.ts
+++ b/lib/node-utils.ts
@@ -392,7 +392,7 @@ export function getPropertyIdentifierNode(
 }
 
 /**
- * Gets the deepest identifier node from a given node.
+ * Gets the deepest identifier node in the expression from a given node.
  *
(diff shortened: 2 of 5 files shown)


lib/node-utils.ts
```diff
@@ -392,7 +392,7 @@ export function getPropertyIdentifierNode(
 }
 
 /**
- * Gets the deepest identifier node from a given node.
+ * Gets the deepest identifier node in the expression from a given node.
```
Belco90, line 395: Clarifying the scope for getting the deepest identifier.

lib/node-utils.ts
```diff
@@ -416,11 +416,15 @@ export function getDeepestIdentifierNode(
...
+  if (ASTUtils.isAwaitExpression(node)) {
+    return getDeepestIdentifierNode(node.argument);
+  }
+
   return null;
 }
 
 /**
- * Gets the farthest node from a given node.
+ * Gets the farthest node in the expression from a given node.
```
Belco90, line 427: Clarifying the scope for getting the farthest identifier.

lib/node-utils.ts
```diff
@@ -416,11 +416,15 @@ export function getDeepestIdentifierNode(
     return getDeepestIdentifierNode(node.callee);
   }
 
+  if (ASTUtils.isAwaitExpression(node)) {
```
Belco90, line 419: I discovered this case thanks to existing tests from `no-debug` !

lib/rules/no-debug.ts
```diff
@@ -46,154 +36,67 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
+          for (const property of node.id.properties) {
+            if (
               isProperty(property) &&
               ASTUtils.isIdentifier(property.key) &&
-              property.key.name === 'screen'
-          );
-      },
-      // checks if import has shape:
-      // import { screen } from '@testing-library/dom';
-      ImportDeclaration(node: TSESTree.ImportDeclaration) {
```
Belco90, line 114: it's really pleasant to remove this unnecessary checks 😌

MichaelDeBoey (approved):

Belco90:
@timdeschryver @thomaslombart @gndelia Merging but, as always, I can address whatever feedback you may have in another PR.

docs/rules/no-debug.md
```diff
@@ -28,12 +28,6 @@ const { screen } = require('@testing-library/react');
 screen.debug();
 ```
 
-If you use [custom render functions](https://testing-library.com/docs/example-react-redux) then you can set a config option in your `.eslintrc` to look for these.
```
timdeschryver, line 31: This LGTM, I just wonder if we should mention that it can be configured via settings?

Belco90: Yes, I'm leaving all the documentation stuff for the last step after I've migrated all the rules. I'll mention the custom render settings in the README, but not sure if we should mention it on each rule depending on `render` (which is what I understand you are suggesting).

timdeschryver: Yea, correct. While I was writing it I was also considering if it's worth it to mention in for every rule or not 😅.
I think having a global doc is good.

PR #295, opened 2021-03-20
Move rules settings to ESLint shared config: refactor no-container rule 

Belco90:
Relates to #198

This refactor for `no-container` includes:

- improve location of the node reported
- improve test coverage and errors assertions
- add ability to detect `container` usage coming from functions wrapping `render`
- `no-debug`: add ability to detect `debug` usage coming from functions wrapping `render`

Ideally, the ability to detect functions wrapping a _reportable_ `render` should be part of `isRenderUtil`, or extracted into some helper. However, I'm gonna replicate the same behavior I applied in this PR (and others previously) to detect wrappers around `render` for remaining rules. After v4 I'll spend more time trying to make that check generic.

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -154,6 +154,10 @@ export function detectTestingLibraryUtils<
         originalNodeName?: string
       ) => boolean
     ): boolean {
+      if (!node) {
+        return false;
+      }
+
       const referenceNode = getReferenceNode(node);
       const referenceNodeIdentifier = getPropertyIdentifierNode(referenceNode);
       const importedUtilSpecifier = getImportedUtilSpecifier(
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -5,6 +5,7 @@ import {
 } from '@typescript-eslint/experimental-utils';
 import {
   getAssertNodeInfo,
+  getDeepestIdentifierNode,
   getImportModuleName,
   getPropertyIdentifierNode,
   getReferenceNode,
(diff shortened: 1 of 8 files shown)


lib/detect-testing-library-utils.ts
```diff
@@ -149,6 +154,10 @@ export function detectTestingLibraryUtils<
         originalNodeName?: string
       ) => boolean
     ): boolean {
+      if (!node) {
```
Belco90, line 157: Safety check. Apparently there was some place calling this method with `null` node. I need to investigate this properly, but I did this workaround meanwhile.

lib/node-utils.ts
```diff
@@ -477,31 +477,6 @@ export function isRenderFunction(
   });
 }
 
-// TODO: should be removed after v4 is finished
-export function isRenderVariableDeclarator(
```
Belco90, line 481: Removed finally. A better equivalent has been created in `detect-testing-library-utils.ts`

lib/rules/no-container.ts
```diff
@@ -29,48 +27,52 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
           containerCallsMethod =
             ASTUtils.isIdentifier(innerNode.property) &&
             innerNode.property.name === 'container' &&
             isRenderWrapper;
 
-          if (isContainerName || containerCallsMethod) {
+          if (containerCallsMethod) {
             context.report({
-              node: innerNode,
+              node: innerNode.property,
```
Belco90, line 72: If `container` is chained, report `container` node rather than chained one.

lib/rules/no-container.ts
```diff
@@ -120,6 +99,47 @@ export default ESLintUtils.RuleCreator(getDocsUrl)<Options, MessageIds>({
...
+        const isRenderWrapperVariableDeclarator = initIdentifierNode
+          ? renderWrapperNames.includes(initIdentifierNode.name)
+          : false;
+
+        if (
+          !helpers.isRenderVariableDeclarator(node) &&
+          !isRenderWrapperVariableDeclarator
+        ) {
+          return;
+        }
```
Belco90, line 115: The rule implementation is mostly the same, except this bit to detect _reportable_ `render` or some wrapper around it.

lib/rules/no-debug.ts
```diff
@@ -41,12 +43,28 @@ export default createTestingLibraryRule<Options, MessageIds>({
   create(context, [], helpers) {
     const suspiciousDebugVariableNames: string[] = [];
     const suspiciousReferenceNodes: TSESTree.Identifier[] = [];
+    const renderWrapperNames: string[] = [];
+
+    function detectRenderWrapper(node: TSESTree.Identifier): void {
```
Belco90, line 48: I forgot to detect this properly in the previous PR, so I added it here.

timdeschryver (approved):

gndelia (commented):
Late to the party, but great work here 🚀 🎉

PR #308, opened 2021-04-02
fix: first round of bugfixes for v4

### LINKED ISSUE #304: [v4][no-await-sync-query] False positives on appearance/disappearance examples
In https://testing-library.com/docs/guide-disappearance, there's some techniques outlined for waiting for elements to be removed:

```js
test('movie title appears', async () => {
  // element is initially not present...

  // wait for appearance inside an assertion
  await waitFor(() => {
    expect(getByText('the lion king')).toBeInTheDocument()
  })
})
```

```
test('movie title no longer present in DOM', async () => {
  // element is removed
  await waitForElementToBeRemoved(() => queryByText('the mummy'))
})
```

Both of these examples are flagged by `no-await-sync-query`, and they probably shouldn't be.

### LINKED ISSUE #307: First round of v4 beta issues
The first round of issues I found while testing v4 beta:

- problem A: `await-async-utils` doesn't report the right node name
  CleanShot 2021-03-28 at 20 54 18@2x
- problem B: `prefer-find-by` is reporting an incorrect message
  CleanShot 2021-03-28 at 21 24 41@2x
- problem C: `no-await-sync-query` reports an error when a parent is awaited (related to #304 )
  CleanShot 2021-03-28 at 20 55 14@2x
- problem D: `no-wait-for-multiple-assertions` should display the error on each assertion other than the first one instead of a single error on `waitFor`
- problem E: `no-wait-for-side-effects` should display the error on each side effect instead of a single error on `waitFor`
- ~~problem F: ESLint `peerDependency` seems set incorrectly~~ this is actually fine, it seems a warning due to installing it through `resolutions` and ESLint being a sub-dependency of `react-scripts`
  CleanShot 2021-03-29 at 21 54 25@2x

Belco90:
Closes #307. Closes #304

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -335,12 +335,20 @@ export function detectTestingLibraryUtils<
      * Not to be confused with {@link isUserEventMethod}
      */
     const isUserEventUtil = (node: TSESTree.Identifier): boolean => {
-      return isTestingLibraryUtil(
-        node,
-        (identifierNodeName, originalNodeName) => {
-          return [identifierNodeName, originalNodeName].includes('userEvent');
-        }
-      );
+      const userEvent = findImportedUserEventSpecifier();
+      let userEventName: string | undefined;
+
+      if (userEvent) {
+        userEventName = userEvent.name;
+      } else if (isAggressiveModuleReportingEnabled()) {
+        userEventName = USER_EVENT_NAME;
+      }
+
+      if (!userEventName) {
+        return false;
+      }
+
+      return node.name === userEventName;
     };
 
     /**
--- a/lib/node-utils.ts
+++ b/lib/node-utils.ts
@@ -153,19 +153,6 @@ export function findClosestCallNode(
   }
 }
 
-export function isCallExpressionCallee(
-  node: TSESTree.CallExpression,
-  identifier: TSESTree.Identifier
-): boolean {
-  const nodeInnerIdentifier = getDeepestIdentifierNode(node);
-
-  if (nodeInnerIdentifier) {
-    return nodeInnerIdentifier.name === identifier.name;
-  }
-
-  return false;
-}
-
 export function isObjectExpression(
   node: TSESTree.Expression
 ): node is TSESTree.ObjectExpression {
--- a/lib/rules/await-async-utils.ts
+++ b/lib/rules/await-async-utils.ts
@@ -80,7 +80,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
                   node,
                   messageId: 'awaitAsyncUtil',
                   data: {
-                    name: referenceNode.name,
+                    name: node.name,
                   },
                 });
               }
--- a/lib/rules/no-await-sync-query.ts
+++ b/lib/rules/no-await-sync-query.ts
@@ -1,9 +1,6 @@
 import { TSESTree } from '@typescript-eslint/experimental-utils';
 import { createTestingLibraryRule } from '../create-testing-library-rule';
-import {
-  findClosestCallExpressionNode,
-  isCallExpressionCallee,
-} from '../node-utils';
+import { getDeepestIdentifierNode } from '../node-utils';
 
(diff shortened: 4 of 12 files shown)


lib/detect-testing-library-utils.ts
```diff
@@ -335,12 +335,20 @@ export function detectTestingLibraryUtils<
...
+
+      if (userEvent) {
+        userEventName = userEvent.name;
+      } else if (isAggressiveModuleReportingEnabled()) {
+        userEventName = USER_EVENT_NAME;
+      }
+
+      if (!userEventName) {
+        return false;
+      }
```
Belco90, line 349: Same issue I had with `isUserEventMethod`: `userEvent` util won't be imported from Testing Library module and it's a default import, so it has to be checked against its own import specifier.

lib/node-utils.ts
```diff
@@ -153,19 +153,6 @@ export function findClosestCallNode(
   }
 }
 
-export function isCallExpressionCallee(
```
Belco90, line 156: Not used anymore. It was a little bit useless, to be honest.

lib/rules/await-async-utils.ts
```diff
@@ -80,7 +80,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
                   node,
                   messageId: 'awaitAsyncUtil',
                   data: {
-                    name: referenceNode.name,
+                    name: node.name,
```
Belco90, line 83: This fixes problem A from #307

lib/rules/no-await-sync-query.ts
```diff
@@ -29,22 +26,19 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
 
-        if (helpers.isSyncQuery(node)) {
+        if (helpers.isSyncQuery(deepestIdentifierNode)) {
           context.report({
-            node,
+            node: deepestIdentifierNode,
             messageId: 'noAwaitSyncQuery',
             data: {
-              name: node.name,
+              name: deepestIdentifierNode.name,
```
Belco90, line 41: This fixes problem C from #307

lib/rules/no-wait-for-multiple-assertions.ts
```diff
@@ -46,14 +54,20 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
         return;
       }
 
-      context.report({
-        node: callExpressionNode,
-        messageId: 'noWaitForMultipleAssertion',
-      });
+      for (let i = 0; i < expectNodes.length; i++) {
+        if (i !== 0) {
+          context.report({
```
Belco90, line 65: This fixes problem D from #307

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -49,14 +57,17 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+      const sideEffectNodes = getSideEffectNodes(node.body);
+      if (sideEffectNodes.length === 0) {
         return;
       }
 
-      context.report({
-        node: callExpressionNode,
-        messageId: 'noSideEffectsWaitFor',
-      });
+      for (const sideEffectNode of sideEffectNodes) {
```
Belco90, line 65: This fixes problem E from #307

lib/rules/prefer-find-by.ts
```diff
@@ -53,13 +53,13 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
       description:
-        'Suggest using find* instead of waitFor to wait for elements',
+        'Suggest using `find*` query instead of `waitFor` + `get*` to wait for elements',
       category: 'Best Practices',
       recommended: 'warn',
     },
     messages: {
       preferFindBy:
-        'Prefer {{queryVariant}}{{queryMethod}} method over using await {{fullQuery}}',
+        'Prefer `{{queryVariant}}{{queryMethod}}` query over using `{{waitForMethodName}}` + `{{prevQuery}}`',
```
Belco90, line 62: This fixes problem B from #307

timdeschryver (approved):

PR #309, opened 2021-04-02
fix: guard against null identifier nodes

### LINKED ISSUE #305: [v4] Crash `Cannot read property 'name' of null` in detect-testing-library-utils getImportedUtilSpecifier
One of our tests has some code that looks like:

```ts
(window as any).__THING = false;
```

This is causing a crash with `Cannot read property 'name' of null`.

Looking at [`getImportedUtilSpecifier`](https://github.com/testing-library/eslint-plugin-testing-library/blob/82ace4946fca958698adcc367ba3e8fa9ab070a4/lib/detect-testing-library-utils.ts#L604-L611), it calls `getPropertyIdentifierNode` and checks `.name` on it, even though that [can return `null`](https://github.com/testing-library/eslint-plugin-testing-library/blob/82ace4946fca958698adcc367ba3e8fa9ab070a4/lib/node-utils.ts#L401). It looks like the root cause might be `isExpressionStatement` not handling `TSAsExpression` types.

### LINKED ISSUE #306: [v4][no-debug] Crash with `Cannot read property 'name' of null`
We've got some code that looks like:

```js
thing.method.lastCall.args[0]();
```

This is causing a crash in the CallExpression handler in `no-debug`. In this case, `callExpressionIdentifier` is null. I suspect that `getDeepestIdentifierNode` doesn't handle array accessors correctly, and it also [can explicitly return `null`](https://github.com/testing-library/eslint-plugin-testing-library/blob/beta/lib/node-utils.ts#L433).

Belco90:
Closes #305. Closes #306.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -6,12 +6,6 @@ module.exports = {
   },
   coverageThreshold: {
     global: {
-      branches: 100,
-      functions: 100,
-      lines: 100,
-      statements: 100,
-    },
-    './lib/node-utils.ts': {
       branches: 90,
       functions: 90,
       lines: 90,
--- a/lib/rules/no-debug.ts
+++ b/lib/rules/no-debug.ts
@@ -26,7 +26,6 @@ export default createTestingLibraryRule<Options, MessageIds>({
     messages: {
       noDebug: 'Unexpected debug statement',
     },
-    fixable: null,
     schema: [],
   },
   defaultOptions: [],
@@ -46,6 +45,9 @@ export default createTestingLibraryRule<Options, MessageIds>({
 
     return {
       VariableDeclarator(node) {
+        if (!node.init) {
+          return;
+        }
         const initIdentifierNode = getDeepestIdentifierNode(node.init);
 
         if (!initIdentifierNode) {
--- a/lib/rules/no-debug.ts
+++ b/lib/rules/no-debug.ts
@@ -101,6 +101,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
         const referenceNode = getReferenceNode(node);
         const referenceIdentifier = getPropertyIdentifierNode(referenceNode);
 
+        if (!referenceIdentifier) {
+          return;
+        }
+
         const isDebugUtil = helpers.isDebugUtil(callExpressionIdentifier);
         const isDeclaredDebugVariable = suspiciousDebugVariableNames.includes(
           callExpressionIdentifier.name
--- a/lib/rules/no-debug.ts
+++ b/lib/rules/no-debug.ts
@@ -48,6 +48,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
       VariableDeclarator(node) {
         const initIdentifierNode = getDeepestIdentifierNode(node.init);
 
+        if (!initIdentifierNode) {
+          return;
+        }
+
         const isRenderWrapperVariableDeclarator = initIdentifierNode
           ? renderWrapperNames.includes(initIdentifierNode.name)
           : false;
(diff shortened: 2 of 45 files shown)


jest.config.js
```diff
@@ -6,12 +6,6 @@ module.exports = {
...
     global: {
-      branches: 100,
-      functions: 100,
-      lines: 100,
-      statements: 100,
-    },
-    './lib/node-utils.ts': {
       branches: 90,
       functions: 90,
       lines: 90,
```
Belco90, line 11: I've dropped the coverage threshold a bit since I prefer to guard against null nodes everywhere even if there is no _possible_ code for it. We'll see what I do with this after releasing v4.

lib/rules/no-debug.ts
```diff
@@ -83,13 +89,22 @@ export default createTestingLibraryRule<Options, MessageIds>({
       },
       CallExpression(node) {
         const callExpressionIdentifier = getDeepestIdentifierNode(node);
```
Belco90, line 91: @timdeschryver This is really weird: `getDeepestIdentifierNode` (and some other utils for finding identifiers) are typed as returning `TSESTree.Identifier | null`. However, TS is assuming they always returning `TSESTree.Identifier` and never `null`:

CleanShot 2021-04-02 at 12 02 28

Then, if I update the function to `return ''` instead of `return null` and update the returned type to `TSESTree.Identifier | string`, TS starts complaining about returned value saved in all these rules might not be an `Identifier`. Is this a bug on TS? Am I missing something else? I don't get why it always assumes `null` won't be returned.

tests/lib/rules/no-debug.test.ts
```diff
@@ -139,6 +139,10 @@ ruleTester.run(RULE_NAME, rule, {
       debug()
       `,
     },
+
+    `// cover edge case for https://github.com/testing-library/eslint-plugin-testing-library/issues/306
+    thing.method.lastCall.args[0]();
```
Belco90, line 144: Test for issue mentioned by @epmatsw

lib/rules/no-debug.ts
epmatsw: I suspect turning on [strictNullChecks](https://www.typescriptlang.org/tsconfig#strictNullChecks) in the tsconfig would fix that.

timdeschryver: Oh wow... TIL...
@epmatsw is right, you'll have to turn the strict mode on.
I would encourage to turn on strict mode though, not just strict null checks.

Belco90: Aah damn! I thought we had `strict` enabled, but anyway I didn't expect those nullable values to be silent. I'm gonna enable strict mode and fix all corresponding issues then. Thanks guys.

timdeschryver (approved):

Belco90:
Ok, so I've enabled TS `strict` mode and fix all issues. But now tests files generating invalid cases spreading an array of cases are flagged as errors by TS. I'm not sure what to do with that. Any ideas? @timdeschryver @epmatsw

epmatsw:
Hmm. Maybe throw an `as const` somewhere in there?

Belco90:
> Hmm. Maybe throw an `as const` somewhere in there?

That fixed almost all of them. The problem is the `messageId` for the errors, so TS complains about receiving an `string` but is declared as an enum. The `as const` does the trick for it. I'm learning more TS with this PR than in my whole life 😂 

Now I got some tests failing, I'm looking into them.

Belco90:
Everything ok now, merging to test it on beta 3.

PR #311, opened 2021-04-04
fix(prefer-user-event): format list of userEvent methods correctly 

### LINKED ISSUE #310: [v4][prefer-user-event] malformed error message for `fireEvent.keyDown`
We have code that (minimized for repro) looks like:

```js
await fireEvent.keyDown(thing);
```

That produces an error: `Prefer using  or userEvent.tab() over fireEvent.keyDown()`. Seems like the first alternative suggestion is missing (note the double-space).

This is with `4.0.0-beta.3`

Belco90:
Closes #310

### DIFF
--- a/lib/rules/prefer-user-event.ts
+++ b/lib/rules/prefer-user-event.ts
@@ -23,7 +23,7 @@ export const UserEventMethods = [
 type UserEventMethodsType = typeof UserEventMethods[number];
 
 // maps fireEvent methods to userEvent. Those not found here, do not have an equivalent (yet)
-export const MappingToUserEvent: Record<string, UserEventMethodsType[]> = {
+export const MAPPING_TO_USER_EVENT: Record<string, UserEventMethodsType[]> = {
   click: ['click', 'type', 'selectOptions', 'deselectOptions'],
   change: ['upload', 'type', 'clear', 'selectOptions', 'deselectOptions'],
   dblClick: ['dblClick'],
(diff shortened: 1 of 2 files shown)


lib/rules/prefer-user-event.ts
```diff
@@ -49,17 +49,14 @@ export const MappingToUserEvent: Record<string, UserEventMethodsType[]> = {
...
-    (method: string) => `userEvent.${method}()`
+  const userEventMethods = MAPPING_TO_USER_EVENT[fireEventMethod].map(
+    (methodName) => `userEvent.${methodName}`
   );
-  const { length } = allMethods;
 
-  const init = length > 2 ? allMethods.slice(0, length - 2).join(', ') : '';
-  const last = `${length > 1 ? ' or ' : ''}${allMethods[length - 1]}`;
-  return `${init}${last}`;
+  return userEventMethods.join(', ').replace(/, ([a-zA-Z.]+)$/, ', or $1');
```
Belco90, line 56: I'd like to just use `Intl.ListFormat` here, but it's available from node v13 so we can't stick to it until we drop node v10 and node v12 compatibility 😅

tests/lib/rules/prefer-user-event.test.ts
```diff
@@ -28,6 +28,26 @@ function createScenarioWithImport<
 
 const ruleTester = createRuleTester();
 
+function formatUserEventMethodsMessage(fireEventMethod: string): string {
```
Belco90, line 31: I reimplemented here the same method so I don't use the very same implementation. Otherwise, it would be tricky to catch errors formatting the message.

tests/lib/rules/prefer-user-event.test.ts
```diff
@@ -295,10 +320,21 @@ ruleTester.run(RULE_NAME, rule, {
           endColumn: 22,
           data: {
             userEventMethods:
-              'userEvent.click(), userEvent.type() or userEvent.deselectOptions()',
+              'userEvent.click, userEvent.type, userEvent.selectOptions, or userEvent.deselectOptions',
```
Belco90, line 323: @epmatsw nice catch opening this issue! This test should have caught that, but has the wrong expected output :(

PR #314, opened 2021-04-04
fix: second round of bugfixes for v4

### LINKED ISSUE #313: Second round of v4 beta issues
The second round of issues I found while testing v4 beta:
- problem A: `Cannot read property 'properties' of undefined` from a jest config file when importing jest-dom
```javascript
require('@testing-library/jest-dom')
// ...
foo()
```
- problem B: `prefer-user-event` shouldn't be included in any preset, since `user-event` might not be installed. In the future, it would be nice to detect if `user-event` installed to auto-enable this rule. For now, it's better to leave out of all presets
- problem C: `no-debug` should be set as "error" in presets, in the same way as `no-console` is at "error" level on `eslint:recommended` preset.

Belco90:
Closes #313


Belco90:
Merging this to keep testing beta version. If you have any feedback, let me know and I'll address it in a separate PR.

Issue #186, opened 2020-06-22
ESLint plugin Testing Library v4


Belco90:
I'd like to keep everything planned for v4 update here, so it's easier for everyone involved to organize pending work. This was started in #165.

**Changes planned for [v4](https://github.com/testing-library/eslint-plugin-testing-library/milestone/2):**

- [x] new rule: `no-container` - #171, #177 
- [x] new rule: `render-result-naming-convention` - #170, #200 
- [x] new rule: `prefer-user-event` - #162, #192 
- [x] new rule: `no-promise-in-fire-event` -  #166, #180 
- [x] new rule: `no-wait-for-side-effects` - #134, #196 
- [x] new rule: `no-node-access` - #190 
- [x] new rule: `no-wait-for-multiple-assertions` - #133, #189 
- [x] BREAKING CHANGE: enable `no-wait-for-empty-callback` in config presets - #168
- [x] BREAKING CHANGE: enable `prefer-screen-queries` in config presets - #169
- [x] BREAKING CHANGE: rename `recommended` config to `dom` - #182, #184 
- [x] BREAKING CHANGE: move rules settings to ESLint shared config - #198 
- [x] ci: migrate CI to [GitHub Actions](https://github.blog/2021-03-04-4-things-you-didnt-know-you-could-do-with-github-actions/#generate-semantic-release-notes-with-github-actions)

**Extra:**

- [x] Cleanup after all rules are migrated: #303 
- [x] Update README with shared settings: #317 
- [X] Update README to explain "aggressive mode", its motivation and how to restrict its scope: #312
- [x] Update rules docs to delete references to removed options
- [x] Update CONTRIBUTING to mention new rules creation + detection helper + realistic test cases: #321 
- [X] Write migrating to v4 guide: #312
- [x] Prepare release notes

**Future Improvements:**
- Allow multiple utils modules on Shared Settings: #315 
- New rule: `no-unnecessary-act` - #259 

Thanks to @gndelia @timdeschryver @nickmccurdy @thebinaryfelix @renatoagds for your work on this one!

Belco90:
Do you think we need to wait for those 4 new rules for releasing v4? I think `render-result-naming-convention` and `prefer-user-event` are necessary, but we could release `prefer-user-event` and `no-multiple-assertions-wait-for` in a later step?

nickmccurdy:
[...]

PR #317, opened 2021-04-06
docs: update README for v4

Belco90:
Relates to #186 

I'm not sure if the Shared Settings description should be the same on README and migrating guide 🤔 . What do you think? /cc @MichaelDeBoey @CreativeTechGuy


README.md
```diff
@@ -123,6 +146,107 @@ To enable this configuration use the `extends` property in your
 }
 ```
 
+## Configuration
+
+There are some configuration options available that will be shared across all the plugin rules. This is achieved using [ESLint Shared Settings](https://eslint.org/docs/user-guide/configuring/configuration-files#adding-shared-settings). These shared settings are meant to be used if you need to restrict the Aggressive Reporting mechanism, **so please before configuring any of these settings**, read more about [`eslint-plugin-testing-library` Aggressive Reporting mechanism](docs/migrating-to-v4-guide.md#aggressive-reporting), and [how it's affected by these settings](docs/migrating-to-v4-guide.md#shared-settings).
```
CreativeTechGuy, line 151: I think it'd be valuable to call out here that restricting Aggressive Reporting is a specialized/advanced feature. The name makes it seem bad and scary and might make people think they need to configure this. In reality, this is not needed for most people to ever touch.

README.md
```diff
@@ -123,6 +146,107 @@ To enable this configuration use the `extends` property in your
...
+
+If you are sure about configuring the settings, these are the options available:
```
CreativeTechGuy, line 153: Does it make sense to put all of this in the main README given that it's not something that most users would need? Maybe a separate page that explains this in detail and this document can just contain the summary and explanation of what can be configured and why someone might need to configure it. Probably doesn't need to go into such detail about how to actually do it.

Belco90: Yeah, I think I'm gonna move this Configuration section to the very bottom of the README, put this detailed explanation into migrating guide, and leave just the "how to" here, referencing the detailed version.

README.md
```diff
@@ -54,22 +62,37 @@ Add `testing-library` to the plugins section of your `.eslintrc` configuration f
...
+// ❌ Don't do this
+{
+  "extends": ["plugin:testing-library/dom", "plugin:testing-library/react"]
+}
+
+// ✅ Do just this instead
+{
+  "extends": ["plugin:testing-library/react"]
+}
+```
```
gndelia, line 95: nice example here, as it might confuse people 👏

gndelia (approved):
Looks great 🚀  I love how it is looking!

PR #332, opened 2021-04-12
test: add extra tests

MichaelDeBoey:
Took some inspiration from `eslint-plugin-jest` (see https://github.com/jest-community/eslint-plugin-jest/blob/main/src/__tests__/rules.test.ts).

With these tests, we ensure that
- all rules have docs
- all rules have tests
- all configs are referencing existing rules

### DIFF
--- a/tests/index.test.ts
+++ b/tests/index.test.ts
@@ -1,25 +1,67 @@
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const { configs, rules } = require('../lib');
-import * as fs from 'fs';
-import * as path from 'path';
+import { existsSync } from 'fs';
+import { resolve } from 'path';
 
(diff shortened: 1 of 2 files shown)


tests/index.test.ts
```diff
@@ -1,25 +1,67 @@
...
+  expect(Object.keys(allConfigs)).toEqual(['dom', 'angular', 'react', 'vue']);
+  const allConfigRules = Object.values(allConfigs)
+    .map((config) => Object.keys(config.rules))
+    .reduce((previousValue, currentValue) => [
+      ...previousValue,
+      ...currentValue,
+    ]);
+
+  allConfigRules.forEach((rule) => {
+    const ruleNamePrefix = 'testing-library/';
```
Belco90, line 59: I guess this can be declared outside the loop.

Belco90 (approved):

PR #338, opened 2021-04-13
fix(no-node-access): skip reporting files without Testing Library import

### LINKED ISSUE #334: limit `no-node-access` errors reported
Originated by #331. I don't think `no-node-access` should report DOM elements queried from `document` since we can't detect properly if being used for a test because of Aggressive Reporting.

In the rule doc it says this case is invalid:
> // If is not importing a testing-library package
>
> `document.getElementById('submit-btn').closest('button');`

But we can't actually know this for the reason mentioned before.

Belco90:
Closes #334

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -56,7 +56,7 @@ type GetTestingLibraryImportNodeFn = () => ImportModuleNode | null;
 type GetCustomModuleImportNodeFn = () => ImportModuleNode | null;
 type GetTestingLibraryImportNameFn = () => string | undefined;
 type GetCustomModuleImportNameFn = () => string | undefined;
-type IsTestingLibraryImportedFn = () => boolean;
+type IsTestingLibraryImportedFn = (isStrict?: boolean) => boolean;
 type IsGetQueryVariantFn = (node: TSESTree.Identifier) => boolean;
 type IsQueryQueryVariantFn = (node: TSESTree.Identifier) => boolean;
 type IsFindQueryVariantFn = (node: TSESTree.Identifier) => boolean;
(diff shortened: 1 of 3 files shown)


lib/detect-testing-library-utils.ts
```diff
@@ -242,11 +242,15 @@ export function detectTestingLibraryUtils<
      * then this method will return `true` ONLY IF a testing-library package
      * or custom module are imported.
      */
-    const isTestingLibraryImported: IsTestingLibraryImportedFn = () => {
+    const isTestingLibraryImported: IsTestingLibraryImportedFn = (
+      isStrict = false
+    ) => {
+      const isSomeModuleImported =
+        !!importedTestingLibraryNode || !!importedCustomModuleNode;
```
MichaelDeBoey, line 249:
```suggestion
        Boolean(importedTestingLibraryNode) || Boolean(importedCustomModuleNode);
```

Belco90: Why so picky with this detail?

MichaelDeBoey: Whenever I see a double negation, I try to replace it by using the `Boolean` function, as that makes the intent a lot clearer to new people (especially for less experienced developers)

Belco90: I guess it's more _self-explanatory_, but I'd say it's not consistent with the rest of the codebase. I'd prefer tackling this sort of styling preferences in a separate PR, changing all related code to leave the codebase consistent in that sense (and even enabling some ESLint rule if necessary, like `no-implicit-coercion` for this case).

MichaelDeBoey: A separate PR sounds good to me

Belco90: Are we happy to merge this then? 😅

MichaelDeBoey (approved):
LGTM! 👍

PR #297, opened 2021-03-22
refactor: remove mechanism to match files to be reported

Belco90:
Relates to #198 

Remove everything related to mechanism for matching files to be reported, as discussed in #296 

@MichaelDeBoey adding you here in case you want to double-check it too!

PR #339, opened 2021-04-13
docs: update README with more info around Aggressive Reporting

### LINKED ISSUE #331: V4 plugin runs against code
I upgraded to the latest version, 4.0.0, and I cannot build my code anymore since the plugin finds errors in my code.

With the following configuration :
```
extends: ['airbnb', 'plugin:prettier/recommended', 'plugin:testing-library/react'],
plugins: ['react-hooks', 'testing-library'],
```
I am getting the ```no-node-access``` error for the code ```document.getElementById('root')``` :
```
ReactDOM.render(
  <CommProvider useDsa={false}>
    <ThemeProvider>
      <GlobalStyle />
      <Suspense fallback={<Loader />}>
        <AppLogin />
      </Suspense>
    </ThemeProvider>
  </CommProvider>,
  document.getElementById('root')
);
```

I was expecting the rules to be applied to only .test.js files.

Belco90:
Closes #331

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,10 +281,7 @@ If you think the error you are getting is not related to this at all, please [fi
 
 If you are getting false positive ESLint errors in your testing files, this could be caused by [Aggressive Reporting](#aggressive-reporting).
 
-You can avoid this by:
-
-1. [running `eslint-plugin-testing-library` only against testing files](#run-the-plugin-only-against-test-files)
-2. [limiting the scope of Aggressive Reporting through Shared Settings](#shared-settings)
+You can avoid this by [limiting the scope of Aggressive Reporting through Shared Settings](#shared-settings)
 
 If you think the error you are getting is not related to this at all, please [fill a new issue](https://github.com/testing-library/eslint-plugin-testing-library/issues/new/choose) with as many details as possible.
 
--- a/README.md
+++ b/README.md
@@ -74,7 +74,8 @@ Then configure the rules you want to use within `rules` property of your `.eslin
 
 With the default setup mentioned before, `eslint-plugin-testing-library` will be run against your whole codebase. If you want to run this plugin only against your tests files, you have the following options:
 
-**ESLint `overrides`**
+#### ESLint `overrides`
+
 One way of restricting ESLint config by file patterns is by using [ESLint `overrides`](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-based-on-glob-patterns).
 
 Assuming you are using the same pattern for your test files as [Jest by default](https://jestjs.io/docs/configuration#testmatch-arraystring), the following config would run `eslint-plugin-testing-library` only against your test files:
@@ -98,7 +99,8 @@ Assuming you are using the same pattern for your test files as [Jest by default]
 };
 ```
 
-**ESLint Cascading and Hierachy**
+#### ESLint Cascading and Hierachy
+
 Another approach for customizing ESLint config by paths is through [ESLint Cascading and Hierachy](https://eslint.org/docs/user-guide/configuring/configuration-files#cascading-and-hierarchy). This is useful if all your tests are placed under the same folder, so you can place there another `.eslintrc` where you enable `eslint-plugin-testing-library` for applying it only to the files under such folder, rather than enabling it on your global `.eslintrc` which would apply to your whole project.
 
 ## Shareable configurations
--- a/README.md
+++ b/README.md
@@ -72,27 +72,34 @@ Then configure the rules you want to use within `rules` property of your `.eslin
 
 ### Run the plugin only against test files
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,6 +70,30 @@ Then configure the rules you want to use within `rules` property of your `.eslin
...
+  // here you have your usual config which applies to the whole project...
+  extends: ['airbnb', 'plugin:prettier/recommended'],
+  plugins: ['react-hooks'],
+
+  overrides: [
+    {
+      // ... and here you expand the ESLint plugins run for your tests files
+      files: ['**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+(spec|test).[jt]s?(x)'],
+      extends: ['plugin:testing-library/react'],
+      plugins: ['testing-library'],
```
MichaelDeBoey, line 91: No need to do this, as this is already included in the config I think?

README.md
```diff
@@ -70,6 +70,30 @@ Then configure the rules you want to use within `rules` property of your `.eslin
...
 
+### Run the plugin only against test files
+
+With the default setup mentioned before `eslint-plugin-testing-library` will be run against your whole codebase. If want to run this plugin only against your tests files, you can do that by using [ESLint `overrides`](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-based-on-glob-patterns).
+
+Assuming you are using the same pattern for your test files as [Jest by default](https://jestjs.io/docs/configuration#testmatch-arraystring), the following config would run `eslint-plugin-testing-library` only against your test files:
+
+```javascript
+// .eslintrc
+{
```
MichaelDeBoey, line 81:
```suggestion
module.exports = {
```

README.md
```diff
@@ -227,6 +257,30 @@ A list of function names that are valid as Testing Library custom renders. Relat
...
+If you think the error you are getting is not related to this at all, please [fill a new issue](https://github.com/testing-library/eslint-plugin-testing-library/issues/new/choose) with as many details as possible.
+
+### There are false positives in testing files
+
+If you are getting false positive ESLint errors in your testing files, this could be caused by [Aggressive Reporting](#aggressive-reporting).
+
+You can avoid this by:
+
+1. [running `eslint-plugin-testing-library` only against testing files](#run-the-plugin-only-against-test-files)
+2. [limiting the scope of Aggressive Reporting through Shared Settings](#shared-settings)
```
MichaelDeBoey, line 280: We're already in testing files, so the only option is limiting the scope I think?

```suggestion
You can avoid this by [limiting the scope of Aggressive Reporting through Shared Settings](#shared-settings)
```

MichaelDeBoey (changes requested):

README.md
Belco90: Good point, I'll move this to the main config side.

README.md
Belco90: Actually, I'm gonna rewrite it as json for consistency with the rest of the examples here.

README.md
Belco90: We are not running the plugin only for testing files. We had a discussion around this when [I implemented the glob pattern for reporting only on testing files](https://github.com/testing-library/eslint-plugin-testing-library/pull/296#issuecomment-804216452), which finally [was discarded and the whole mechanism was reverted](https://github.com/testing-library/eslint-plugin-testing-library/pull/297) in favor leaving the user to decide which files to lint with proper ESLint config.

MichaelDeBoey: In this case we are, hence the "There are false positives in testing files" title

https://github.com/testing-library/eslint-plugin-testing-library/pull/339/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R280

README.md
```diff
@@ -70,6 +70,37 @@ Then configure the rules you want to use within `rules` property of your `.eslin
 }
 ```
 
+### Run the plugin only against test files
+
+With the default setup mentioned before, `eslint-plugin-testing-library` will be run against your whole codebase. If you want to run this plugin only against your tests files, you have the following options:
+
+**ESLint `overrides`**
```
MichaelDeBoey, line 77: Making this a title would be better I think

```suggestion
#### ESLint `overrides`

```

README.md
```diff
@@ -70,6 +70,37 @@ Then configure the rules you want to use within `rules` property of your `.eslin
...
+    {
+      // 3) Now we enable eslint-plugin-testing-library rules or preset only for matching files!
+      "files": ["**/__tests__/**/*.[jt]s?(x)", "**/?(*.)+(spec|test).[jt]s?(x)"],
+      "extends": ["plugin:testing-library/react"]
+    },
+  ],
+};
+```
+
+**ESLint Cascading and Hierachy**
```
MichaelDeBoey, line 101:
```suggestion
#### ESLint Cascading and Hierachy

```

MichaelDeBoey (approved):
LGTM! 👊

Would make these 2 a title instead of just bold though

MichaelDeBoey (approved):

PR #342, opened 2021-04-15
fix(prefer-screen-queries): avoid reporting custom queries

### LINKED ISSUE #340: testing-library/prefer-screen-queries with self-written helpers
Hey

```
expect(await findByComplexText(screen, 'Selected campaigns 0')).toBeInTheDocument()
```

`eslint-plugin-testing-library@3.10.2` - no errors

`eslint-plugin-testing-library@4.0.1` - Avoid destructuring queries from `render` result, use `screen.findByComplexText` instead  testing-library/prefer-screen-queries

should this rule only apply to `testing-library` methods?

Belco90:
Closes #340


Belco90:
Merging this to release the corresponding patch.

PR #346, opened 2021-04-19
refactor: automatically generate all rules

MichaelDeBoey:
Took some inspiration from `eslint-plugin-jest` (see https://github.com/jest-community/eslint-plugin-jest/blob/913bc73673338c644b6f5eee60981b39ac991e39/src/index.ts#L18-L39).


lib/rules/index.ts
```diff
@@ -0,0 +1,29 @@
...
+const rulesDir = __dirname;
+const excludedFiles = ['index'];
+
+export default readdirSync(rulesDir)
+  .map((rule) => parse(rule).name)
+  .filter((ruleName) => !excludedFiles.includes(ruleName))
+  .reduce<Record<string, RuleModule>>(
+    (allRules, ruleName) => ({
+      ...allRules,
+      [ruleName]: importDefault(join(rulesDir, ruleName)) as RuleModule,
```
Belco90, line 26: Can't we extract here the actual type of each rule? It could be interesting to get that to generate docs automatically for each rule.

Belco90 (changes requested):
Too much magic but seems like good automation. I wonder if would be possible to get proper rule types in `rules/index.ts`

lib/rules/index.ts
MichaelDeBoey: I can't think of a way to do this tbh, but I'm open to suggestions

Belco90: I guess the type is getting lost by the `importDefault` return value type? Maybe @timdeschryver can help us here 🙏

MichaelDeBoey: `importDefault` is indeed the issue here I think, but that one needs to be generic in order for my automation to work

Belco90: Would help if we assume we are gonna always have ES module default exports? So we can remove `interopRequireDefault` and simplify `importDefault`, which could help to get the proper type from each module.

MichaelDeBoey: `importDefault` will then return of type `any`, so that's not the solution you're looking for either I'm afraid 😕

Belco90: Of course... we would have to manually import all rules to get corresponding types. I'll think about it, but we can probably merge this one until we actually need those types.

timdeschryver: I don't think it's possible to interfer the types here. If needed we could look into adding the types manually since these would all be Rules.

MichaelDeBoey: The real question is: do we need the types outside of the plugin? 🤔
If not, we're fine as we still have types for our tests, ...

Belco90: We don't need it outside the plugin, that wasn't my idea either. But at some point, we will generate rules docs from rules meta info automatically, and I don't know if at that point we would need the types for being able to generate the docs properly.

MichaelDeBoey: Docs will never be completely generated automatically, we'll only need meta info for creating the tables in README and for creating configs automatically

That's already included in the `TSESLint.RuleModule` type, so no need to worry about that

Belco90: I wasn't expecting to generate all the docs automatically for sure :) Anyway, it should be enough with `TSESLint.RuleModule` indeed. Thanks!

Belco90 (approved):

PR #349, opened 2021-04-20
fix: allow console.debug() in no-debug

### LINKED ISSUE #327: no-debug also affects console.debug()
The `testing-library/no-debug` test also warns about using `console.debug()`, although there is a separate built-in test for that: `no-console`. I would like to allow `console.debug()` statements (our config contains `"no-console": ["error", {"allow": ["debug", "warn", "error"]}]`), but still keep `testing-library/no-debug` enabled for testing-library debug statements.

gndelia:
closes #327

I intentionally left out the following scenario

```js
const { debug } = console
debug()
```

because it goes against the first invalid tests that against any `debug()`, it raises a flag.

### DIFF
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -16,7 +16,6 @@ import {
   isImportNamespaceSpecifier,
   isImportSpecifier,
   isLiteral,
-  isCallExpression,
   isMemberExpression,
   isObjectPattern,
   isProperty,
@@ -584,11 +583,8 @@ export function detectTestingLibraryUtils<
     const isDebugUtil: IsDebugUtilFn = (identifierNode) => {
       const isBuiltInConsole =
         isMemberExpression(identifierNode.parent) &&
-        identifierNode.parent.parent &&
-        isCallExpression(identifierNode.parent.parent) &&
-        isMemberExpression(identifierNode.parent.parent.callee) &&
-        ASTUtils.isIdentifier(identifierNode.parent.parent.callee.object) &&
-        identifierNode.parent.parent.callee.object.name === 'console';
+        ASTUtils.isIdentifier(identifierNode.parent.object) &&
+        identifierNode.parent.object.name === 'console';
 
       return (
         !isBuiltInConsole &&
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -16,6 +16,7 @@ import {
   isImportNamespaceSpecifier,
   isImportSpecifier,
   isLiteral,
+  isCallExpression,
   isMemberExpression,
   isObjectPattern,
   isProperty,
@@ -78,10 +79,7 @@ type IsRenderUtilFn = (node: TSESTree.Identifier) => boolean;
 type IsRenderVariableDeclaratorFn = (
   node: TSESTree.VariableDeclarator
 ) => boolean;
-type IsDebugUtilFn = (
-  identifierNode: TSESTree.Identifier,
-  callExpressionNode: TSESTree.CallExpression
-) => boolean;
+type IsDebugUtilFn = (identifierNode: TSESTree.Identifier) => boolean;
 type IsPresenceAssertFn = (node: TSESTree.MemberExpression) => boolean;
 type IsAbsenceAssertFn = (node: TSESTree.MemberExpression) => boolean;
 type CanReportErrorsFn = () => boolean;
@@ -583,11 +581,14 @@ export function detectTestingLibraryUtils<
       return isRenderUtil(initIdentifierNode);
     };
 
-    const isDebugUtil: IsDebugUtilFn = (identifierNode, callExpressionNode) => {
+    const isDebugUtil: IsDebugUtilFn = (identifierNode) => {
       const isBuiltInConsole =
-        isMemberExpression(callExpressionNode.callee) &&
-        ASTUtils.isIdentifier(callExpressionNode.callee.object) &&
-        callExpressionNode.callee.object.name === 'console';
+        isMemberExpression(identifierNode.parent) &&
+        identifierNode.parent.parent &&
+        isCallExpression(identifierNode.parent.parent) &&
+        isMemberExpression(identifierNode.parent.parent.callee) &&
+        ASTUtils.isIdentifier(identifierNode.parent.parent.callee.object) &&
+        identifierNode.parent.parent.callee.object.name === 'console';
 
       return (
         !isBuiltInConsole &&
--- a/lib/detect-testing-library-utils.ts
+++ b/lib/detect-testing-library-utils.ts
@@ -78,7 +78,10 @@ type IsRenderUtilFn = (node: TSESTree.Identifier) => boolean;
 type IsRenderVariableDeclaratorFn = (
   node: TSESTree.VariableDeclarator
 ) => boolean;
-type IsDebugUtilFn = (node: TSESTree.Identifier) => boolean;
+type IsDebugUtilFn = (
+  identifierNode: TSESTree.Identifier,
+  callExpressionNode: TSESTree.CallExpression
+) => boolean;
 type IsPresenceAssertFn = (node: TSESTree.MemberExpression) => boolean;
 type IsAbsenceAssertFn = (node: TSESTree.MemberExpression) => boolean;
 type CanReportErrorsFn = () => boolean;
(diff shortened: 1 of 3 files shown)


Belco90:
> I intentionally left out the following scenario
> 
> ```js
> const { debug } = console
> debug()
> ```
> 
> because it goes against the first invalid tests that against any `debug()`, it raises a flag.

I'm not sure if I understand the reason. The first invalid case is way too simple but from the Aggressive Reporting point of view is invalid. However, the case you mentioned shouldn't be reported ideally. Am I missing something else?

gndelia:
> > I intentionally left out the following scenario
> > ```js
> > const { debug } = console
> > debug()
> > ```
> > 
> > 
> > because it goes against the first invalid tests that against any `debug()`, it raises a flag.
> 
> I'm not sure if I understand the reason. The first invalid case is way too simple but from the Aggressive Reporting point of view is invalid. However, the case you mentioned shouldn't be reported ideally. Am I missing something else?

I don't know how I can differentiate

```js
const { debug } = console
debug()
```

from simply (which is reported from the aggressive reporting point of view)

```js
debug()
```

as this 👆  is included in the above one

lib/rules/no-debug.ts
```diff
@@ -50,7 +51,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
         }
         const initIdentifierNode = getDeepestIdentifierNode(node.init);
 
-        if (!initIdentifierNode) {
+        if (!initIdentifierNode || initIdentifierNode.name === 'console') {
```
MichaelDeBoey, line 54: Not sure if we should check for `console` tbh, I think this should be handled more generically, no?

Belco90:
@gndelia ok, I see what you mean now.

The difference is basically where they come from: one has been declared destructuring from `console`, the other one hasn't been declared (which is not a real-ish scenario so it should be improved to import it from some module).

Since `console` is a JS built-in module, we don't need to be that picky to check if the `debug` method has been declared from a module/object coming from Testing Library or not. Just checking that the `init.name` of the `VariableDeclarator` is `"console"` should be enough to avoid reporting in that case.

Could you include that case in the rule then, please?

lib/rules/no-debug.ts
Belco90: What do you mean by "more generically"? `console` is a built-in object so we should be safe just checking that name in particular.

Belco90, line 54: Shouldn't this actually avoid reporting the discussed case? 🤔

lib/rules/no-debug.ts
```diff
@@ -107,7 +108,11 @@ export default createTestingLibraryRule<Options, MessageIds>({
           return;
         }
 
-        const isDebugUtil = helpers.isDebugUtil(callExpressionIdentifier);
+        const isDebugUtil =
+          helpers.isDebugUtil(callExpressionIdentifier) &&
+          (!isMemberExpression(node.callee) ||
+            !ASTUtils.isIdentifier(node.callee.object) ||
+            node.callee.object.name !== 'console');
```
Belco90, line 115: Can you extract this into something more explicit? This sort of chained checks were really difficult to follow when refactoring to v4. Something like this would be enough and easier to understand:

```suggestion
        const isGlobalConsoleProperty = (isMemberExpression(node.callee) &&
            ASTUtils.isIdentifier(node.callee.object) &&
            node.callee.object.name === 'console');
            
        const isDebugUtil = helpers.isDebugUtil(callExpressionIdentifier) && !isGlobalConsoleProperty;
```

Belco90 (changes requested):

lib/rules/no-debug.ts
Belco90: It could even be done inside `isDebugUtil` since it always applies to Testing Library `debug` util.

lib/rules/no-debug.ts
Belco90: Ah, this is not enough. This would just avoid including such a reference in `isDeclaredDebugVariable`. For handling this properly then I'd suggest adding a new list `builtinConsoleNodes` or something like that, then inside `VariableDeclarator` callback if `initIdentifierNode.name === 'console'` you add that node to `builtinConsoleNodes`, and finally before reporting it you need to check `isDebugUtil || isDeclaredDebugVariable || isChainedReferenceDebug` as now AND is not coming from a node within `builtinConsoleNodes`.

I hope that's clear enough 😅

MichaelDeBoey: I mean that we should probably only check if it's coming from TestingLibrary, no?

Belco90: That's already done inside `isDebugUtil`. However, if the Aggressive Reporting is enabled, it will assume any method called `debug` has to be reported, hence the necessity of ignoring those coming from the `console` object.

Belco90: It's also valid if just right before reporting the node you get variables references from the node being analyzed with `getVariableReferences` and check if some of those references are `console` to prevent reporting.

gndelia: right, that was my reasoning; I thought only the built-in should be excluded

lib/rules/no-debug.ts
gndelia: I liked this idea - I applied it here https://github.com/testing-library/eslint-plugin-testing-library/pull/349/commits/eb5748f4a475954648335e648bf422067517ee6c

lib/rules/no-debug.ts
gndelia: I have some doubts regarding the approach, but it seems to work so here it is https://github.com/testing-library/eslint-plugin-testing-library/pull/349/commits/268a598efa31af9b2b64f80644891c1f294f887b Thanks for the help!

lib/detect-testing-library-utils.ts
```diff
@@ -580,14 +583,22 @@ export function detectTestingLibraryUtils<
...
 
-    const isDebugUtil: IsDebugUtilFn = (node) => {
-      return isTestingLibraryUtil(
-        node,
-        (identifierNodeName, originalNodeName) => {
-          return [identifierNodeName, originalNodeName]
-            .filter(Boolean)
-            .includes('debug');
-        }
+    const isDebugUtil: IsDebugUtilFn = (identifierNode, callExpressionNode) => {
```
Belco90, line 586: I'd avoid passing `callExpressionNode` as a second arg. Just use the member expression from `identifierNode.parent` and check it's defined and `isMemberExpression` to keep doing the same checks you are doing now.

Belco90 (changes requested):

lib/rules/no-debug.ts
```diff
@@ -120,7 +126,19 @@ export default createTestingLibraryRule<Options, MessageIds>({
           }
         );
 
-        if (isDebugUtil || isDeclaredDebugVariable || isChainedReferenceDebug) {
+        const isVariableFromBuiltInConsole = builtInConsoleNodes.some(
+          (variableDeclarator) => {
+            const variables = context.getDeclaredVariables(variableDeclarator);
+            return variables.some(
+              ({ name }) => name === callExpressionIdentifier.name
```
Belco90, line 133: I'm not sure if this is too brittle. Can you add an invalid test case mixing both console and Testing Library `debug`? to make sure they work fine together. Something like:

```javascript
import { debug as testingLibraryDebug } from '@testing-library/package'

test('invalid scenario: mixing console and Testing Library debug', () => {
  const { debug } = console
  debug()
  console.debug()
  testingLibraryDebug() // <-- this is the only one reported
})
```

Belco90 (changes requested):

lib/rules/no-debug.ts
gndelia: added here https://github.com/testing-library/eslint-plugin-testing-library/pull/349/commits/1778fa8ed6771e4869acbe0de254e1435cecab2a I had to add an extra check to make it work. Nice catch! I went too easy on the check 😆  👏

lib/detect-testing-library-utils.ts
gndelia: Updated here https://github.com/testing-library/eslint-plugin-testing-library/pull/349/commits/1778fa8ed6771e4869acbe0de254e1435cecab2a

lib/detect-testing-library-utils.ts
```diff
@@ -580,14 +581,25 @@ export function detectTestingLibraryUtils<
...
-        (identifierNodeName, originalNodeName) => {
-          return [identifierNodeName, originalNodeName]
-            .filter(Boolean)
-            .includes('debug');
-        }
+    const isDebugUtil: IsDebugUtilFn = (identifierNode) => {
+      const isBuiltInConsole =
+        isMemberExpression(identifierNode.parent) &&
+        identifierNode.parent.parent &&
+        isCallExpression(identifierNode.parent.parent) &&
```
Belco90, line 588: We aren't interested in checking if the received identifier comes from a Call Expression, that's up to whichever code is calling `isDebugUtil`. This should only check if the identifier itself represents a debug, discarding those belonging to a Member Expression with "console object. Can you remove this check please?

lib/detect-testing-library-utils.ts
```diff
@@ -580,14 +581,25 @@ export function detectTestingLibraryUtils<
...
+        isMemberExpression(identifierNode.parent.parent.callee) &&
+        ASTUtils.isIdentifier(identifierNode.parent.parent.callee.object) &&
+        identifierNode.parent.parent.callee.object.name === 'console';
```
Belco90, line 591: This is redundant, isn't it? `identifierNode.parent.parent.callee.object.name` is just `identifierNode.parent.object.name` right? Can you simplify this check please?

Belco90 (changes requested):
Nice, almost there! Just a couple of small tweaks and this is good to go 🚀

lib/detect-testing-library-utils.ts
gndelia: I actually needed to know thee parent is a CallExpression because that's the only way to differentiate `console.debug()` from plain `debug()`. For the first case, I know I can ignore it. For the 2nd one, I need further info, which are the following lines of code

Belco90 (approved):
Awesome 👏

Issue #351, opened 2021-04-22
no-wait-for-side-effects does not catch returned side effect


zaicevas:
This yields an error:

```
    await waitFor(() => {
      userEvent.click(screen.getByText('Continue'))
    })
```

This does not:

```
    await waitFor(() => userEvent.click(screen.getByText('Continue')))
```

Is that expected behaviour?

Belco90:
Hey @zaicevas! It's not an expected behavior indeed. It seems the plugin is only looking for arrow functions with a body but ignoring those with an implicit return.

PR #352, opened 2021-04-23
fix(no-wait-for-side-effects): catch implicit arrow function return

zaicevas:
Closes https://github.com/testing-library/eslint-plugin-testing-library/issues/351

### DIFF
--- a/lib/rules/no-wait-for-side-effects.ts
+++ b/lib/rules/no-wait-for-side-effects.ts
@@ -76,8 +76,44 @@ export default createTestingLibraryRule<Options, MessageIds>({
       }
     }
 
+    function reportImplicitReturnSideEffect(node: TSESTree.CallExpression) {
+      if (!node.parent) {
+        return;
+      }
+      const callExpressionNode = node.parent.parent as TSESTree.CallExpression;
+      const callExpressionIdentifier = getPropertyIdentifierNode(
+        callExpressionNode
+      );
+
+      if (!callExpressionIdentifier) {
+        return;
+      }
+
+      if (!helpers.isAsyncUtil(callExpressionIdentifier, ['waitFor'])) {
+        return;
+      }
+
+      const expressionIdentifier = getPropertyIdentifierNode(node.callee);
+      if (!expressionIdentifier) {
+        return false;
+      }
+
+      if (
+        !helpers.isFireEventUtil(expressionIdentifier) &&
+        !helpers.isUserEventUtil(expressionIdentifier)
+      ) {
+        return;
+      }
+
+      context.report({
+        node,
+        messageId: 'noSideEffectsWaitFor',
+      });
+    }
+
     return {
       'CallExpression > ArrowFunctionExpression > BlockStatement': reportSideEffects,
+      'CallExpression > ArrowFunctionExpression > CallExpression': reportImplicitReturnSideEffect,
       'CallExpression > FunctionExpression > BlockStatement': reportSideEffects,
     };
   },
--- a/lib/rules/no-wait-for-side-effects.ts
+++ b/lib/rules/no-wait-for-side-effects.ts
@@ -87,7 +87,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
     }
 
     function reportImplicitReturnSideEffect(node: TSESTree.CallExpression) {
-      if (!isCallerWaitFor) {
+      if (!isCallerWaitFor(node)) {
         return;
       }
 
--- a/lib/rules/no-wait-for-side-effects.ts
+++ b/lib/rules/no-wait-for-side-effects.ts
@@ -93,7 +93,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
 
       const expressionIdentifier = getPropertyIdentifierNode(node.callee);
       if (!expressionIdentifier) {
-        return false;
+        return;
       }
 
       if (
--- a/lib/rules/no-wait-for-side-effects.ts
+++ b/lib/rules/no-wait-for-side-effects.ts
@@ -28,7 +28,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
   create: function (context, _, helpers) {
     function isCallerWaitFor(
       node: TSESTree.BlockStatement | TSESTree.CallExpression
-    ) {
+    ): boolean {
       if (!node.parent) {
         return false;
       }
@@ -37,15 +37,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
         callExpressionNode
       );
 
-      if (!callExpressionIdentifier) {
-        return false;
-      }
-
-      if (!helpers.isAsyncUtil(callExpressionIdentifier, ['waitFor'])) {
-        return false;
-      }
-
-      return true;
+      return (
+        !!callExpressionIdentifier &&
+        helpers.isAsyncUtil(callExpressionIdentifier, ['waitFor'])
+      );
     }
 
     function getSideEffectNodes(
--- a/lib/rules/no-wait-for-side-effects.ts
+++ b/lib/rules/no-wait-for-side-effects.ts
@@ -26,6 +26,28 @@ export default createTestingLibraryRule<Options, MessageIds>({
   },
   defaultOptions: [],
   create: function (context, _, helpers) {
+    function isCallerWaitFor(
+      node: TSESTree.BlockStatement | TSESTree.CallExpression
+    ) {
+      if (!node.parent) {
+        return false;
+      }
+      const callExpressionNode = node.parent.parent as TSESTree.CallExpression;
+      const callExpressionIdentifier = getPropertyIdentifierNode(
+        callExpressionNode
+      );
+
+      if (!callExpressionIdentifier) {
+        return false;
+      }
+
+      if (!helpers.isAsyncUtil(callExpressionIdentifier, ['waitFor'])) {
+        return false;
+      }
+
+      return true;
+    }
+
     function getSideEffectNodes(
       body: TSESTree.Node[]
     ): TSESTree.ExpressionStatement[] {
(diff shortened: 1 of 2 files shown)


lib/rules/no-wait-for-side-effects.ts
```diff
@@ -76,8 +86,32 @@ export default createTestingLibraryRule<Options, MessageIds>({
       }
     }
 
+    function reportImplicitReturnSideEffect(node: TSESTree.CallExpression) {
+      if (!isCallerWaitFor) {
```
gndelia, line 90: I think this is checking for the existance of the function, instead of the result of calling it

```suggestion
      if (!isCallerWaitFor(node)) {
```

gndelia (changes requested):
Thanks for your first PR ❤️

lib/rules/no-wait-for-side-effects.ts
zaicevas: Totally 😅

zaicevas: Fixed in 71c97fc

tests/lib/rules/no-wait-for-side-effects.test.ts
```diff
@@ -163,7 +163,14 @@ ruleTester.run(RULE_NAME, rule, {
     // fireEvent
     {
```
Belco90, line 164: Can you add some valid test cases too? So we check `waitFor` with an arrow function returning a call expression that shouldn't be reported. A couple of cases like these should be fine:

```javascript
`await waitFor(() => somethingElse());`
```

```javascript
{
  settings: { 'testing-library/utils-module': 'test-utils' },
  code: `
    import { waitFor } from 'somewhere-else';
    await waitFor(() => fireEvent.keyDown(input, {key: 'ArrowDown'}))
  `,
},
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -26,6 +26,28 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+
+      if (!callExpressionIdentifier) {
+        return false;
+      }
+
+      if (!helpers.isAsyncUtil(callExpressionIdentifier, ['waitFor'])) {
+        return false;
+      }
+
+      return true;
```
MichaelDeBoey, line 48: No need for early return anymore in this part

```suggestion
      return Boolean(callExpressionIdentifier) &&
        helpers.isAsyncUtil(callExpressionIdentifier, ['waitFor']);
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -26,6 +26,28 @@ export default createTestingLibraryRule<Options, MessageIds>({
   },
   defaultOptions: [],
   create: function (context, _, helpers) {
+    function isCallerWaitFor(
+      node: TSESTree.BlockStatement | TSESTree.CallExpression
+    ) {
```
Belco90, line 31: Can you specify here that this function returns a `boolean` please?

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -76,8 +86,32 @@ export default createTestingLibraryRule<Options, MessageIds>({
       }
     }
 
+    function reportImplicitReturnSideEffect(node: TSESTree.CallExpression) {
+      if (!isCallerWaitFor(node)) {
+        return;
```
MichaelDeBoey, line 91: What's the difference between breaking out of the function and returning `false`? 🤔

MichaelDeBoey (changes requested):

Belco90 (changes requested):
Hey, @zaicevas thanks for your first PR! It looks great, just a couple of small tweaks, and should be ready to go.

May I ask how did you feel working within the codebase? We have put a lot of effort into making it more maintainable not only for usual contributors but also for outside ones, so any feedback you have around this would be really appreciated.

lib/rules/no-wait-for-side-effects.ts
zaicevas: image

Because of  ⬆️ , I'd vote for this:

```
      return (
        !!callExpressionIdentifier &&
        helpers.isAsyncUtil(callExpressionIdentifier, ['waitFor'])
      );
```

What do you think?

lib/rules/no-wait-for-side-effects.ts
Belco90: I'd say there is no difference? It's just a way to stop the lining as soon as you know it won't be reported, but I guess we could stick to just `return;` to make it more consistent.

zaicevas: That's unnecessary. Leftover from my part 😶

lib/rules/no-wait-for-side-effects.ts
gndelia: shouldn't the compiler be smart enough to infer that itself? I usually tend to avoid adding the return types for inline functions.

lib/rules/no-wait-for-side-effects.ts
MichaelDeBoey: Let's use double negation for now then

Belco90: I think we can leave it as it was since 1) it works and 2) it's an inherited code, you just moved it somewhere else.

Belco90: @zaicevas oh wow, if you cast it with `Boolean` TS complains, but if you cast with `!!` TS doesn't complain? :O

lib/rules/no-wait-for-side-effects.ts
zaicevas: Fixed in b08d377

lib/rules/no-wait-for-side-effects.ts
Belco90: I find that useful when you are not really sure what type you are returning (like a React component), but here are pretty sure we want to return a boolean, so specifying the type would prevent us from returning something else. I don't know what you think, and I've probably missed few like this in the past myself 😅

lib/rules/no-wait-for-side-effects.ts
zaicevas: Yep 😄

lib/rules/no-wait-for-side-effects.ts
gndelia: I'm open to either option 🤷

lib/rules/no-wait-for-side-effects.ts
Belco90: Oh wow, TIL https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAcwE4FN2zMgFGAQwFt0AuRAZylRhwEpyqadEBvAKES8QyhFSQADABLoANmLiIAJK0IkAvoPYL27MVkRg4AOWJlK1WskQAfLSAmIAvBYkBuNQCE4cDQTD5d+uogBkfigYWMZeeiR07ACEUWE+-oFomNh42uHodEA

tests/lib/rules/no-wait-for-side-effects.test.ts
zaicevas: From my understanding, this case should pass as invalid. It does not, because `isUserEventUtil()` returns false. Am I missing something? 🤔 

```
    {
      settings: { 'testing-library/utils-module': '~/test-utils' },
      code: `
        import { waitFor, userEvent } from '~/test-utils';
        await waitFor(() => userEvent.click(button))
      `,
      errors: [{ line: 4, column: 29, messageId: 'noSideEffectsWaitFor' }],
    },
```

lib/rules/no-wait-for-side-effects.ts
zaicevas: I added explicit return type in 8e6044a. I think that's useful if you want to make sure the API does not change in the long term

lib/rules/no-wait-for-side-effects.ts
zaicevas: Added double negation 8e6044a

tests/lib/rules/no-wait-for-side-effects.test.ts
Belco90: `userEvent` comes from a different package so it's not affected by the Aggressive Reporting mechanism, so it would be only reported if imported from `@testing-library/user-event`

MichaelDeBoey (approved):

tests/lib/rules/no-wait-for-side-effects.test.ts
zaicevas: Added more test cases in e991eed

Belco90 (approved):
Awesome! Nice work @zaicevas 🚀

MichaelDeBoey (approved):

Belco90:
@all-contributors please add @zaicevas for bug, code and test

zaicevas:
> Hey, @zaicevas thanks for your first PR! It looks great, just a couple of small tweaks, and should be ready to go.
> 
> May I ask how did you feel working within the codebase? We have put a lot of effort into making it more maintainable not only for usual contributors but also for outside ones, so any feedback you have around this would be really appreciated.

The development experience is great. I enjoyed it and it's motivating to contribute more. Tests are fast, CI/CD is convenient, so it looks like it's easy to quickly make an impact. I've never coded anything eslint related, yet it didn't take me more than a couple of hours to throw this together.

For the first 30 minutes I had no idea where to start, until I've realized it's all based on AST and found https://astexplorer.net/ 
After that it all started to make sense and the code proved to be very clear & extendable. Perhaps it could be useful mentioning this or other similar tools in https://github.com/testing-library/eslint-plugin-testing-library/blob/main/CONTRIBUTING.md

Another that I think could be of use in `CONTRIBUTING.md` is a section on how to try out changes locally (besides tests). I suppose symlinking with `yarn link` or `npm link` is relatively simple and could work here.

> Thanks for your first PR ❤️
> Awesome! Nice work @zaicevas 🚀

More than happy to help ☺️

Belco90:
@zaicevas thanks for your feedback, really useful! It's definitely a good idea to include some AST info and AST Explorer link to our CONTRIBUTING: the explorer is a must when dealing with AST, isn't it?

Also interesting to add a section for trying changes locally (actually someone asked the very same thing the other day in Testing Library Discord).

I'll create an issue for this.

zaicevas:
>  (actually someone asked the very same thing the other day in Testing Library Discord).

That was me, by the way :trollface:

Issue #294, opened 2021-03-20
New rule: no-log-testing-playground-url


MichaelDeBoey:
Just like `no-debug` but for `screen.logTestingPlaygroundURL()`

Reference: https://github.com/testing-library/dom-testing-library/pull/781

Belco90:
It sounds like a good idea, and would be really easy to implement just mirroring the same checks from `no-debug`. However, we have to put it on hold until v4 gets released or it will take forever!

MichaelDeBoey:
I think we can indeed extract a lot of the `no-debug` code in order to implement this one

Belco90:
Rethinking this suggestion: I think `no-debug` should report `logTestingPlaygroundURL` usages. Both `debug` and `logTestingPlaygroundURL` are debugging utils, so `no-debug` should take care of them. Even `prettyDOM` and `logRoles` probably.

MichaelDeBoey:
Maybe we should rename `no-debug` to something like `no-debugging-utils`?

We can create that new rule in a minor release and let that handle all of them (`debug`/`screen.debug`, `screen.logTestingPlaygroundURL`, `prettyDOM`, `logRoles`, `logDOM`, `prettyFormat`, ...).

We can deprecate the `no-debug` rule for now and maybe throw a warning or so to say that using the new `no-debugging-utils` should be used and remove it in v5?
Or we can update `no-debug` and rename it to a better suiting name in v5 too if we want?

MichaelDeBoey:
On the other hand, maybe some people will want to enable this behavior for some utils, but not for others.
So having a setting for which one to enable can be helpful too I think

Belco90:
- I would implement everything under `no-debug` and rename it in a future v5 if necessary.
- The rule should accept an option to decide which debug utils are allowed like [`no-console` ESLint one](https://eslint.org/docs/rules/no-console#disallow-the-use-of-console-no-console)

MichaelDeBoey:
Sounds like a solid plan @Belco90! 👊

MichaelDeBoey:
Since it's a breaking change to error on these new functions, we should set the default of the `allow` option to be `['debug', 'screen.debug']`

Only in v5 we can update it to default to all functions then. 🤔

Belco90:
[...]

PR #357, opened 2021-04-26
feat(no-debug): support `utilNames` option

G-Rath:
Fixes half of #294

Should be followed up with creating a new rule called `no-debugging-utils` that this one is deprecated for.

I've gone with the very average "utilNames" name for the option - very happy to change it.
I've also left `isDebugUtil` in `detect-testing-library-utils` even though it looks like its now unused, as I don't know if you're publishing or exposing it for third-parties to use? (so removing it might be breaking?)

I've done this as a first pass implementation, so happy to make changes to docs, names, etc as desired :)

(this is also why I've not refactored the `noDebug` message to be `noDebugUtils` etc, in case this PR is going in the wrong direction)


Belco90:
Oh wow, thank you very much @G-Rath! I'll take a deeper look during the week. The direction of the PR seems ok in general, but there are a couple of things I'd like to discuss (like `utilNames` option vs dedicated option per debug util name, or reusing `isDebugUtil` with a 2nd param `validNames` like `isAsyncUtil` accepting `Array<'debug' | 'logTestingPlaygroundURL'>`

tests/lib/rules/no-debug.test.ts
```diff
@@ -87,6 +87,19 @@ ruleTester.run(RULE_NAME, rule, {
...
+        import { screen } from '@testing-library/dom'
+        screen.logTestingPlaygroundURL()
+      `,
+    },
+    {
+      code: `
+        import { screen } from '@testing-library/dom'
+        screen.debug()
+      `,
+      options: [{ utilNames: ['anotherUtil'] }],
```
Belco90, line 101: This test should not be correct after you set the option type properly, so it will have to be updated.

tests/lib/rules/no-debug.test.ts
```diff
@@ -87,6 +87,19 @@ ruleTester.run(RULE_NAME, rule, {
         screen.debug
       `,
     },
+    {
+      code: `
+        import { screen } from '@testing-library/dom'
+        screen.logTestingPlaygroundURL()
+      `,
```
Belco90, line 94: This test case will be invalid when you opt-in `logTestingPlaygroundURL` by default.

docs/rules/no-debug.md
```diff
@@ -28,6 +28,12 @@ const { screen } = require('@testing-library/react');
 screen.debug();
 ```
 
+If you want to disallow the use of other debugging functions, you can configure what names this rule checks for with the `utilNames` option:
```
Belco90, line 31: The description of this rule must be updated to mention which debugging utils are reported by default.

lib/detect-testing-library-utils.ts
```diff
@@ -580,6 +585,26 @@ export function detectTestingLibraryUtils<
       return isRenderUtil(initIdentifierNode);
     };
 
+    const isOneOfDebugUtils: IsOneOfDebugUtils = (
```
Belco90, line 588: Instead of this new method, I'd update the existing `isDebugUtils` to mirror `isAsyncUtil` so its type would be
```typescript
(identifierNode: TSESTree.Identifier, validNames: Array<'debug' | 'logTestingPlaygroundURL'>) => boolean;
```

It would set `validNames` to `['debug', 'logTestingPlaygroundURL']` by default in the function.

Belco90 (changes requested):
Review finished! I requested few changes so:
-  we reuse `isDebugUtil` to be able to detect `logTestingPlaygroundURL`
- we opt-in reporting `logTestingPlaygroundURL` by default

Thanks!

lib/rules/no-debug.ts
```diff
@@ -13,7 +13,7 @@ import { ASTUtils, TSESTree } from '@typescript-eslint/experimental-utils';
 
 export const RULE_NAME = 'no-debug';
 export type MessageIds = 'noDebug';
-type Options = [];
+type Options = [{ utilNames: string[] }];
```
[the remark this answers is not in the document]
G-Rath: I was under the impression there were more functions that users might like to ignore and that you'd not want to maintain a list of all them in the code; given that you do, I think it might be nicer to use an object rather than array, since the latter gives the impression (imo) that it accepts arbitrary strings.

i.e:

```
interface DebugUtilsToCheckFor {
  debug?: boolean;
  logTestingPlaygroundURL?: boolean;
}

type Options = [{ utilsToCheckFor: DebugUtilsToCheckFor }]
```

It probably has a more sensible default behaviour too, since with an array as soon as you specify that option all the defaults go out the door unless the user maintains the list themselves; whereas with an object, users can toggle specific utils without having to care about the others.

(all of this a bit moot when it's just two names, but tbh if more exist currently we might as well include them all right now).

I'm not too fussed personally, but just know from experience how much of a pain these sorts of things can be to change afterwards 😅

lib/rules/no-debug.ts
```diff
@@ -27,11 +27,22 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+            type: 'array',
+            items: { type: 'string' },
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
   },
-  defaultOptions: [],
+  defaultOptions: [{ utilNames: ['debug'] }],
```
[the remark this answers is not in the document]
G-Rath: Find by me; fwiw we'd consider this sort of change breaking over at `eslint-plugin-jest` (*but* it wouldn't be breaking to do this in the renamed version of the rule, which you could ship next to this one).

lib/rules/no-debug.ts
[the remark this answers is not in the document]
Belco90: To be honest I'm happy with this implementation too, just wanted to type possibilities for the final user. Although I prefer the list in case we need to include more debugging names (which might happen to cover dom-testing-library ones).

lib/rules/no-debug.ts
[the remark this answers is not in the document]
Belco90: We are [still debating around SemVer Policy](https://github.com/testing-library/eslint-plugin-testing-library/discussions/343). Your input would be appreciated!

lib/utils/index.ts
```diff
@@ -66,6 +66,11 @@ const ASYNC_UTILS = [
   'waitForDomChange',
 ] as const;
 
+const DEBUG_UTILS = [
+  'debug', //
```
Belco90, line 70: There is an empty comment at the end of the line, I guess is a leftover 😅

tests/lib/rules/no-debug.test.ts
```diff
@@ -419,6 +426,20 @@ ruleTester.run(RULE_NAME, rule, {
...
+      options: [{ utilNames: ['logTestingPlaygroundURL'] }],
+      errors: [
+        {
+          line: 3,
+          column: 16,
+          messageId: 'noDebug',
+        },
+      ],
+    },
```
Belco90, line 442: Can you add another test exactly like this one but without setting `options` so we test `logTestingPlaygroundURL` is reported by default?

Belco90 (changes requested):
Looking good! Just a couple of small tweaks and should be ready to go. Thanks!

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -606,9 +613,11 @@ export function detectTestingLibraryUtils<
...
           identifierNode,
           (identifierNodeName, originalNodeName) => {
-            return [identifierNodeName, originalNodeName]
-              .filter(Boolean)
-              .includes('debug');
+            return (
+              (validNames as string[]).includes(identifierNodeName) ||
+              (!!originalNodeName &&
+                (validNames as string[]).includes(originalNodeName))
+            );
```
MichaelDeBoey, line 620:
```suggestion
            return [identifierNodeName, originalNodeName]
              .filter((nodeName): nodeName is string => Boolean(nodeName))
              .some((nodeName) => (validNames as string[]).includes(nodeName));
```

lib/rules/no-debug.ts
```diff
@@ -8,12 +8,13 @@ import {
...
   isProperty,
 } from '../node-utils';
+import { DEBUG_UTILS } from '../utils';
 import { createTestingLibraryRule } from '../create-testing-library-rule';
 import { ASTUtils, TSESTree } from '@typescript-eslint/experimental-utils';
 
 export const RULE_NAME = 'no-debug';
 export type MessageIds = 'noDebug';
-type Options = [];
+type Options = [{ utilNames: Array<typeof DEBUG_UTILS[number]> }];
```
MichaelDeBoey, line 17: `utilNames` is optional

```suggestion
type Options = [{ utilNames?: Array<typeof DEBUG_UTILS[number]> }];
```

lib/rules/no-debug.ts
```diff
@@ -32,11 +33,25 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+              enum: DEBUG_UTILS,
+            },
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
   },
-  defaultOptions: [],
+  defaultOptions: [{ utilNames: ['debug', 'logTestingPlaygroundURL'] }],
```
MichaelDeBoey, line 52: Use `DEBUG_UTILS` by default

```suggestion
  defaultOptions: [{ utilNames: [...DEBUG_UTILS] }],
```

lib/utils/index.ts
MichaelDeBoey, line 70:
```suggestion
  'debug',
```

MichaelDeBoey (changes requested):
This one's looking good, thanks @G-Rath! 👊

Added a couple of tweaks that you need to do, but other than that this one looks good for me.

I'm still not sure about the array vs. object (I can see the benefit of both of them) and the fact if we should make it breaking or not in this case though. 🤔

lib/rules/no-debug.ts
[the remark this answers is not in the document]
G-Rath: I'm not sure I follow you - in both cases you have the types for the final user and it's equal amount of work to add more debugging names.

But I'm happy to go with whichever you want :)

lib/utils/index.ts
G-Rath: No actually this is a marker for `prettier` to have it put the array items on their own lines even if it can fit on a single line with 80 chars.

I find that useful for lists like these where it's known that more items will be added later, since it reduces the diff.

Happy to remove this if  you want it formatted to a single line, or otherwise happy to add a few more utils (which was my original plan).

lib/create-testing-library-rule/detect-testing-library-utils.ts
G-Rath: I'm not sure if I'd call that simpler; but I'd prefer to keep it how it is currently because it matches how the same conditional is laid out in other utils (i.e `isAsyncUtil`)

lib/rules/no-debug.ts
G-Rath: Oh yeah, nice catch!

lib/rules/no-debug.ts
G-Rath: The issue with this is that any change to `DEBUG_UTILS` will implicitly affect this rule, which semver thoughts aside, is a very easy way to introduce an unintentional change.

If you're happy with that risk I don't mind changing this.

G-Rath:
@Belco90 I'm keen to add a few more debug utils to the list of what's supported (but keeping the current defaults, unless you'd like them added there too), if that's alright with you?

Specifically, the ones @MichaelDeBoey mentioned [here](https://github.com/testing-library/eslint-plugin-testing-library/issues/294#issuecomment-817709073).

lib/rules/no-debug.ts
MichaelDeBoey: Let's keep the array as it was

lib/create-testing-library-rule/detect-testing-library-utils.ts
MichaelDeBoey: Let's see what @Belco90 says, but I think we should update `isAsyncUtil` too (in a separate PR) as this is much cleaner an easier to understand imo

MichaelDeBoey:
@G-Rath I would be OK with adding the other ones I mentioned in https://github.com/testing-library/eslint-plugin-testing-library/issues/294#issuecomment-817709073.

lib/create-testing-library-rule/detect-testing-library-utils.ts
G-Rath: Totally! in fact, you could even deduplicate the code with a little local function that accepts the node + an array of strings as names - something like this:

```
    const isTestingLibraryUtilOfName = (
      node: TSESTree.Identifier,
      validNames: string[]
    ): boolean => {
      return isTestingLibraryUtil(
        node,
        (identifierNodeName, originalNodeName) => {
          return (
            validNames.includes(identifierNodeName) ||
            (!!originalNodeName && validNames.includes(originalNodeName))
          );
        }
      );
    };
```

MichaelDeBoey: Extracting it in that new PR sounds indeed a better idea too

lib/utils/index.ts
Belco90: Oh wow, I didn't know this way of forcing prettier to format array elements per line! It's still weird tho, but since we are gonna add more utils, I guess it will disappear.

Belco90:
> @Belco90 I'm keen to add a few more debug utils to the list of what's supported (but keeping the current defaults, unless you'd like them added there too), if that's alright with you?
> 
> Specifically, the ones @MichaelDeBoey mentioned [here](https://github.com/testing-library/eslint-plugin-testing-library/issues/294#issuecomment-817709073).

By all means! Also, regarding the way of declaring the options as array vs boolean properties, I think you are right and the latter is better for the final user without implying more work to maintain, so I'd be happy to go ahead with that too.

Sorry for being late confirming this, and thanks for your patience :)

lib/rules/no-debug.ts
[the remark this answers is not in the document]
Belco90: You were right: defining the utils to be checked as an object is way better than as an array.

docs/rules/no-debug.md
```diff
@@ -28,6 +30,12 @@ const { screen } = require('@testing-library/react');
 screen.debug();
 ```
 
+If you want to allow the use of some debugging functions, you can configure what names this rule checks for with the `utilNames` option:
+
+```
+   "testing-library/no-debug": ["error", {"utilNames": ["debug"}],
```
Belco90, line 36: This option must be updated.

Belco90 (changes requested):
Looking amazing! The last necessary change is to update the rule doc accordingly with the final implementation and should be ready to go. Thanks!

G-Rath:
@Belco90 have updated the docs.

on an aside, I wrote a [script for automatically keeping rule docs](https://github.com/jest-community/eslint-plugin-jest/blob/main/tools/regenerate-docs.ts) consistent across rule tables in the README, the header in each rules doc, and the `meta.docs.description` property for each rule - you're welcome to take it if you'd find that useful to have here :) (I'm also happy to help set it up too if you like).

docs/rules/no-debug.md
```diff
@@ -28,6 +30,12 @@ const { screen } = require('@testing-library/react');
 screen.debug();
 ```
 
+If you want to allow the use of some debugging functions, you can configure what names this rule checks for with the `utilsToCheckFor` option:
```
Belco90, line 33: Could you include here the full list of debugging utils that can be switched on/off, please? Basically those from `DEBUG_UTILS`.

Belco90 (changes requested):
> on an aside, I wrote a script for automatically keeping rule docs consistent across rule tables in the README, the header in each rules doc, and the meta.docs.description property for each rule - you're welcome to take it if you'd find that useful to have here :) (I'm also happy to help set it up too if you like).

That's awesome! We are actually doing some improvements around this, so we will get inspiration from your script for sure. Thanks!

docs/rules/no-debug.md
G-Rath: also it would be good if I can write valid json 🤦

Belco90 (approved):
Fantastic! Thank you very much for your contribution, and sorry for the initial back and forth.

G-Rath:
No problem - happy to help a fellow eslint plugin 😄

Belco90:
@all-contributors please add @G-Rath for code, docs and test

PR #358, opened 2021-04-26
refactor: automatically generate all configs

### LINKED ISSUE #187: Autogenerate configs from rule files
See extended discussion in #182

MichaelDeBoey:
I first tried to re-use `recommended` and let that be an object that has the supported testing frameworks as keys and as value either just `'error'`, `'warn'` or `false` **or** a tuple that has one of these values as first value and the possible config as second.
This way each plugin can have totally different config options (as seen in `no-dom-import`.

When automatically generating all configs (see 252b0b9), I got empty `rules` objects as we are now hardcoding `recommended` to `false` in `createTestingLibraryRule` (even though the input has it in the new way).
That's why I decided to use a new `recommendedConfig` key in the meta docs.

Everything is working as it should be now, except for `consistent-data-testid`.  
Since we're not using `createTestingLibraryRule` there, TS is complaining when adding the new `recommendedConfig` key in the meta docs.  
I've seen why we're not using `createTestingLibraryRule`, but can we switch to it either way or should we look into a totally different solution?

This PR is only generating configs.
I'll create follow-up PRs for generating docs tables in README and other docs related tables in the rules specific docs once this one's merged.

---

Closes #187


lib/create-testing-library-rule.ts
```diff
@@ -1,34 +1,35 @@
...
   return ESLintUtils.RuleCreator(getDocsUrl)({
     ...remainingConfig,
     create: detectTestingLibraryUtils<TOptions, TMessageIds, TRuleListener>(
       create
     ),
+    meta: {
+      ...meta,
+      docs: {
+        ...meta.docs,
+        recommended: false,
```
Belco90, line 31: Could you add a comment here indicating why we pass this always as `false` please?

lib/rules/await-fire-event.ts
```diff
@@ -19,7 +19,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
     docs: {
       description: 'Enforce promises from fire event methods to be handled',
       category: 'Best Practices',
-      recommended: false,
+      recommendedConfig: {
+        dom: false,
+        angular: false,
+        react: false,
```
Belco90, line 25: I was thinking to have properties within this object optional so this one could be expressed as just:
```javascript
recommendedConfig: {
  vue: 'error'
}
```
skipping all false ones. But I actually prefer having to specify all of them, it's more explicit 👍

lib/utils/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  | TSESLint.RuleMetaDataDocs['recommended']
+  | [TSESLint.RuleMetaDataDocs['recommended'], ...TOptions];
+
+// These 2 types are copied from @typescript-eslint/experimental-utils' CreateRuleMeta
+// and modified to our needs
+type TestingLibraryRuleMetaDocs<TOptions extends readonly unknown[]> = Omit<
+  TSESLint.RuleMetaDataDocs,
+  'recommended' | 'url'
+> & {
+  recommendedConfig: Record<
```
Belco90, line 13: Could you add a jsdoc to this property like [`@typescript-eslint/experimental-utils` does when typing `RuleMetaDataDocs`](https://github.com/typescript-eslint/typescript-eslint/blob/1c1b572c3000d72cfe665b7afbada0ec415e7855/packages/experimental-utils/src/ts-eslint/Rule.ts#L26) please?

package.json
```diff
@@ -27,10 +27,11 @@
   "scripts": {
     "build": "tsc",
     "postbuild": "cpy README.md ./dist && cpy package.json ./dist && cpy LICENSE ./dist",
-    "lint": "eslint . --max-warnings 0 --ext .js,.ts",
-    "lint:fix": "npm run lint -- --fix",
     "format": "prettier --write README.md \"{lib,docs,tests}/**/*.{js,ts,md}\"",
     "format:check": "prettier --check README.md \"{lib,docs,tests}/**/*.{js,json,yml,ts,md}\"",
+    "generate:configs": "ts-node tools/generate-configs",
```
Belco90, line 32: We need to generate this at some point in the CI and CD. I guess it makes sense to automatically run in a new `generate_docs` step within `pipeline` jobs, before the `release` one. Inside this job, we only run this `generate:configs` for now, but we can include more steps when we have more doc stuff automated. What do you think?

Belco90 (changes requested):
Oh, I forgot that `consistent-data-testid` wasn't using our custom rule creator. When I refactored that rule for v4 I originally added an option to skip the reporting checks. You can see that [in the original PR](https://github.com/testing-library/eslint-plugin-testing-library/pull/248). I added a new `detectionOptions` property to `createTestingLibraryRule` so we could pass `skipRuleReportingCheck` to avoid preventing rules to be reported with the mechanism at the very bottom of `detect-testing-library-utils.ts` (this wasn't implemented).

I eventually reverted the mechanism since I didn't get any advantage using our custom rule creator in that rule. I think what we can do is to get that idea back, tweaking it as we need and implementing the actual skip in `detect-testing-library-utils.ts` (I can give you more details if you are not really sure about this).

Apart from that, the PR looks really nice! I've left a couple of comments for some small tweaks. We also need to mention this `recommendedConfig` property in `CONTRIBUTING.md`

lib/rules/await-fire-event.ts
MichaelDeBoey: I like the explicitness too, so that's why did it this way

MichaelDeBoey:
@Belco90 I'll create a new PR where I refactor `consistent-data-testid` to work with `createTestingLibraryRule` too, so this PR doesn't become too big.

Once that's merged, I can rebase this one and fix all the comments.

timdeschryver (commented):
This is looking great! ✨

package.json
timdeschryver: FWIW, I added this to the commit hook in the NgRx ESLint plugin.
Just to keep the configs up-to-date in the repository.

Belco90: That's a good point, but then this is gonna be really easy to forget? I'm not sure where to include it, to be honest.

Belco90: I've been thinking about this and really don't know what would be the best approach. When I suggested generating this in CI/CD I was thinking the final config wasn't in the repo, but of course, it is so we should avoid writing it in CI/CD. On the other hand, we need to automate it somehow to prevent releasing a new version without updating the shareable configs properly.

I think what we can do is something similar to what to do to make sure exported rules are the expected ones etc: Add a test to `index.test.ts` which manually builds each shareable config from rules and compares them against the current shareable config generated in the codebase. This way, we avoid writing this on CI/CD but we prevent forgetting about updating those presets. What do you think? /cc @MichaelDeBoey

Belco90 (changes requested):
Fantastic work @MichaelDeBoey 💯 

The last thing would be referencing `recommendedConfig` in `CONTRIBUTING.md` file, can you add some doc about it please?

tools/generate-configs/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { LinterConfig, writeConfig } from './utils';
+
+const RULE_NAME_PREFIX = 'testing-library/';
+
+const getRecommendedRulesForTestingFramework = (
+  framework: SupportedTestingFramework
+): LinterConfigRules =>
+  Object.entries(rules)
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
timdeschryver, line 16: Nit, I prefer to tweak the config to ignore variables with the `_` prefix:

```
    "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }]
```

timdeschryver (dismissed):

Belco90 (approved):

nickmccurdy:
Personally I think it's simpler to dynamically generate objects in the package, and it avoids issues with the generated files getting out of date in version control. Would you be interested in an additional PR for that?

Belco90:
@nickmccurdy that's how I assumed it would work at the beginning, but I'm happy with the current implementation so we can include the generated files in the repo and check the expected config in CI. Thanks anyway!

MichaelDeBoey:
@nickmccurdy That's how I wanted to implement it at first.
But thinking about this, I didn't find a way to check out each config file completely (which is a use case I have often myself) so that's why I decided to go with the same way @typescript-eslint is going.

nickmccurdy:
You can use `eslint --print-config FILE` for that. I personally don't think showing the configs in the source code additionally is worth the burden of having to manually commit updates to them.

PR #361, opened 2021-04-26
chore: prettify codebase

MichaelDeBoey:
My IDE keeps complaining about this file (probably because I have enabled a Prettier plugin)

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -4,5 +4,5 @@
     "prettier --write",
     "jest --findRelatedTests"
   ],
-  "*.md": ["prettier --write"]
+  "*.{md,yml}": ["prettier --write"]
 }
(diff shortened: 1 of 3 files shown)


.lintstagedrc
```diff
@@ -4,5 +4,5 @@
     "prettier --write",
     "jest --findRelatedTests"
   ],
-  "*.md": ["prettier --write"]
+  "*.{md,yml}": ["prettier --write"]
```
Belco90, line 7: We need to make this formatting and the ones within `package.json` `scripts` consistent since the latter is only formatting inside docs, lib, and tests folders.

Belco90 (changes requested):

.lintstagedrc
Belco90, line 7: I think we need to include `json` here right? It's the only file type missing from our scripts.

Belco90 (changes requested):

Belco90 (approved):

PR #362, opened 2021-04-27
refactor(consistent-data-testid): use createTestingLibraryRule

MichaelDeBoey:
I didn't add tests for the new `skipRuleReportingCheck` detection option (yet) as I don't know how to test that one.

@Belco90 could you point me in the right direction of how I would add tests for it?

---

This PR does some ground work for #358


lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -0,0 +1,917 @@
...
+export type DetectionOptions = {
+  /**
+   * If true, force `detectTestingLibraryUtils` to skip `canReportErrors`
+   * so it doesn't opt-out rule listener.
+   *
+   * Useful when some rule apply to files other than testing ones
+   * (e.g. `consistent-data-testid`)
+   */
+  skipRuleReportingCheck: boolean;
+};
```
MichaelDeBoey, line 136: Part 1

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -0,0 +1,917 @@
...
+/**
+ * Enhances a given rule `create` with helpers to detect Testing Library utils.
+ */
+export const detectTestingLibraryUtils = <
+  TOptions extends readonly unknown[],
+  TMessageIds extends string,
+  TRuleListener extends TSESLint.RuleListener = TSESLint.RuleListener
+>(
+  ruleCreate: EnhancedRuleCreate<TOptions, TMessageIds, TRuleListener>,
+  { skipRuleReportingCheck = false }: Partial<DetectionOptions> = {}
```
MichaelDeBoey, line 147: Part 2

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -0,0 +1,917 @@
...
+    }
+
+    return findImportedUtilSpecifier(identifierName);
+  };
+
+  /**
+   * Determines if file inspected meets all conditions to be reported by rules or not.
+   */
+  const canReportErrors: CanReportErrorsFn = () => {
+    return skipRuleReportingCheck || isTestingLibraryImported();
```
MichaelDeBoey, line 752: Part 3

MichaelDeBoey (commented):
I don't understand why GH is saying that `detect-testing-library-utils.ts` is a new file, as I only changed these parts and moved the file to the `create-testing-library-rule` folder

Belco90:
> I don't understand why GH is saying that `detect-testing-library-utils.ts` is a new file, as I only changed these parts and moved the file to the `create-testing-library-rule` folder

Could you avoid moving the file then? It's doing more harm than good, and we can move it in a different PR to avoid adding more noise here.

Belco90:
> @Belco90 could you point me in the right direction of how I would add tests for it?

Sure! So the plugin on v4 won't report anything on a file if there is nothing related to Testing Library imported (this is only checked when Aggressive Reporting is opted out). Where does this check happen? In `detect-testing-library-utils.ts` you can see at the very bottom something like this:
```javascript
if (canReportErrors() && ruleInstructions[instruction]) {
  return ruleInstructions[instruction]?.(node);
}
```

That's the check described before. This check is fine for all rules except `consistent-data-testid`, since this rule doesn't apply to test but component files.

It should be easy to skip the check with the `skipRuleReportingCheck` option, so if it's `true` then the previous check must be skipped:
```javascript
if ((skipRuleReportingCheck || canReportErrors()) && ruleInstructions[instruction]) {
  return ruleInstructions[instruction]?.(node);
}
```

About how to test it, all current `consistent-data-testid` tests must pass. You should duplicate some of the current cases, both valid and invalid ones, but switching Aggressive Reporting off through `settings` property, and the test result should be exactly the same as the original test case.

Let me know if you need more assistance over this!

lib/create-testing-library-rule/index.ts
```diff
@@ -11,24 +14,25 @@ type CreateRuleMeta<TMessageIds extends string> = {
   docs: CreateRuleMetaDocs;
 } & Omit<TSESLint.RuleMetaData<TMessageIds>, 'docs'>;
 
-export function createTestingLibraryRule<
+export const createTestingLibraryRule = <
```
Belco90, line 17: Why changing this from regular function to arrow function? Could we please avoid this sort of personal preference? They don't solve anything and add noise to the PR. We could even spend the whole life switching from one definition way to the other every time someone else modifies this file 😓

lib/create-testing-library-rule/detect-testing-library-utils.ts
Belco90: Oh, I see you already did the change. This is even better than where I suggested putting the check! Moving the file added a lot of noise tho, please let's revert that.

Belco90 (changes requested):

Belco90 (approved):

Issue #360, opened 2021-04-26
Make `no-wait-for-side-effects` report render() inside `waitFor`


zaicevas:
Consider this: `await waitFor(() => render(<Component />))`. AFAIC it doesn't seem like a legit use case 🤔   Should we report every usage of `render*` wrapped in `waitFor` ?

Belco90:
Hey @zaicevas! This is an interesting proposal. I think you are right and `no-wait-for-side-effects` should report `render` util calls inside `waitFor` indeed, and I guess `rerender` too by extension.

It would be nice to include this report in `no-wait-for-side-effects` but this can't be enabled by default. Following our [Semantic Versioning Policy](https://github.com/testing-library/eslint-plugin-testing-library/discussions/343), "A new option to an existing rule that results in ESLint reporting more linting errors by default" would be a major version. So we need to aim for "A new option to an existing rule that does not result in ESLint reporting more linting errors by default" change which would result in a minor (and then make it a default option in the next major).

Following this workflow, we would need to include a new option (e.g `shouldReportRender`) accepting a boolean which defaults to `false`, and report this new case when it's enabled so we don't force a major release. What do you think?

zaicevas:
> Hey @zaicevas! This is an interesting proposal. I think you are right and `no-wait-for-side-effects` should report `render` util calls inside `waitFor` indeed, and I guess `rerender` too by extension.
> 
> It would be nice to include this report in `no-wait-for-side-effects` but this can't be enabled by default. Following our [Semantic Versioning Policy](https://github.com/testing-library/eslint-plugin-testing-library/discussions/343), "A new option to an existing rule that results in ESLint reporting more linting errors by default" would be a major version. So we need to aim for "A new option to an existing rule that does not result in ESLint reporting more linting errors by default" change which would result in a minor (and then make it a default option in the next major).
> 
> Following this workflow, we would need to include a new option (e.g `shouldReportRender`) accepting a boolean which defaults to `false`, and report this new case when it's enabled so we don't force a major release. What do you think?

What you're saying makes total sense. Do you think we could enable `shouldReportRender` by default in v15?

By the way, I'd like to try to implement it 😄

Belco90:
[We are still debating about our SemVer Policy](https://github.com/testing-library/eslint-plugin-testing-library/discussions/343), so I think it would be better if we report that error by default in the rule without any option, and then we see if we release it as a minor or we need to opt it out with some option.  Are you happy with that?

[...]

PR #363, opened 2021-05-01
feat(no-wait-for-side-effects): report render in waitFor

zaicevas:
Closes https://github.com/testing-library/eslint-plugin-testing-library/issues/360

We should decide how we want to release this enhancement. Seems like the discussion is centered around:
a) minor version bump without an option
b) minor version bump with an option, which is off by default

Also, I've noticed this rule is NOT reporting this case. Is that intentional? 🤔 

```
    await waitFor(() => {
      userEvent.click(button) // reported
      const a = userEvent.click(button) // NOT reported
    })
```

### DIFF
--- a/lib/create-testing-library-rule/detect-testing-library-utils.ts
+++ b/lib/create-testing-library-rule/detect-testing-library-utils.ts
@@ -75,7 +75,7 @@ type IsAsyncUtilFn = (
 ) => boolean;
 type IsFireEventMethodFn = (node: TSESTree.Identifier) => boolean;
 type IsUserEventMethodFn = (node: TSESTree.Identifier) => boolean;
-type IsRenderUtilFn = (node: TSESTree.Identifier) => boolean;
+type IsRenderUtilFn = (node: TSESTree.Identifier | null) => boolean;
 type IsRenderVariableDeclaratorFn = (
   node: TSESTree.VariableDeclarator
 ) => boolean;
@@ -105,8 +105,8 @@ export interface DetectionHelpers {
   isCustomQuery: IsCustomQueryFn;
   isBuiltInQuery: IsBuiltInQueryFn;
   isAsyncUtil: IsAsyncUtilFn;
-  isFireEventUtil: (node: TSESTree.Identifier) => boolean;
-  isUserEventUtil: (node: TSESTree.Identifier) => boolean;
+  isFireEventUtil: (node: TSESTree.Identifier | null) => boolean;
+  isUserEventUtil: (node: TSESTree.Identifier | null) => boolean;
   isFireEventMethod: IsFireEventMethodFn;
   isUserEventMethod: IsUserEventMethodFn;
   isRenderUtil: IsRenderUtilFn;
@@ -429,7 +429,9 @@ export function detectTestingLibraryUtils<
      *
      * Not to be confused with {@link isFireEventMethod}
      */
-    const isFireEventUtil = (node: TSESTree.Identifier): boolean => {
+    const isFireEventUtil = (node: TSESTree.Identifier | null): boolean => {
+      if (!node) return false;
+
       return isTestingLibraryUtil(
         node,
         (identifierNodeName, originalNodeName) => {
@@ -443,7 +445,9 @@ export function detectTestingLibraryUtils<
      *
      * Not to be confused with {@link isUserEventMethod}
      */
-    const isUserEventUtil = (node: TSESTree.Identifier): boolean => {
+    const isUserEventUtil = (node: TSESTree.Identifier | null): boolean => {
+      if (!node) return false;
+
       const userEvent = findImportedUserEventSpecifier();
       let userEventName: string | undefined;
 
@@ -569,18 +573,25 @@ export function detectTestingLibraryUtils<
      * Testing Library. Otherwise, it means `custom-module` has been set up, so
      * only those nodes coming from Testing Library will be considered as valid.
      */
-    const isRenderUtil: IsRenderUtilFn = (node) =>
-      isTestingLibraryUtil(node, (identifierNodeName, originalNodeName) => {
-        if (isAggressiveRenderReportingEnabled()) {
-          return identifierNodeName.toLowerCase().includes(RENDER_NAME);
-        }
+    const isRenderUtil: IsRenderUtilFn = (node) => {
+      if (!node) return false;
 
-        return [RENDER_NAME, ...getCustomRenders()].some(
-          (validRenderName) =>
-            validRenderName === identifierNodeName ||
-            (Boolean(originalNodeName) && validRenderName === originalNodeName)
-        );
-      });
+      return isTestingLibraryUtil(
+        node,
+        (identifierNodeName, originalNodeName) => {
+          if (isAggressiveRenderReportingEnabled()) {
+            return identifierNodeName.toLowerCase().includes(RENDER_NAME);
+          }
+
+          return [RENDER_NAME, ...getCustomRenders()].some(
+            (validRenderName) =>
+              validRenderName === identifierNodeName ||
+              (Boolean(originalNodeName) &&
+                validRenderName === originalNodeName)
+          );
+        }
+      );
+    };
 
     const isRenderVariableDeclarator: IsRenderVariableDeclaratorFn = (node) => {
       if (!node.init) {
--- a/lib/create-testing-library-rule/detect-testing-library-utils.ts
+++ b/lib/create-testing-library-rule/detect-testing-library-utils.ts
@@ -173,7 +173,7 @@ export function detectTestingLibraryUtils<
      *    reporting)
      */
     function isTestingLibraryUtil(
-      node: TSESTree.Identifier,
+      node: TSESTree.Identifier | null,
       isUtilCallback: (
         identifierNodeName: string,
         originalNodeName?: string
@@ -430,8 +430,6 @@ export function detectTestingLibraryUtils<
      * Not to be confused with {@link isFireEventMethod}
      */
     const isFireEventUtil = (node: TSESTree.Identifier | null): boolean => {
-      if (!node) return false;
-
       return isTestingLibraryUtil(
         node,
         (identifierNodeName, originalNodeName) => {
@@ -573,25 +571,18 @@ export function detectTestingLibraryUtils<
      * Testing Library. Otherwise, it means `custom-module` has been set up, so
      * only those nodes coming from Testing Library will be considered as valid.
      */
-    const isRenderUtil: IsRenderUtilFn = (node) => {
-      if (!node) return false;
-
-      return isTestingLibraryUtil(
-        node,
-        (identifierNodeName, originalNodeName) => {
-          if (isAggressiveRenderReportingEnabled()) {
-            return identifierNodeName.toLowerCase().includes(RENDER_NAME);
-          }
-
-          return [RENDER_NAME, ...getCustomRenders()].some(
-            (validRenderName) =>
-              validRenderName === identifierNodeName ||
-              (Boolean(originalNodeName) &&
-                validRenderName === originalNodeName)
-          );
+    const isRenderUtil: IsRenderUtilFn = (node) =>
+      isTestingLibraryUtil(node, (identifierNodeName, originalNodeName) => {
+        if (isAggressiveRenderReportingEnabled()) {
+          return identifierNodeName.toLowerCase().includes(RENDER_NAME);
         }
-      );
-    };
+
+        return [RENDER_NAME, ...getCustomRenders()].some(
+          (validRenderName) =>
+            validRenderName === identifierNodeName ||
+            (Boolean(originalNodeName) && validRenderName === originalNodeName)
+        );
+      });
 
     const isRenderVariableDeclarator: IsRenderVariableDeclaratorFn = (node) => {
       if (!node.init) {
--- a/lib/create-testing-library-rule/detect-testing-library-utils.ts
+++ b/lib/create-testing-library-rule/detect-testing-library-utils.ts
@@ -444,7 +444,9 @@ export function detectTestingLibraryUtils<
      * Not to be confused with {@link isUserEventMethod}
      */
     const isUserEventUtil = (node: TSESTree.Identifier | null): boolean => {
-      if (!node) return false;
+      if (!node) {
+        return false;
+      }
 
       const userEvent = findImportedUserEventSpecifier();
       let userEventName: string | undefined;
--- a/lib/create-testing-library-rule/detect-testing-library-utils.ts
+++ b/lib/create-testing-library-rule/detect-testing-library-utils.ts
@@ -75,7 +75,7 @@ type IsAsyncUtilFn = (
 ) => boolean;
 type IsFireEventMethodFn = (node: TSESTree.Identifier) => boolean;
 type IsUserEventMethodFn = (node: TSESTree.Identifier) => boolean;
-type IsRenderUtilFn = (node: TSESTree.Identifier | null) => boolean;
+type IsRenderUtilFn = (node: TSESTree.Identifier) => boolean;
 type IsRenderVariableDeclaratorFn = (
   node: TSESTree.VariableDeclarator
 ) => boolean;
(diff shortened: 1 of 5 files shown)


lib/rules/no-wait-for-side-effects.ts
```diff
@@ -52,18 +60,31 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
         const expressionIdentifier = getPropertyIdentifierNode(node);
-        if (!expressionIdentifier) {
-          return false;
-        }
+
+        const isRenderInVariableDeclaration =
+          isVariableDeclaration(node) &&
+          node.declarations.some((declaration) =>
+            helpers.isRenderVariableDeclarator(declaration)
+          );
```
MichaelDeBoey, line 73: I think early returns are cleaner
```suggestion
        const isRenderInVariableDeclaration =
          isVariableDeclaration(node) &&
          node.declarations.some((declaration) =>
            helpers.isRenderVariableDeclarator(declaration)
          );
          if (!isRenderInVariableDeclaration) {
            return false;
          }
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -52,18 +60,31 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+
+        const isRenderInAssignment =
+          isExpressionStatement(node) &&
+          isAssignmentExpression(node.expression) &&
+          helpers.isRenderUtil(
+            getPropertyIdentifierNode(node.expression.right)
+          );
```
MichaelDeBoey, line 80:
```suggestion
        const isRenderInAssignment =
          isExpressionStatement(node) &&
          isAssignmentExpression(node.expression) &&
          helpers.isRenderUtil(
            getPropertyIdentifierNode(node.expression.right)
          );
          if (!isRenderInAssignment) {
            return false;
          }
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -52,18 +60,31 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
 
         return (
           helpers.isFireEventUtil(expressionIdentifier) ||
-          helpers.isUserEventUtil(expressionIdentifier)
+          helpers.isUserEventUtil(expressionIdentifier) ||
+          helpers.isRenderUtil(expressionIdentifier) ||
+          isRenderInVariableDeclaration ||
+          isRenderInAssignment
```
MichaelDeBoey, line 87:
```suggestion
          helpers.isRenderUtil(expressionIdentifier)
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -52,18 +60,31 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+          );
+
+        const isRenderInAssignment =
+          isExpressionStatement(node) &&
+          isAssignmentExpression(node.expression) &&
+          helpers.isRenderUtil(
+            getPropertyIdentifierNode(node.expression.right)
+          );
 
         return (
```
MichaelDeBoey, line 82:
```suggestion
        const expressionIdentifier = getPropertyIdentifierNode(node);
        return (
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -52,18 +60,31 @@ export default createTestingLibraryRule<Options, MessageIds>({
       body: TSESTree.Node[]
     ): TSESTree.ExpressionStatement[] {
       return body.filter((node) => {
-        if (!isExpressionStatement(node)) {
+        if (!isExpressionStatement(node) && !isVariableDeclaration(node)) {
           return false;
         }
 
         const expressionIdentifier = getPropertyIdentifierNode(node);
```
MichaelDeBoey, line 67:
```suggestion
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -86,19 +107,35 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
 
-      const expressionIdentifier = getPropertyIdentifierNode(node.callee);
-      if (!expressionIdentifier) {
-        return;
-      }
+      const expressionIdentifier =
+        isCallExpression(node) && getPropertyIdentifierNode(node.callee);
+      const isRenderInAssignment =
+        isAssignmentExpression(node) &&
+        helpers.isRenderUtil(getPropertyIdentifierNode(node.right));
```
MichaelDeBoey, line 124:
```suggestion
      const isRenderInAssignment =
        isAssignmentExpression(node) &&
        helpers.isRenderUtil(getPropertyIdentifierNode(node.right));
      if (!isRenderInAssignment) {
        return;
      }
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -86,19 +107,35 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+      const isRenderInSequenceAssignment =
+        isSequenceExpression(node) &&
+        node.expressions.some(
+          (expression) =>
+            isAssignmentExpression(expression) &&
+            helpers.isRenderUtil(getPropertyIdentifierNode(expression.right))
+        );
```
MichaelDeBoey, line 131:
```suggestion

      const isRenderInSequenceAssignment =
        isSequenceExpression(node) &&
        node.expressions.some(
          (expression) =>
            isAssignmentExpression(expression) &&
            helpers.isRenderUtil(getPropertyIdentifierNode(expression.right))
        );
        if (!isRenderInSequenceAssignment) {
          return
        }
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -86,19 +107,35 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
 
       if (
-        !helpers.isFireEventUtil(expressionIdentifier) &&
-        !helpers.isUserEventUtil(expressionIdentifier)
+        !helpers.isFireEventUtil(expressionIdentifier || null) &&
+        !helpers.isUserEventUtil(expressionIdentifier || null) &&
+        !helpers.isRenderUtil(expressionIdentifier || null) &&
+        !isRenderInAssignment &&
+        !isRenderInSequenceAssignment
```
MichaelDeBoey, line 138:
```suggestion
        !helpers.isFireEventUtil(expressionIdentifier) &&
        !helpers.isUserEventUtil(expressionIdentifier) &&
        !helpers.isRenderUtil(expressionIdentifier)
```

MichaelDeBoey (changes requested):
Some small improvements to make the code a bit more readable, but otherwise LGTM! 👊

Belco90:
> Also, I've noticed this rule is NOT reporting this case. Is that intentional? 🤔 
> 
> 
> 
> ```
> 
>     await waitFor(() => {
> 
>       userEvent.click(button) // reported
> 
>       const a = userEvent.click(button) // NOT reported
> 
>     })
> 
> ```

Good catch. It's not intentional, but I detected it when implementing `no-unnecessary-act`, so it will be addressed there. I'll make sure to include this case in the tests.

lib/rules/no-wait-for-side-effects.ts
zaicevas: Early returns wouldn't work here, because we need to account for `helpers.isFireEventUtil(expressionIdentifier)` and other cases. Am I missing something?

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -443,7 +443,9 @@ export function detectTestingLibraryUtils<
      *
      * Not to be confused with {@link isUserEventMethod}
      */
-    const isUserEventUtil = (node: TSESTree.Identifier): boolean => {
+    const isUserEventUtil = (node: TSESTree.Identifier | null): boolean => {
+      if (!node) return false;
```
MichaelDeBoey, line 447: Add brackets around body, to make it less error-prone

```suggestion
      if (!node) {
        return false;
      }
```

MichaelDeBoey (changes requested):

lib/rules/no-wait-for-side-effects.ts
MichaelDeBoey: As you're using `||` instead of `&&`, you're probably right for this function

zaicevas: We could have an early return when the clause is `true`. Do you think that's better?

```
    function getSideEffectNodes(
      body: TSESTree.Node[]
    ): TSESTree.ExpressionStatement[] {
      return body.filter((node) => {
        if (!isExpressionStatement(node) && !isVariableDeclaration(node)) {
          return false;
        }

        const isRenderInVariableDeclaration =
          isVariableDeclaration(node) &&
          node.declarations.some((declaration) =>
            helpers.isRenderVariableDeclarator(declaration)
          );

        if (isRenderInVariableDeclaration) {
          return true;
        }

        const isRenderInAssignment =
          isExpressionStatement(node) &&
          isAssignmentExpression(node.expression) &&
          helpers.isRenderUtil(
            getPropertyIdentifierNode(node.expression.right)
          );

        if (isRenderInAssignment) {
          return true;
        }

        const expressionIdentifier = getPropertyIdentifierNode(node);

        return (
          helpers.isFireEventUtil(expressionIdentifier) ||
          helpers.isUserEventUtil(expressionIdentifier) ||
          helpers.isRenderUtil(expressionIdentifier)
        );
      }) as TSESTree.ExpressionStatement[];
    }
```

MichaelDeBoey: That's maybe a good idea indeed

Just to not make the final return too big and making the cognitive load lower when looking at the code.

MichaelDeBoey: Extracting the calculations into a separate function can help too for readability btw 🤔

And I think I prefer to do it that way actually 🤔

CC/ @Belco90

lib/rules/no-wait-for-side-effects.ts
zaicevas: Early returns worked for `getSideEffectNodes` function, doesn't seem like we can introduce them here 🤔

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -48,22 +56,63 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
-        const expressionIdentifier = getPropertyIdentifierNode(node);
-        if (!expressionIdentifier) {
-          return false;
+        if (isRenderInVariableDeclaration(node)) {
+          return true;
         }
 
+        if (isRenderInAssignment(node)) {
+          return true;
+        }
```
MichaelDeBoey, line 108:
```suggestion
        if (isRenderInVariableDeclaration(node) || isRenderInAssignment(node)) {
          return true;
        }
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -86,19 +135,26 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
-    function reportImplicitReturnSideEffect(node: TSESTree.CallExpression) {
+    function reportImplicitReturnSideEffect(
+      node:
+        | TSESTree.CallExpression
+        | TSESTree.AssignmentExpression
+        | TSESTree.SequenceExpression
+    ) {
       if (!isCallerWaitFor(node)) {
         return;
       }
```
MichaelDeBoey, line 146:
```suggestion
      if (!isCallerWaitFor(node) || isRenderInImplicitReturnAssignment(node) || isRenderInSequenceAssignment(node)) {
        return;
      }
```

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -86,19 +135,26 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+        ? getPropertyIdentifierNode(node.callee)
+        : null;
 
       if (
         !helpers.isFireEventUtil(expressionIdentifier) &&
-        !helpers.isUserEventUtil(expressionIdentifier)
+        !helpers.isUserEventUtil(expressionIdentifier) &&
+        !helpers.isRenderUtil(expressionIdentifier) &&
+        !isRenderInImplicitReturnAssignment(node) &&
+        !isRenderInSequenceAssignment(node)
```
MichaelDeBoey, line 157:
```suggestion
        !helpers.isUserEventUtil(expressionIdentifier) &&
        !helpers.isRenderUtil(expressionIdentifier)
```

MichaelDeBoey (changes requested):

lib/rules/no-wait-for-side-effects.ts
zaicevas: I am afraid that won't work, as it can still be one of those:
```
        !helpers.isFireEventUtil(expressionIdentifier) &&
        !helpers.isUserEventUtil(expressionIdentifier) &&
        !helpers.isRenderUtil(expressionIdentifier)
```

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -105,8 +105,8 @@ export interface DetectionHelpers {
   isCustomQuery: IsCustomQueryFn;
   isBuiltInQuery: IsBuiltInQueryFn;
   isAsyncUtil: IsAsyncUtilFn;
-  isFireEventUtil: (node: TSESTree.Identifier) => boolean;
-  isUserEventUtil: (node: TSESTree.Identifier) => boolean;
+  isFireEventUtil: (node: TSESTree.Identifier | null) => boolean;
+  isUserEventUtil: (node: TSESTree.Identifier | null) => boolean;
```
Belco90, line 109: Could you revert all helpers here so they don't expect any `null` node, please?  So all helpers follow a similar signature, and it's also up to whichever code calling this to make sure it's expected node type before calling these helpers.

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -48,22 +56,59 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+    function isRenderInImplicitReturnAssignment(node: TSESTree.Node) {
+      return (
+        isAssignmentExpression(node) &&
+        helpers.isRenderUtil(getPropertyIdentifierNode(node.right))
+      );
+    }
+
+    function isRenderInSequenceAssignment(node: TSESTree.Node) {
+      return (
+        isSequenceExpression(node) &&
```
Belco90, line 85: I think you'll be able to reuse [this `getStatementCallExpression` method I added for `no-unnecessary-act`](https://github.com/testing-library/eslint-plugin-testing-library/commit/7a2cba1c171b63f112f3515b147334baf958cb27#diff-823fe8bec495554448323badd523e6b09e436e4ad7e77b5ef6add3ba4024cc38R515).

I'm really close to creating my PR. If mine goes first, you can reuse it here. Otherwise, I'll try to refactor these functions with `getStatementCallExpression`.

Belco90 (changes requested):
@zaicevas Thanks for your PR, it looks great! Just some small tweaks about helpers receiving `null` nodes and I think it's ready to go.

lib/create-testing-library-rule/detect-testing-library-utils.ts
zaicevas: Sure, but as you can see in 5518751, it adds a lot of noise, so I am not too sure if that's a better approach 😅

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -48,22 +56,92 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+      }
+
+      const expressionIdentifier = getPropertyIdentifierNode(expression.right);
+
+      if (!expressionIdentifier) {
+        return false;
+      }
+
+      return helpers.isRenderUtil(expressionIdentifier);
+    }
```
MichaelDeBoey, line 113: As these 2 functions are doing exactly the same and only having the argument type differently, I would merge them

MichaelDeBoey (changes requested):

Belco90 (changes requested):
Pretty happy with the implementation and its tests! Last requirement: could you mention `render` will be reported in the rule docs, and include it in some invalid example?

zaicevas:
> Pretty happy with the implementation and its tests! Last requirement: could you mention `render` will be reported in the rule docs, and include it in some invalid example?

Done :) Not sure which link to add in *Further Reading*, tho (if any)

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -48,22 +56,80 @@ export default createTestingLibraryRule<Options, MessageIds>({
       );
     }
 
+    function isRenderInVariableDeclaration(node: TSESTree.Node) {
+      return (
+        isVariableDeclaration(node) &&
+        node.declarations.some((declaration) =>
+          helpers.isRenderVariableDeclarator(declaration)
+        )
```
MichaelDeBoey, line 64:
```suggestion
        node.declarations.some(helpers.isRenderVariableDeclarator)
```

MichaelDeBoey (changes requested):
Just one small thing and than this one can be merged for me

MichaelDeBoey (approved):
LGTM! 👊

Belco90 (approved):
No more further reading links are needed, with existing ones + rule description is more than enough.

Releasing this as a minor then. We haven't set the final SemVer Policy but this sort of extra error reported would be expected at the Feature release level.

I finally won't adapt this implementation with the new function I added to `no-unnecessary-act` since my PR is gonna have tons of changes, but I'll note it to do it in a separate PR (or even you can do it if you feel so!)

Thank you very much for your work here @zaicevas!

PR #364, opened 2021-05-01
refactor(no-wait-for-side-effects): simplify `reportSideEffects`

MichaelDeBoey:
When looking at #363, I saw this part could be simplified

### DIFF
--- a/lib/rules/no-wait-for-side-effects.ts
+++ b/lib/rules/no-wait-for-side-effects.ts
@@ -73,17 +73,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
         return;
       }
 
-      const sideEffectNodes = getSideEffectNodes(node.body);
-      if (sideEffectNodes.length === 0) {
-        return;
-      }
-
-      for (const sideEffectNode of sideEffectNodes) {
+      getSideEffectNodes(node.body).forEach((sideEffectNode) =>
         context.report({
           node: sideEffectNode,
           messageId: 'noSideEffectsWaitFor',
-        });
-      }
+        })
+      );
     }
 
     function reportImplicitReturnSideEffect(node: TSESTree.CallExpression) {


gndelia (approved):

gndelia:
Nice simplification, LGTM 🚀

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -73,17 +73,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
         return;
       }
 
-      const sideEffectNodes = getSideEffectNodes(node.body);
-      if (sideEffectNodes.length === 0) {
-        return;
-      }
-
-      for (const sideEffectNode of sideEffectNodes) {
+      getSideEffectNodes(node.body).forEach((sideEffectNode) =>
```
Belco90, line 76: I was planning to do the opposite in the repo: replace all `forEach` by `for...of` enabling [this rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md) so can you go back to `for...of` please?

Belco90 (changes requested):

lib/rules/no-wait-for-side-effects.ts
MichaelDeBoey: Any reason why you think `for...of` is preferred?

Using `forEach` is most of the time shorter and doesn't set the door open for populating arrays inside of loops, but makes `filter`/`map`/... a more logical way to go
```ts
// from
let resultArray = [];

for (const item of otherArray) {
  if (conditional) {
    resultArray.push(item);
  }
}

// to
const resultArray = otherArray.filter(condition);
```
or
```ts
// from
let resultArray = [];

for (const item of otherArray) {
  resultArray.push(item.key);
}

// to
const resultArray = otherArray.map(item => item.key);
```

Belco90: The rule mentioned sums it up pretty well:

- Faster
- Better readability
- Ability to exit early with `break` or `return`

Belco90: The examples you mentioned doesn't apply to `for...of` vs `forEach`

When you want to loop an array in functional way for manipulating it of course it's ok to use `map`, `filter` and so on (it should actually be the way).

In the case here we are not manipulating the array, but looping it to do something else than manipulating the array itself. In that case there are 3 variants: `for`, `forEach` and `for...of`. `forEach` looks like you are doing something _funcional_ but you aren't, that's why `for...of` has "better readability: at first look you know there is no array manipulation and it's just iterating the array without changing it.

MichaelDeBoey: > * Faster

This is about milliseconds, so I wouldn't care about that.  
Also: V8 team is working on making `forEach`/`map`/`filter` faster than it is (and eventually faster than `for...of`)

> * Better readability

"Better" is very subjective. But using `for...of` has a higher cognitive load for sure, so that's why I don't like it.

> * Ability to exit early with `break` or `return`

You can always use a `for...of` when this is the case, but mostly you're mapping/filtering/reducing/looping over an array that doesn't need a `break` or `return`

Belco90: Technically all for loop variants ends in the same for shape with V8, that's why I did't focus on that advantage on my prev explanation.

The point is prefer `for...of` over `forEach`, not any other functional array method, because it helps with "manipulating array" vs "iterating array" readability.

Anyway, I don't think we should be spending this time in discussing such preferences. If it works, it works, let's move on to something else and let tools as ESLint to make styles consistent. Otherwise we will spend 90% of our time in this repo rewriting code which is already working fine only because it's not written by our personal preferences.

MichaelDeBoey: For me using `for...of` over `forEach` sets the door open to using it for other cases (like map/filter/reduce/...) too, so that's why I prefer to always use the Array functions.

But I agree that we should discuss this in a separate issue/PR and not pollute this one. 👍

Belco90 (approved):

PR #365, opened 2021-05-02
feat: new `no-unnecessary-act` rule

### LINKED ISSUE #259: add no-unnecessary-act rule
We should warn against the use of `act` unnecessarily:

```javascript
import {screen, fireEvent, render, act} from '@testing-library/react'
import userEvent from '@testing-library/user-event`
import ReactTestUtils from 'react-dom/test-utils'

// ReactTestUtils.act should be treated exactly the same as `act` from `@testing-library/react`

// these are ALL wrong:
act(() => fireEvent.click(el))
act(() => screen.getByText('blah'))
act(() => userEvent.click(el))
await act(async () => userEvent.type('hi', el))
act(() => render(<div />))
await act(async () => render(<div />))
await act(async () => {})
act(() => {})

// this one's fine:
act(() => {
  stuffThatDoesNotUseRTL()
})

// actually, so is this one
act(() => {
  fireEvent.click(el)
  stuffThatDoesNotUseRTL()
})
```

So basically the rule is: "You should only wrap non-testing-library function calls in `act`"

Reading material: https://kcd.im/react-act

Belco90:
Closes #259 

Co-authored-by: @alessbell

I've found few issues detecting Testing Library packages during the development of this rule. I've fixed one of them, and I'll pinpoint the rest of them in this very PR so I can fix them separately after this rule gets merged.


README.md
```diff
@@ -193,15 +193,16 @@ To enable this configuration use the `extends` property in your
...
 | [testing-library/prefer-presence-queries](docs/rules/prefer-presence-queries.md)                 | Enforce specific queries when checking element is present or not                         |                                                                   |                    |
-| [testing-library/prefer-user-event](docs/rules/prefer-user-event.md)                             | Suggest using `userEvent` library instead of `fireEvent` for simulating user interaction |                                                                   |                    |
 | [testing-library/prefer-screen-queries](docs/rules/prefer-screen-queries.md)                     | Suggest using screen while using queries                                                 | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |                    |
+| [testing-library/prefer-user-event](docs/rules/prefer-user-event.md)                             | Suggest using `userEvent` library instead of `fireEvent` for simulating user interaction |                                                                   |                    |
```
Belco90, line 205: Just sorting alphabetically here.

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -172,9 +176,9 @@ export function detectTestingLibraryUtils<
      * - it's imported from valid Testing Library module (depends on aggressive
      *    reporting)
      */
-    function isTestingLibraryUtil(
+    function isPotentialTestingLibraryFunction(
```
Belco90, line 179: Renaming this method to avoid confusion with the actual method to check if a node is a Testing Library util.

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -200,7 +204,7 @@ export function detectTestingLibraryUtils<
           ? importedUtilSpecifier.imported.name
           : undefined;
 
-      if (!isUtilCallback(node.name, originalNodeName)) {
+      if (!isPotentialFunctionCallback(node.name, originalNodeName)) {
```
Belco90, line 207: This check is one of the problems I found that I'll addresss separately. Instead of the current check, this should be done using `hasImportMatch` + `referenceNodeIdentifier.name`.

This causes some edge cases using wildcard imports (e.g. `import * as RTL from '@testing-library/react') not to be detected properly.

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -614,6 +624,91 @@ export function detectTestingLibraryUtils<
...
+      const isTestingLibraryAct = isPotentialTestingLibraryFunction(
+        node,
+        (identifierNodeName, originalNodeName) => {
+          return [identifierNodeName, originalNodeName]
+            .filter(Boolean)
+            .includes('act');
+        }
+      );
+
+      const isReactDomTestUtilsAct = (() => {
```
Belco90, line 644: I know there are several IIFE used here and some blocks could be refactored or extracted. I'm gonna leave it like this so first I fix the issues found and then, having proper context, I'll refactor and extract corresponding functions.

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -614,6 +624,91 @@ export function detectTestingLibraryUtils<
...
+        }
+
+        if (importDeclarationName !== REACT_DOM_TEST_UTILS_PACKAGE) {
+          return false;
+        }
+
+        return hasImportMatch(
+          importedUtilSpecifier,
+          referenceNodeIdentifier.name
+        );
```
Belco90, line 695: This is actually how `isPotentialTestingLibraryFunction` should check if the import matches the given node.

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -769,20 +822,57 @@ export function detectTestingLibraryUtils<
...
       if (!identifierName) {
         return false;
       }
 
-      return hasImportMatch(importNode, identifierName);
+      const hasImportElementMatch = hasImportMatch(importNode, identifierName);
+      const hasImportModuleMatch =
+        /testing-library/g.test(importDeclarationName) ||
+        (typeof customModuleSetting === 'string' &&
+          importDeclarationName.endsWith(customModuleSetting));
```
Belco90, line 873: This is another of the issues I found and I needed to fix already. Without this check, an identifier matching some Testing Library util name but imported from an unrelated module could be assumed as coming from Testing Library package.

```js
import { act as renamedAct } from '@testing-library/react'
import { act } from 'somewhere-else'

// this should not be reported since it's an `act` but not coming from Testing Library,
// but without previous fix it was reported since it found an import matching the specifier.
act(() => {})
```

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -769,20 +822,57 @@ export function detectTestingLibraryUtils<
...
+
       const identifierName: string | undefined = getPropertyIdentifierNode(node)
         ?.name;
 
       if (!identifierName) {
         return false;
       }
 
-      return hasImportMatch(importNode, identifierName);
+      const hasImportElementMatch = hasImportMatch(importNode, identifierName);
```
Belco90, line 869: This has to be updated too, so it uses `referenceNodeIdentifier.name` rather than `identifierName`.

lib/node-utils/index.ts
```diff
@@ -76,6 +81,23 @@ export function findClosestCallExpressionNode(
...
+  if (!node) {
+    return null;
+  }
+
+  if (ASTUtils.isVariableDeclarator(node)) {
+    return node;
+  }
+
+  return findClosestVariableDeclaratorNode(node.parent);
+}
```
Belco90, line 96: I'm pretty sure this could be refactored in a similar way to `isNodeOfType` so it would be possible to find the closest node of any kind, but I don't think this is the PR to do so.

lib/node-utils/index.ts
```diff
@@ -510,3 +532,95 @@ export function hasImportMatch(
 
   return importNode.local.name === identifierName;
 }
+
+export function getStatementCallExpression(
```
Belco90, line 536: This is the new util that should help to fix #368

lib/node-utils/index.ts
```diff
@@ -510,3 +532,95 @@ export function hasImportMatch(
...
+    return node.body.body.length === 0;
+  }
+
+  return false;
+}
+
+/**
+ * Finds the import specifier matching a given name for a given import module node.
+ */
+export function findImportSpecifier(
```
Belco90, line 580: This was extracted from `detect-testing-library-utils` module.

tests/lib/rules/no-unnecessary-act.test.ts
```diff
@@ -0,0 +1,782 @@
...
+        { messageId: 'noUnnecessaryActEmptyFunction', line: 6, column: 9 },
+        { messageId: 'noUnnecessaryActEmptyFunction', line: 7, column: 15 },
+        { messageId: 'noUnnecessaryActEmptyFunction', line: 8, column: 9 },
+      ],
+    },
+
+    // cases for act related to React Test Utils
+    {
+      settings: {
+        'testing-library/utils-module': 'custom-testing-module',
```
Belco90, line 479: We usually set this setting to `"test-utils"` for testing Aggressive Reporting disabled.

It has been named differently here and this is not by chance. This is another of the issues I found. If I set this to `"test-utils"`, I would have 2 different imports ending with the same string: `"test-utils"` and `"react-dom/test-utils"`. This causes our import checks to work incorrectly, so I'll have to handle it in a separate PR.

Meanwhile, I'm using a different custom module here to avoid collision with `"react-dom/test-utils"`.

Belco90:
forever alone meme

docs/rules/no-unnecessary-act.md
```diff
@@ -0,0 +1,117 @@
...
+
+// ...
+
+act(() => {});
+
+await act(async () => {});
+```
+
+```js
+// ❌ wrapping things related to Testing Library in React DOM Test Utils `act` is also incorrect
```
kentcdodds, line 64: Isn't this the same as the code examples above?

kentcdodds (commented):
I don't know enough about this plugin to provide very valuable feedback on the code, but here's one bit of feedback on the docs

kentcdodds:
Other than that, this looks great!

docs/rules/no-unnecessary-act.md
Belco90: Almost. The only difference is that `act` is imported from `react-dom/test-utils` rather than `@testing-library/react`. I guess I can add a comment with an arrow pointing that difference or make it clear changing the example.

kentcdodds: Couldn't you remove this duplicate stuff and add a comment above indicating it's the same for both.

docs/rules/no-unnecessary-act.md
```diff
@@ -0,0 +1,94 @@
+# Disallow wrapping Testing Library utils or empty callbacks in `act` (`testing-library/no-unnecessary-act`)
+
+> ⚠️ The `act` method is only available on the following Testing Library packages:
+>
+> - `@testing-library/react` (supported by this plugin)
+> - `@testing-library/preact` (not supported yet by this plugin)
+> - `@testing-library/svelte` (not supported yet by this plugin)
```
gndelia, line 7: Should we include ReactTestUtils.act from `react-dom/test-utils` ? It was mentioned in the original issue

docs/rules/no-unnecessary-act.md
Belco90: I've removed the duplicated block and added a simple comment about `'react-dom/test-utils'`. Thanks!

docs/rules/no-unnecessary-act.md
gndelia: ok no, it is not part of Testing Library packages 😅 , but it should work the same according to the issue

Belco90: `react-dom/test-utils` is mentioned in the rule doc, reported by the rule, and covered in test cases.

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -614,6 +624,91 @@ export function detectTestingLibraryUtils<
...
+     * - it's related to Testing Library module (this depends on Aggressive Reporting)
+     * - it's related to React DOM Test Utils
+     */
+    const isActUtil = (node: TSESTree.Identifier): boolean => {
+      const isTestingLibraryAct = isPotentialTestingLibraryFunction(
+        node,
+        (identifierNodeName, originalNodeName) => {
+          return [identifierNodeName, originalNodeName]
+            .filter(Boolean)
+            .includes('act');
```
gndelia, line 640:
```suggestion
          return [identifierNodeName, originalNodeName]
            .includes('act');
```

Minor suggestion, but I think you can skip the filtering as using includes should be enough (all falsey values are different from `'act'`)

Belco90: Yeah, although I need to refactor all this mechanism in a separate PR as mentioned in [this comment](https://github.com/testing-library/eslint-plugin-testing-library/pull/365/files#r626109283). So not sure if worth it, I just added for consistency with the other `is*Util` methods to be honest.

tests/lib/rules/no-unnecessary-act.test.ts
```diff
@@ -0,0 +1,782 @@
...
+        act(function() {
+          return waitFor();
+        });
+
+        act(() => waitFor());
+      });
+      `,
+    },
+    {
+      code: `// case: RTL act wrapping both RTL and non-RTL calls
```
zaicevas, line 94: If I understand correctly, the exception for allowing both RTL+non-RTL calls is made because of possible use cases that'd be blocked if this is reported as an error. 

On the other hand, I can't think of any such use case and I haven't yet stumbled upon one in my experience, so what I worry about is that users might take this wrong and use this exception to "workaround" the error for regular use cases, when `act()` is indeed unnecessary. This brings the question: if the use cases for RTL+non-RTL calls in `act()` are rare, why not report an error and let users block disable it?

tests/lib/rules/no-unnecessary-act.test.ts
```diff
@@ -0,0 +1,782 @@
...
+
+      test('valid case', async () => {
+        act(() => {
+          render(element);
+          stuffThatDoesNotUseRTL();
+        });
+
+        await act(async () => {
+          waitFor();
+          stuffThatDoesNotUseRTL();
```
zaicevas, line 105: `await stuffThatDoesNotUseRTL();` This would make the test fail. Is it intended?

lib/node-utils/index.ts
```diff
@@ -510,3 +532,95 @@ export function hasImportMatch(
...
+ * Note that comments don't affect the check.
+ *
+ * If node given is not a function, `false` will be returned.
+ */
+export function isEmptyFunction(node: TSESTree.Node): boolean | undefined {
+  if (ASTUtils.isFunction(node) && isBlockStatement(node.body)) {
+    return node.body.body.length === 0;
+  }
+
+  return false;
```
gndelia, line 574:
```suggestion
  return ASTUtils.isFunction(node) && isBlockStatement(node.body) && node.body.body.length === 0
```

Minor, but I think the if can be simplified just to

lib/rules/no-unnecessary-act.ts
```diff
@@ -0,0 +1,141 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Disallow wrapping Testing Library utils or empty callbacks in `act`',
+      category: 'Possible Errors',
+      recommendedConfig: {
+        dom: false,
+        angular: false,
+        react: false, // this should be enabled on v5 of the plugin
```
gndelia, line 25: maybe we could create a ticket/discussion for V5, so this kind of stuff doesn't get lost in comments. Not because there is any release date in mind, but as a planning  (and documenting) session

If you agree and do that, then this comment can be removed

lib/node-utils/index.ts
Belco90: Not sure if more readable tho 😅

tests/lib/rules/no-unnecessary-act.test.ts
zaicevas: On top of that, this test case doesn't pass, although I think that's valid:

```
        await act(async () => {
          await stuffThatDoesNotUseRTL();
        });
```

tests/lib/rules/no-unnecessary-act.test.ts
```diff
@@ -0,0 +1,782 @@
...
+ */
+ruleTester.run(RULE_NAME, rule, {
+  valid: [
+    {
+      code: `// case: RTL act wrapping non-RTL calls
+      import { act } from '@testing-library/react'
+
+      test('valid case', async () => {
+        act(() => {
+          stuffThatDoesNotUseRTL();
```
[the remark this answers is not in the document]
Belco90: That's weird, I'll check it.

tests/lib/rules/no-unnecessary-act.test.ts
```diff
@@ -0,0 +1,782 @@
...
+
+/**
+ * - AGR stands for Aggressive Reporting
+ * - RTL stands for React Testing Library (@testing-library/react)
+ * - RTU stands for React Test Utils (react-dom/test-utils)
+ */
+ruleTester.run(RULE_NAME, rule, {
+  valid: [
+    {
+      code: `// case: RTL act wrapping non-RTL calls
```
gndelia, line 14: seeing this code makes me wonder if it is worth (not for this PR tho) adding a property to the test case that contains a "description" or something (that will be added as the test description when running)

tests/lib/rules/no-unnecessary-act.test.ts
```diff
@@ -0,0 +1,782 @@
...
+
+        act(() => waitFor());
+
+        act(() => {})
+        await act(async () => {})
+        act(function() {})
+      });
+      `,
+    },
+  ],
```
gndelia, line 148: in my current project, we are using promises instead of async/await, maybe it's worth adding a case that considers using `.then()` instead of `await` to add more confidence

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: [I double-check it in the original issue](https://github.com/testing-library/eslint-plugin-testing-library/issues/259#issuecomment-803413239). Not really strong opinion about it, but we thought it could be useful not to report it.

tests/lib/rules/no-unnecessary-act.test.ts
```diff
@@ -0,0 +1,782 @@
...
+        act(() => {
+          fireEvent.click(el);
+        });
+
+        await act(async () => {
+          waitFor(() => {});
+        });
+
+        await act(async () => {
+          waitForElementToBeRemoved(el);
```
gndelia, line 166: Minor, but given the `async` identifier in the function these two should probably have the `await` (though it is not relevant for the test)

```suggestion
          await waitFor(() => {});
        });

        await act(async () => {
          await waitForElementToBeRemoved(el);
```

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: Mmm not really. Fail in terms of reporting an unexpected error or throwing an exception?

lib/rules/no-unnecessary-act.ts
Belco90: Yes! I have some notes written about things to do in a future v5. I'll try to create a proper project soon, so I'm gonna remove this comment.

gndelia (approved):
This looks great. I left some minor suggestions, but LGTM. Kudos to @alessbell  and @Belco90 ! 🚀

tests/lib/rules/no-unnecessary-act.test.ts
zaicevas: > Mmm not really. Fail in terms of reporting an unexpected error or throwing an exception?

Reports unexpected error (`no-unnecessary-act`), if you're referring to my 1st comment

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: I've tried to do that in the past but miserably failed. Definitely interested in giving this a try again, but I think ESLint itself is working into adding this feature.

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: Sure. Just out of curiosity: are you chaining several `.then()` in those tests? If so, doesn't that make the code really messy?

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: As you mentioned, it's not relevant for the test so I don't tend to put attention to that 😅

tests/lib/rules/no-unnecessary-act.test.ts
gndelia: I think this should work https://eslint.org/docs/developer-guide/nodejs-api#customizing-ruletester

what do you think? I can create a ticket and work on it at some point in the next week (after all I don't think this is high priority so it matches my current free time haha)

gndelia: I'm assuming the method parameter will have the info I need 😅

tests/lib/rules/no-unnecessary-act.test.ts
gndelia: I just joined the project and the team prefer a more functional approach chaining `.then()` as they think async/await looks more imperative. If well chained, it does not look messy 😛

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: That's the customization I tried to do in the past but didn't work. It was back when still using ESLint v6 so maybe it's possible now (or I just didn't find how to do it properly). Sounds like a plan! Create the issue and give it a try 👌

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: Not arguing against it, but I can't imagine writing tests chaining `.then()`! Anyway, please share any feedback you have if find anything to improve around thenables promises.

tests/lib/rules/no-unnecessary-act.test.ts
gndelia: yeah I went to their docs and there's no useful information to retrieve it https://github.com/eslint/eslint/blob/c981fb1994cd04914042ced1980aa86b68ba7be9/lib/rule-tester/rule-tester.js#L881 here is what it is being called.

I think overriding `describe` in [here](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/tests/lib/test-utils.ts#L31) just before calling `super.run` would allow me to iterate the rules and call the tests ( I have the reference from the `run` method. I would essentially re-implementing those ~10 lines from eslint test runner but it should work

tests/lib/rules/no-unnecessary-act.test.ts
gndelia: I will! the project is fairly new (startup mode 😅 ) so we are not using RTL yet, but when the time comes, this plugin will be added and feedback be gathered 💪

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: This is an interesting invalid usage! @kentcdodds it seems it would be a good idea to report also mixed RTL and non-RTL calls.

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: I got it. Basically, I ignoring`AwaitExpression` and `AssigmentExpression`. Potentially, [all different expressions listed here](https://eslint-utils.mysticatea.dev/api/ast-utils.html#hassideeffect) would be checked here, but I think the current list is enough.

lib/node-utils/index.ts
```diff
@@ -510,3 +533,115 @@ export function hasImportMatch(
...
+        return expression.right;
+      }
+
+      if (
+        ASTUtils.isAwaitExpression(expression.right) &&
+        isCallExpression(expression.right.argument)
+      ) {
+        return expression.right.argument;
+      }
+    }
```
Belco90, line 564: @zaicevas these are the new checks for the cases you mentioned. There must be a better way of finding those calls in the tree, this seems a little bit brittle but I guess it works for now.

tests/lib/rules/no-unnecessary-act.test.ts
Belco90: Not really sure what to do with this finally. Completely in favor of reporting any `act` with some react-testing-library util call in it, even if mixed with non-testing-library calls. Not sure if there is a really obvious exception to implement or some note to add to the doc about mixed calls.

zaicevas: How are you planning to release this rule? Is it gonna be a part of recommended ruleset with minor bump?

Belco90: Minor bump but without including the rule in any existing shareable config, that would be a breaking change for sure.

zaicevas: How about reporting mixed RTL and non-RTL calls, assuming that's what users expect, unless someone creates an issue, after which we can reconsider?

Belco90: Sure, we can do that.

zaicevas: Wait, but the currently released version **does not** report mixed calls, right?

Belco90: Oh it doesn't. I understood we wanted to leave it unreported and see how it goes. My bad.

Belco90: I'm gonna create an issue for adding a new `isStrict` option to this rule so it reports all `act` with an RTL call inside, no matter if mixed with non-RTL or not.

PR #374, opened 2021-05-09
fix(await-async-query): get correct Identifier related to CallExpression

### LINKED ISSUE #359: False positive for `testing-library/await-async-query`?
## Problem

I upgraded from version `3.10.2` to version `4.1.2` and am now getting errors for `await-async-query` that I cannot really explain. I read the migration guide to v4 and as far as I can see there is nothing wrong with my code. So I thought I would open this issue. 

This is a minimal example of the issue, it popped up all over my codebase

```ts
// import from custom-module
import { render, screen } from 'mocks/test-utils'
import userEvent from '@testing-library/user-event'

// I have an object of reusable testdata
const testData = {
  name: 'Theodore Tester',
  email: 'tester@parker.com',
  password: 'extremeSecret',
}

// ... and one of reusable selectors
const selectors = {
  username: () => screen.findByRole('textbox', { name: /username/i }),
  email: () => screen.findByRole('textbox', { name: /e-mail/i }),
  password: () => screen.findByLabelText(/password/i),
}

// which I use in a test
test('when a valid form is submitted, then a success notification is displayed', async () => {
  render(<CreateNewAdminUserPage />)
  userEvent.type(await selectors.username(), testData.name)
  userEvent.type(await selectors.email(), testData.email)
  userEvent.type(await selectors.password(), testData.password)

  // ... more test code and assertions
})
```

With `4.1.2` I get linting errors  for `await-async-query` **on my testdata**, but only on some of it:

CleanShot 2021-04-26 at 10 16 18@2x

CleanShot 2021-04-26 at 10 16 52@2x

As you can see in the screenshot there is not linting error on `testData.name`, but there are on `testData.email` and `testData.password`. 

What I cannot figure out is:

1. Why I see the error in the first place. To my understanding I handled the promises as expected in the rule
2. Why the error is indicated on my `testData`-object, which does not use promises
3. Why it is only reported for some of its properties 

## What I tried

* Switch off all aggressive reporting as described here: https://github.com/testing-library/eslint-plugin-testing-library#switching-all-aggressive-reporting-mechanisms-off => no change
* Configuring my custom-utils-module as described here: https://github.com/testing-library/eslint-plugin-testing-library#testing-libraryutils-module => no change
* Switch to `getByRole`-Queries instead of `findByRole` => fixes the issue, but is not really what I want

## Versions/Config

```sh
npm ls eslint-plugin-testing-library
-- eslint-plugin-testing-library@4.1.2
```

Config

```js
{
      files: ['*.test.{tsx,ts}'],
      extends: ['plugin:jest-dom/recommended', 'plugin:testing-library/react', 'plugin:jest/recommended'],
      plugins: ['jest-dom', 'testing-library', 'jest'],
      rules: {
        'jest/valid-title': ['off'],
        'jest/expect-expect': [
          'error',
          {
            assertFunctionNames: ['expect', 'assert*'],
          },
        ],
        'testing-library/custom-queries': ['off'],
        'testing-library/utils-module': ['off'],
        'testing-library/custom-renders': ['off'],
      },
  },
```

Thank you for your help!

Belco90:
Closes #359

### DIFF
--- a/lib/rules/await-async-query.ts
+++ b/lib/rules/await-async-query.ts
@@ -1,6 +1,7 @@
 import { ASTUtils, TSESTree } from '@typescript-eslint/experimental-utils';
 import {
   findClosestCallExpressionNode,
+  getDeepestIdentifierNode,
   getFunctionName,
   getInnermostReturningFunction,
   getVariableReferences,
@@ -27,9 +28,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
       },
     },
     messages: {
-      awaitAsyncQuery: 'promise returned from {{ name }} query must be handled',
+      awaitAsyncQuery:
+        'promise returned from `{{ name }}` query must be handled',
       asyncQueryWrapper:
-        'promise returned from {{ name }} wrapper over async query must be handled',
+        'promise returned from `{{ name }}` wrapper over async query must be handled',
     },
     schema: [],
   },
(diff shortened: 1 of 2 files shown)


lib/rules/await-async-query.ts
```diff
@@ -46,10 +48,16 @@ export default createTestingLibraryRule<Options, MessageIds>({
     }
 
     return {
-      'CallExpression Identifier'(node: TSESTree.Identifier) {
```
Belco90, line 49: The problem was here. This is getting an `Identifier` n levels within a `CallExpression`, which can lead to an `Identifier` not related at all to the `CallExpression`. Fixed just using `getDeepestIdentifierNode` so we get the Identifier of the `CallExpression` we are interested in.

PR #385, opened 2021-05-14
chore: migrate codebase ESLint config

### LINKED ISSUE #384: Migrate codebase ESLint config
Our current ESLint config is based on `eslint-config-standard`, which have the following issues:

- it hasn't been updated recently, which leads to inconsistent peer deps range. This is causing peer deps issues when working with npm v7, which is preventing us from running the CI with node v16 (see #350)
- this config aims to fix style issues too, which we are not interested in since we use prettier

In order to improve this, it would be better to switch to [`eslint-config-kentcdodds`](https://github.com/kentcdodds/eslint-config-kentcdodds), fixing whichever new code violations we get from this new config.

Additionally, it would be nice to address the following changes too:
- Forbid `@typescript-eslint/experimental-utils/dist/ts-eslint` imports with [`no-restricted-imports`](https://eslint.org/docs/rules/no-restricted-imports). We shouldn't import from `/dist` subfolder, but we have some of them across the codebase (I guess from IDE auto imports). Instead, we need to import the corresponding object from `@typescript-eslint/experimental-utils`.
- Install and setup `recommended` rules from [`eslint-plugin-unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn). It has really nice rules that would make our codebase more consistent. Most of them are autofixable.

**Changes to address**
- [x] replace `eslint-config-standard` by `eslint-config-kentcdodds`
- [x] forbid `@typescript-eslint/experimental-utils/dist` imports with `no-restricted-imports`
- [ ] ~~add `eslint-plugin-unicorn` to the codebase~~
- [x] fix issues generated by previous changes

Belco90:
Closes #384

This PR:
- replace eslint-config-standard by eslint-config-kentcdodds
- forbid @typescript-eslint/experimental-utils/dist imports with no-restricted-imports
- **does not** add eslint-plugin-unicorn to the codebase: it has several rules conflicting others, and some expecting to use node 16
- updates prettier
- fix issues generated by previous changes


lib/node-utils/index.ts
```diff
@@ -235,25 +234,26 @@ export function isPromiseHandled(nodeIdentifier: TSESTree.Identifier): boolean {
...
 
 export function getVariableReferences(
-  context: RuleContext<string, []>,
+  context: TSESLint.RuleContext<string, []>,
   node: TSESTree.Node
 ): TSESLint.Scope.Reference[] {
   return (
     (ASTUtils.isVariableDeclarator(node) &&
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
       context.getDeclaredVariables(node)[0]?.references?.slice(1)) ||
```
Belco90, line 243: I don't know which condition is the rule complaining here.

lib/rules/prefer-wait-for.ts
```diff
@@ -119,6 +120,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
           const [arg] = callExpressionNode.arguments;
           const fixers = [];
 
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
Belco90, line 123: I don't know which condition is the rule complaining here since `arg` can be emtpy.

package.json
```diff
@@ -50,19 +50,19 @@
     "@typescript-eslint/parser": "^4.21.0",
     "cpy-cli": "^3.1.1",
     "eslint": "^7.24.0",
-    "eslint-config-prettier": "^8.1.0",
-    "eslint-config-standard": "^16.0.2",
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-jest": "^24.3.4",
+    "eslint-config-kentcdodds": "^18.0.1",
```
MichaelDeBoey, line 53: https://github.com/kentcdodds/eslint-config-kentcdodds/releases/tag/v18.1.0

```suggestion
    "eslint-config-kentcdodds": "^18.1.0,
```

MichaelDeBoey (changes requested):

lib/node-utils/index.ts
MichaelDeBoey: I think `slice` is always possible, as `references` will always be string/array?

.eslintrc.json
```diff
@@ -22,8 +23,37 @@
...
+      }
+    ],
+
+    // Import
+    "import/order": [
+      "warn",
+      {
+        "groups": ["builtin", "external", "parent", "sibling", "index"],
+        "pathGroups": [
+          { "pattern": "@typescript-eslint/**", "group": "external" }
```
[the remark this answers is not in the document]
Belco90: Yes, you are right! But `eslint-plugin-import` was assuming this was an internal, not sure why, so I had to add this. Could be a bug in the plugin? I couldn't find any reason for assuming `@typescript-eslint/**` was internal.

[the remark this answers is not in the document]
MichaelDeBoey: Still having the issue after updating all the dependencies to latest?

If not, let's remove it.
Otherwise, I think it would be best to create a small reproduction and issue a bug report upstream

Belco90:
I'm getting a new error **without** updating any dependency now 🤔 

> Error: Error while loading rule 'import/no-import-module-exports': Cannot find module '/home/runner/work/eslint-plugin-testing-library/eslint-plugin-testing-library/index.js'. Please verify that the package.json has a valid "main" entry
Occurred while linting /home/runner/work/eslint-plugin-testing-library/eslint-plugin-testing-library/commitlint.config.js

And we are getting [this error now too on CI](https://github.com/testing-library/eslint-plugin-testing-library/pull/385/checks?check_run_id=2604759099). I don't know what's causing that, the CI was passing yesterday (I've checked the lint command locally without changing anything from yesterday and it fails too).

@MichaelDeBoey any idea?

.eslintrc.json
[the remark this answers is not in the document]
Belco90: Yes, actually I started getting that weird sort **after** upgrading to the latest `eslint-plugin-import` version. I see few issues open in their repo regarding the 2.23 version. I'm gonna try downgrading the package.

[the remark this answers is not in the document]
Belco90: I just figured it out: [`import type`s are sorted after regular imports since 2.23.0](https://github.com/benmosher/eslint-plugin-import/pull/2021)

I'm gonna remove my custom config for including the package as internal since that wasn't the problem.

lib/node-utils/index.ts
Belco90: That seems to be the reason indeed! However, there are two optional chainings in the statement that could return a falsy value, and that's where the OR fallback would be necessary.

The rule doesn't seem to be aware of this? It seems by its type the rule assumes `context.getDeclaredVariables(node)` will return an array with at least one element, so `context.getDeclaredVariables(node)[0]` will always exist, but that's not the case.

I'm not sure now how "empty arrays" can be typed in TS, but I'd say that's the problem of this rule?

lib/rules/prefer-wait-for.ts
Belco90: The same explanation from https://github.com/testing-library/eslint-plugin-testing-library/pull/385#discussion_r633902429 could explain this then.

lib/node-utils/index.ts
Belco90: It seems this is the reason indeed: https://github.com/typescript-eslint/typescript-eslint/issues/1798

I'm not if I want to enable this rule then. It could cause introducing errors by removing necessary conditions involving arrays.

lib/node-utils/index.ts
```diff
@@ -235,25 +234,26 @@ export function isPromiseHandled(nodeIdentifier: TSESTree.Identifier): boolean {
...
-    (ASTUtils.isVariableDeclarator(node) &&
-      context.getDeclaredVariables(node)[0]?.references?.slice(1)) ||
-    []
-  );
+  if (ASTUtils.isVariableDeclarator(node)) {
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    return context.getDeclaredVariables(node)[0]?.references?.slice(1) ?? [];
+  }
+
+  return [];
```
MichaelDeBoey, line 245: Maybe do an early return instead?

```suggestion
  if (!ASTUtils.isVariableDeclarator(node)) {
    return [];
  }

  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
  return context.getDeclaredVariables(node)[0]?.references?.slice(1) ?? [];
```

I normally don't like negative conditions, but I'm more in favor of having the "main" thing at the end

Belco90: That would fix it here but could potentially introduce errors in new code, so I've just switched off this rule until there is an option to ignore arrays within conditions.

MichaelDeBoey: I didn't suggest it because of the disabling of ESLint, but because

> I'm more in favor of having the "main" thing at the end

Belco90: Ah sorry, didn't realize that. I don't see any benefit in this case, I guess it's a matter of taste.

MichaelDeBoey (approved):

PR #388, opened 2021-05-18
feat(prefer-screen-queries): detect render in wrappers correctly

### LINKED ISSUE #367: False negative for testing-library/prefer-screen-queries
Hey!

When handling the new failures in our projects due to adding `testing-library/prefer-screen-queries`, I noticed some patterns weren't noticed by the rule. Strangely other similar patterns were noticed.

This pattern is when returning the result of `render`, and then destructuring in the caller, ie some variations of:
```js
function setup() {
  return render(<Foo />);
}

// Then in a test
const { getByText } = setup();
getByText("foo");
```

I couldn't find what difference these cases, so I'll rather pinpoint existing code in our project.

This was noticed by the rule:
https://github.com/firefox-devtools/profiler/blob/1c5dfed055f864413b3a62e43ca2d27b074d6108/src/test/components/TrackContextMenu.test.js#L64-L73
And everything in this file
https://github.com/firefox-devtools/profiler/blob/1c5dfed055f864413b3a62e43ca2d27b074d6108/src/test/components/ZipFileTree.test.js#L56-L70
And everything in this file

This wasn't noticed by the rule:
https://github.com/firefox-devtools/profiler/blob/1c5dfed055f864413b3a62e43ca2d27b074d6108/src/test/components/ButtonWithPanel.test.js#L80-L82
(and below in the same file)
https://github.com/firefox-devtools/profiler/blob/1c5dfed055f864413b3a62e43ca2d27b074d6108/src/test/components/CallNodeContextMenu.test.js#L107-L112

There are other examples, please tell me if you want me to give you all cases.

Initially I thought this was because of the somewhat complex setup, but then I was surprised to see that some cases were actually caught by the rule...

Thanks!

Belco90:
Closes #367

### DIFF
--- a/lib/rules/prefer-screen-queries.ts
+++ b/lib/rules/prefer-screen-queries.ts
@@ -2,6 +2,7 @@ import { ASTUtils, TSESTree } from '@typescript-eslint/experimental-utils';
 
 import { createTestingLibraryRule } from '../create-testing-library-rule';
 import {
+  getDeepestIdentifierNode,
   isCallExpression,
   isMemberExpression,
   isObjectExpression,
@@ -78,6 +79,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
       }
     }
 
+    function isIdentifierAllowed(name: string) {
+      return ['screen', ...withinDeclaredVariables].includes(name);
+    }
+
     // keep here those queries which are safe and shouldn't be reported
     // (from within, from render + container/base element, not related to TL, etc)
     const safeDestructuredQueries: string[] = [];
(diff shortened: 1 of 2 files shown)


tests/lib/rules/prefer-screen-queries.test.ts
```diff
@@ -413,5 +413,80 @@ ruleTester.run(RULE_NAME, rule, {
           ],
         } as const)
     ),
+    {
+      code: ` // issue #367 - example A
```
Belco90, line 417: Two scenarios provided by @julienw added as test cases.

lib/rules/prefer-screen-queries.ts
```diff
@@ -113,52 +136,49 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
-          isCallExpression(node.parent.object) &&
-          ASTUtils.isIdentifier(node.parent.object.callee) &&
-          node.parent.object.callee.name !== 'within' &&
-          helpers.isRenderUtil(node.parent.object.callee) &&
-          !usesContainerOrBaseElement(node.parent.object)
-        ) {
-          reportInvalidUsage(node);
-          return;
-        }
+        if (isMemberExpression(identifierNode.parent)) {
```
MichaelDeBoey, line 158: Can we do early returns instead of these nested if statements?

MichaelDeBoey (changes requested):

lib/rules/prefer-screen-queries.ts
Belco90: I'm afraid not since there are different strategies to follow depending on parent type.

MichaelDeBoey: What do you think of
```ts
        if (!isMemberExpression(identifierNode.parent)) {
          const isSafeDestructuredQuery = safeDestructuredQueries.some(
            (queryName) => queryName === identifierNode.name
          );
          if (isSafeDestructuredQuery) {
            return;
          }
          
          return reportInvalidUsage(identifierNode);
        }

        const memberExpressionNode = identifierNode.parent;
        if (
          isCallExpression(memberExpressionNode.object) &&
          ASTUtils.isIdentifier(memberExpressionNode.object.callee) &&
          memberExpressionNode.object.callee.name !== 'within' &&
          isReportableRender(memberExpressionNode.object.callee) &&
          !usesContainerOrBaseElement(memberExpressionNode.object)
        ) {
          return reportInvalidUsage(identifierNode);
        }

        if (
          ASTUtils.isIdentifier(memberExpressionNode.object) &&
          !isIdentifierAllowed(memberExpressionNode.object.name)
        ) {
          return reportInvalidUsage(identifierNode);
        }
```

MichaelDeBoey: It's not perfect, but it has less nested if statements, which is already a good thing I think

Belco90: Updated. I think the code seems simple in terms of nested if statements, but it's not so clear which are the different paths regarding the parent type.

package.json
```diff
@@ -52,7 +52,7 @@
     "eslint": "^7.26.0",
     "eslint-config-kentcdodds": "^18.1.0",
     "eslint-config-prettier": "^8.3.0",
-    "eslint-plugin-import": "^2.23.2",
+    "eslint-plugin-import": "^2.23.3",
```
Belco90, line 55: I have to update this version since it was generating a different formatting 😕

MichaelDeBoey (approved):

lib/rules/prefer-screen-queries.ts
```diff
@@ -113,52 +136,54 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
-        if (!helpers.isBuiltInQuery(node)) {
+        if (!isMemberExpression(identifierNode.parent)) {
+          const isSafeDestructuredQuery = safeDestructuredQueries.some(
+            (queryName) => queryName === identifierNode.name
+          );
+          if (isSafeDestructuredQuery) {
+            return;
+          }
+
+          reportInvalidUsage(identifierNode);
```
MichaelDeBoey, line 166: Is there a reason why you're not returning this function call?

It's `void`, so doing `return;` or returning a function with a `void` return type is the same, no?

Belco90: Because of [`consistent-return` rule](https://eslint.org/docs/rules/consistent-return)
CleanShot 2021-05-22 at 22 50 38@2x

PR #390, opened 2021-05-22
fix: update dependencies

MichaelDeBoey:


package.json
```diff
@@ -40,31 +40,30 @@
     "semantic-release": "semantic-release"
   },
   "dependencies": {
-    "@typescript-eslint/experimental-utils": "^4.21.0"
+    "@typescript-eslint/experimental-utils": "^4.24.0"
```
Belco90, line 43: This change makes this PR be released as a patch version. Not a problem, but it's not reflected in the PR title so I'm not sure if you are aware of it.

package.json
```diff
@@ -40,31 +40,30 @@
...
+    "@types/jest": "^26.0.23",
+    "@typescript-eslint/eslint-plugin": "^4.24.0",
+    "@typescript-eslint/parser": "^4.24.0",
     "cpy-cli": "^3.1.1",
-    "eslint": "^7.26.0",
+    "eslint": "^7.27.0",
     "eslint-config-kentcdodds": "^18.1.0",
     "eslint-config-prettier": "^8.3.0",
-    "eslint-plugin-import": "^2.23.2",
+    "eslint-plugin-import": "^2.23.3",
```
Belco90, line 55: This changes the order of the imports again. I'm really thinking about fixing exact dependencies versions in package.json to avoid unexpected behaviors as we got recently.

package.json
```diff
@@ -40,31 +40,30 @@
...
     "eslint-plugin-jest": "^24.3.6",
-    "eslint-plugin-jest-formatting": "^2.0.1",
+    "eslint-plugin-jest-formatting": "^3.0.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.4.0",
     "eslint-plugin-promise": "^5.1.0",
     "husky": "^4.3.8",
     "jest": "^26.6.3",
-    "jest-environment-jsdom": "^25.5.0",
```
Belco90, line 63: Oh, so this is not necessary anymore. I don't know why, but great news.

Belco90 (approved):

package.json
MichaelDeBoey: It was a bug in the plugin, which is now fixed.

The current import order is how it should have been the whole time

package.json
MichaelDeBoey: Yeah I figured now we have tests running on Node v16, that we could remove it

package.json
MichaelDeBoey: We could have kept the `chore` imo, as we're not really using anything new from the package.
Even if we did, the code would still be working as it was.

package.json
Belco90: This was actually failing for node v10 😅

package.json
Belco90: Yep, but that made other CI runs fail because they installed the latest version, ending in a different formating. Hence the idea of fixing exact versions in package.json

package.json
MichaelDeBoey: Than I don't know what was causing the issue 😅

But it's fixed so you don't hear me complaining 😉

package.json
Belco90: We need to let users know the deps have been updated. It could end up in a new bug, or just for letting them know about a new version is used under the hood for whatever reason they find it convenient.

package.json
MichaelDeBoey: This is kind of a special case as `v2.23.0` made a breaking change, which shouldn't have happened and is now fixed in `v2.23.3`

Belco90: But it can happen (especially with ESLint and its plugins), and happened a couple of times in last month. So we should probably fix ESLint and its plugins at least, for the same reason prettier is fixed to exact version.

MichaelDeBoey: I never had problems in the past with it, except now with `eslint-plugin-import` tbh, so not sure if we should really do that tbh.

We get a lot of benefits from not using exact version numbers imo

package.json
Belco90: > we're not really using anything new from the package.

This is a reason to avoid bumping this package so often by the way. We are not using anything in particular for the latest version, but we are forcing users using this plugin to bump such sub-dependency in their project, which could make them end up with several `@typescript-eslint/experimental-utils` because they have different version ranges. Not something really problematic, but at least to be aware of.

package.json
Belco90: What are the benefits tho? It's doing more harm than good lately.

PR #395, opened 2021-06-12
chore: Don't test with node 10

nickmccurdy:
**What**: Don't test with node 10

**Why**: Node 10 reached EOL on July 30th so let's free up some CI resources.

**How**: Remove 10 from node test matrix.

**Checklist**:

- ~[ ]~ Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)
- [x] Tests
- ~[ ]~ Typescript definitions updated
- [x] Ready to be merged


MichaelDeBoey:
I think it's best to wait until ESlint drops support for older versions before we do too

Ref: https://github.com/eslint/eslint/issues/14023

nickmccurdy:
It seems like ESLint has dropped it as of August 5th.

.github/workflows/pipeline.yml
```diff
@@ -51,7 +51,7 @@ jobs:
 
     strategy:
       matrix:
-        node: [10.12, 10, '12.0', 12, 14, '16.8']
```
nickmccurdy, line 54: Note that Node 16.9.1 fixes the regression in 16.9.0, but we still want to include newer versions of 16 in the future

.github/workflows/pipeline.yml
```diff
@@ -51,7 +51,7 @@ jobs:
...
+        node: [12, 14, 16]
```
nickmccurdy, line 54: Note that Node 16.9.1 fixes the regression in 16.9.0, but we still want to include newer versions of 16 in the future

nickmccurdy:
Does merging this depend on #462?

Belco90:
> Does merging this depend on #462?

That's right. We can point this MR to the alpha branch for the next major, right? @MichaelDeBoey

MichaelDeBoey:
Going to close this one as #463 already took care of it.

PR #397, opened 2021-06-21
refactor: automatically generate the rules-list in README

MichaelDeBoey:


README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+
+**Configurations**: ![dom-badge][] = dom, ![angular-badge][] = angular, ![react-badge][] = react, ![vue-badge][] = vue
+
+| Name                                                                                                 | Description                                                                          | 🔧  | Included in configurations                                        |
+| ---------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------ | --- | ----------------------------------------------------------------- |
+| [`testing-library/await-async-query`](./docs/rules/await-async-query.md)                             | Enforce promises from async queries to be handled                                    |     | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |
+| [`testing-library/await-async-utils`](./docs/rules/await-async-utils.md)                             | Enforce promises from async utils to be handled                                      |     | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |
```
MichaelDeBoey, line 190: Description was
> Enforce async utils to be awaited properly

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/await-fire-event`](./docs/rules/await-fire-event.md)                               | Enforce promises from fire event methods to be handled                               |     | ![vue-badge][]                                                    |
+| [`testing-library/consistent-data-testid`](./docs/rules/consistent-data-testid.md)                   | Ensures consistent usage of `data-testid`                                            |     |                                                                   |
```
MichaelDeBoey, line 192: Description was
> Ensure `data-testid` values match a provided regex.

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/no-await-sync-events`](./docs/rules/no-await-sync-events.md)                       | Disallow unnecessary `await` for sync events                                         |     |                                                                   |
+| [`testing-library/no-await-sync-query`](./docs/rules/no-await-sync-query.md)                         | Disallow unnecessary `await` for sync queries                                        |     | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |
+| [`testing-library/no-container`](./docs/rules/no-container.md)                                       | Disallow the use of container methods                                                |     | ![angular-badge][] ![react-badge][] ![vue-badge][]                |
```
MichaelDeBoey, line 195: Description was
> Disallow the use of `container` methods

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/no-manual-cleanup`](./docs/rules/no-manual-cleanup.md)                             | Disallow the use of `cleanup`                                                        |     |                                                                   |
+| [`testing-library/no-node-access`](./docs/rules/no-node-access.md)                                   | Disallow direct Node access                                                          |     | ![angular-badge][] ![react-badge][] ![vue-badge][]                |
+| [`testing-library/no-promise-in-fire-event`](./docs/rules/no-promise-in-fire-event.md)               | Disallow the use of promises passed to a `fireEvent` method                          |     | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |
+| [`testing-library/no-render-in-setup`](./docs/rules/no-render-in-setup.md)                           | Disallow the use of `render` in testing frameworks setup functions                   |     |                                                                   |
```
MichaelDeBoey, line 201: Description was
> Disallow the use of `render` in setup functions

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/no-unnecessary-act`](./docs/rules/no-unnecessary-act.md)                           | Disallow wrapping Testing Library utils or empty callbacks in `act`                  |     |                                                                   |
+| [`testing-library/no-wait-for-empty-callback`](./docs/rules/no-wait-for-empty-callback.md)           | It's preferred to avoid empty callbacks in `waitFor` and `waitForElementToBeRemoved` |     | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |
```
MichaelDeBoey, line 203: Description was
> Disallow empty callbacks for `waitFor` and `waitForElementToBeRemoved`

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/no-wait-for-multiple-assertions`](./docs/rules/no-wait-for-multiple-assertions.md) | It's preferred to avoid multiple assertions in `waitFor`                             |     |                                                                   |
```
MichaelDeBoey, line 204: Description was
> Disallow the use of multiple expect inside `waitFor`

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/no-wait-for-side-effects`](./docs/rules/no-wait-for-side-effects.md)               | It's preferred to avoid side effects in `waitFor`                                    |     |                                                                   |
```
MichaelDeBoey, line 205: Description was
> Disallow the use of side effects inside `waitFor`

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/no-wait-for-snapshot`](./docs/rules/no-wait-for-snapshot.md)                       | Ensures no snapshot is generated inside of a `waitFor` call                          |     |                                                                   |
+| [`testing-library/prefer-explicit-assert`](./docs/rules/prefer-explicit-assert.md)                   | Suggest using explicit assertions rather than just `getBy*` queries                  |     |                                                                   |
+| [`testing-library/prefer-find-by`](./docs/rules/prefer-find-by.md)                                   | Suggest using `find*` query instead of `waitFor` + `get*` to wait for elements       | 🔧  | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |
```
MichaelDeBoey, line 208: Description was
> Suggest using `findBy*` methods instead of the `waitFor` + `getBy` queries

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/prefer-presence-queries`](./docs/rules/prefer-presence-queries.md)                 | Ensure appropriate get*/query* queries are used with their respective matchers       |     |                                                                   |
```
MichaelDeBoey, line 209: Description was
> Enforce specific queries when checking element is present or not

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/prefer-screen-queries`](./docs/rules/prefer-screen-queries.md)                     | Suggest using screen while querying                                                  |     | ![dom-badge][] ![angular-badge][] ![react-badge][] ![vue-badge][] |
```
MichaelDeBoey, line 210: Description was
> Suggest using screen while using queries

README.md
```diff
@@ -178,33 +178,41 @@ To enable this configuration use the `extends` property in your
...
+| [`testing-library/prefer-user-event`](./docs/rules/prefer-user-event.md)                             | Suggest using userEvent over fireEvent                                               |     |                                                                   |
```
MichaelDeBoey, line 211: Description was
> Suggest using `userEvent` library instead of `fireEvent` for simulating user interaction

MichaelDeBoey (commented):
Seems like some descriptions changed now that we use the rules description directly.

Belco90 (changes requested):
Nice PR, it's fantastic to automate this process!

> Seems like some descriptions changed now that we use the rules description directly.
I would say that's not a problem. Let's leave the new ones and improve them later if necessary (but looks good to me).

One thing I'd like to request is to check the generated rules list in a test similar to how we check it for recommended presets.

MichaelDeBoey:
> One thing I'd like to request is to check the generated rules list in a test similar to how we check it for recommended presets.

Was thinking about this, but not really sure if we should do so **and** how would be the best to do this 🤔

Belco90:
> Was thinking about this, but not really sure if we should do so **and** how would be the best to do this 🤔

It could definitely be tricky with the formatting, but we need some way to automate this to avoid getting the rule list outdated. There are 2 options I can think of for now:

1. Add a test to check the rule list generated in a specific branch has the same content as the current rule list. Since we have a placeholder in the README to inject the rule list, we can retrieve the current rule list by regexp using those placeholders. We would have to [disable the auto-formatting with prettier in that block](https://prettier.io/docs/en/ignore.html#range-ignore).
2. Regenerate the rule list in CI/CD. We would need some workflow to update the rule list if necessary after a change is merged. This could be done by leaving the section empty and injecting the generated rule list in CI/CD in the build step

I prefer option 1 so we have more control about when to generate the rule list, and we also avoid injecting stuff at CI/CD.

MichaelDeBoey:
Looking at other repos, they don't have anything to test against this.
It's just a thing that needs to be run whenever you create a new rule.
So I wouldn't bother about testing this either tbh 🤷‍♂️

The only thing I would do now is go over all descriptions and see if we want to change them to something more meaningful (like some of them were already in the README list).
This can be done in a separate PR though.

Belco90:
> It's just a thing that needs to be run whenever you create a new rule.

Well, not just when creating a rule, but also when updating or deleting an existing one. Not something critical, but definitely easy to forget about.

Maybe we can add a check to the PR template to indicate "a rule was created/updated/deleted and rule list has been regenerated" or something like that.

MichaelDeBoey:
> Maybe we can add a check to the PR template to indicate "a rule was created/updated/deleted and rule list has been regenerated" or something like that.

Let's create another PR to update our PR template, where we can explain more in depth which things need to be done when creating/updating a rule, ...

Belco90:
I've created #400 for it.

Belco90 (approved):

README.md
Belco90: I think the previous one is better, can you change it in the rule description?

README.md
Belco90: I think the previous one is better, can you change it in the rule description?

README.md
Belco90: I think the previous one is better, can you change it in the rule description?

Belco90 (changes requested):
I've pointed some descriptions that I think were better before. Can we update those before going ahead?

Belco90 (approved):
Looks ok now, but it's tricky to review if you force-push after changes requested since I can't compare it against my last review 😞

PR #398, opened 2021-06-23
feat: add support for fireEvent as function in prefer-user-event

### LINKED ISSUE #261: prefer-user-event should consider the use case where fireEvent is used as a function
In addition to the method events like `fireEvent.click`, it is also possible to create custom events objects and use `fireEvent` as a function, as [the documentation for fireEvent shows](https://testing-library.com/docs/dom-testing-library/api-events/#createeventeventname)

```js
// form 1
const myEvent = createEvent.click(node)
fireEvent(node, myEvent)

// form 2
fireEvent(
  input,
  createEvent('click', input)
)
```

The rule should recommend to use `userEvent.click()` 

the rule should consider this form too

**Note**: This rule has been migrated in the `v4` branch, so anyone taking this issue should do so starting from branch `v4` (And targeting it as well)

gndelia:
closes #261   
It's been a while since the last time I worked in rules here and with eslint/AST. I was a bit out of shape 🤣  so it took me more than I originally expected. Any feedback is appreciated!

### DIFF
--- a/lib/create-testing-library-rule/detect-testing-library-utils.ts
+++ b/lib/create-testing-library-rule/detect-testing-library-utils.ts
@@ -627,25 +627,20 @@ export function detectTestingLibraryUtils<
         identifierNodeName: string,
         originalNodeName?: string
       ) => [identifierNodeName, originalNodeName].includes(CREATE_EVENT_NAME);
-      if (ASTUtils.isIdentifier(node)) {
-        return isPotentialTestingLibraryFunction(node, isCreateEventCallback);
-      }
-      if (ASTUtils.isIdentifier(node.callee)) {
-        return isPotentialTestingLibraryFunction(
-          node.callee,
-          isCreateEventCallback
-        );
-      }
-      if (!isMemberExpression(node.callee)) {
-        return false;
-      }
-      if (ASTUtils.isIdentifier(node.callee.object)) {
+      if (
+        isCallExpression(node) &&
+        isMemberExpression(node.callee) &&
+        ASTUtils.isIdentifier(node.callee.object)
+      ) {
         return isPotentialTestingLibraryFunction(
           node.callee.object,
           isCreateEventCallback
         );
       }
+
       if (
+        isCallExpression(node) &&
+        isMemberExpression(node.callee) &&
         isMemberExpression(node.callee.object) &&
         ASTUtils.isIdentifier(node.callee.object.property)
       ) {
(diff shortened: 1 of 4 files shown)


lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -595,6 +622,41 @@ export function detectTestingLibraryUtils<
...
+      if (!isMemberExpression(node.callee)) {
+        return false;
+      }
+      if (ASTUtils.isIdentifier(node.callee.object)) {
+        return isPotentialTestingLibraryFunction(
+          node.callee.object,
+          isCreateEventCallback
+        );
+      }
+      if (
```
Belco90, line 648: You don't need to manually check the type of the node and navigate its properties to get the deepest identifier! You can do that directly with `getDeepestIdentifierNode`.

Could you check how it's used in other parts of the code and simplify this function with it?

Belco90 (changes requested):
Happy to see you back @gndelia! Nice addition, I requested a small refactor to simplify one of the new functions but everything else looks god to me.

lib/create-testing-library-rule/detect-testing-library-utils.ts
gndelia: cool! I'll check it out. This function was modified a couple of times (I built it iteratively) so while I knew about that function I think that eventually the final form of this function was like this and I totally forgot about that function 😅

gndelia: @Belco90  I applied the suggestion; it doesn't solve all the cases but I was able to simplify two of those. Not in all cases  I want to get the deepest identifier node; see examples below

```ts
// here getDeepestIdentifierNode would retrieve the "click" node, but I need the "createEvent" node
const event = createEvent.click()

// Same here, I need "createEvent" but I'd get "click"
const event = dom.createEvent.click()

// here it works as expected
const event = dom.createEvent('click')

// here the same
const event = createEvent('click')
```

Applied here https://github.com/testing-library/eslint-plugin-testing-library/pull/398/commits/57622f0576989c7bd76a507269e196b24a2ae8be

Belco90 (approved):

lib/create-testing-library-rule/detect-testing-library-utils.ts
Belco90: Cool, it makes sense. At least it simplified traversing the node a bit.

Issue #399, opened 2021-06-24
New rule: reporting `getBy*` inside `waitForElementToBeRemoved`


zaicevas:
[Docs](https://testing-library.com/docs/guide-disappearance#waiting-for-disappearance) use `queryBy*` in examples of waiting for disappearance. AFAICS one could use either `getBy` or `queryBy` and both would work with `waitForElementToBeRemoved`. However, different error messages are displayed if the element is initially not present:

```
      // TestingLibraryElementError: Unable to find an element by: [data-testid="loader"]
      await waitForElementToBeRemoved(screen.getByTestId('loader'))

      // The element(s) given to waitForElementToBeRemoved are already removed. 
      // waitForElementToBeRemoved requires that the element(s) exist(s) before waiting for removal.
      await waitForElementToBeRemoved(screen.queryByTestId('loader'))
```

The error message when using `queryBy` is more descriptive. Additionally, it might be useful for the users to keep consistency in the codebase by using either `queryBy` or `getBy` when asserting disappearance.
**Therefore, we could have a new rule that would force using `queryBy` inside `waitForElementToBeRemoved`.**

Belco90:
Hey @zaicevas! This looks like a nice proposal for a new rule. I guess it could be named `prefer-wait-for-removed-query` or something similar (I don't have anything better in mind).

Would you like to work on this?

zaicevas:
> Would you like to work on this?

I'll take a look at it this weekend and evaluate how big of an effort it is going to be.

zaicevas:
Yes, I'll work on it. I've already started implementing the new rule.

Regarding the name - what about `prefer-query-by-disappearance` ?

Belco90:
> Regarding the name - what about `prefer-query-by-disappearance` ?

I like that name! 🚀

PR #402, opened 2021-07-04
feat: new `prefer-query-by-disappearance` rule

zaicevas:
Closes https://github.com/testing-library/eslint-plugin-testing-library/issues/399


Belco90 (changes requested):
Hey @zaicevas, thanks for your PR!

The implementation looks really great, it was pretty easy to follow. I just added some comments here and there to check a couple of things, but after that should be ready to go.

lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -0,0 +1,190 @@
...
+        argumentProperty = getPropertyIdentifierNode(node);
+      }
+
+      if (!argumentProperty) {
+        return false;
+      }
+
+      return (
+        helpers.isGetQueryVariant(argumentProperty) ||
+        helpers.isFindQueryVariant(argumentProperty)
```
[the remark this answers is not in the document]
zaicevas: There're already some examples with `findBy*`: https://github.com/testing-library/eslint-plugin-testing-library/pull/402/files#diff-49ba8c94275c7a8d748451b044ced483c64a12aea5dc6067cf5f556467f27938R22

lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -0,0 +1,190 @@
...
+export const RULE_NAME = 'prefer-query-wait-disappearance';
+export type MessageIds = 'preferQueryByDisappearance';
+
+export default createTestingLibraryRule<[], MessageIds>({
+  name: RULE_NAME,
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Disallow using getBy* query inside waitForElementToBeRemoved',
```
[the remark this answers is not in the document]
zaicevas: Yeah, I've added this description before writing the docs and forgot to update it 😅

lib/rules/prefer-query-by-disappearance.ts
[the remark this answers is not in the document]
Belco90: My bad!

Belco90 (approved):
Fantastic work, thanks for your contribution!

Belco90 (changes requested):
Sorry, one last change for an extra valid test.

tests/lib/rules/prefer-query-by-disappearance.test.ts
```diff
@@ -0,0 +1,741 @@
+import rule, {
+  RULE_NAME,
+} from '../../../lib/rules/prefer-query-by-disappearance';
+import { createRuleTester } from '../test-utils';
+
+const ruleTester = createRuleTester();
+
+ruleTester.run(RULE_NAME, rule, {
+  valid: [
```
Belco90, line 9: You can also pass directly an element to `waitForElementToBeRemoved`, can you add a test for that? E.g.

```js
import { screen } from '@testing-library/react';

const button = screen.getByRole('button')
// ...
await waitForElementToBeRemoved(button)
```

Belco90 (changes requested):

tests/lib/rules/prefer-query-by-disappearance.test.ts
zaicevas: There is a test for that: https://github.com/testing-library/eslint-plugin-testing-library/pull/402/files#diff-85beea9fec5d5f9c3167a9bb1ca393a7a5c511cf3d4fec7acbe4a5558c1d1a79R97

Or do you mean specifically when having the element in a separate variable?

zaicevas: And an equivalent invalid test: https://github.com/testing-library/eslint-plugin-testing-library/blob/71f541070b8f496ab1d6e6fc2f4ae1b184f6ab36/tests/lib/rules/prefer-query-by-disappearance.test.ts#L290

Belco90: I mean specifically having the element in a separate variable, so we cover how the rule behaves when a call expression is not found but an identifier.

Belco90: For clarification: the example I added doesn't have to be reported. If the rule finds an identifier as an argument of `waitForElementToBeRemoved` we can just assume is valid.

docs/rules/prefer-query-by-disappearance.md
```diff
@@ -0,0 +1,33 @@
+# Suggest using `queryBy*` queries when waiting for disappearance (`testing-library/prefer-query-by-disappearance`)
+
+## Rule Details
+
+This rule enforces using `queryBy*` queries when waiting for disappearance with `waitForElementToBeRemoved`.
+
+Using `queryBy*` queries in a `waitForElementToBeRemoved` yield more descriptive error messages and helps achieving more consistency in a codebase.
```
gndelia, line 7:
```suggestion
Using `queryBy*` queries in a `waitForElementToBeRemoved` yields more descriptive error messages and helps to achieve more consistency in a codebase.
```

minor spelling fixes (I'm not an English native speaker though, but Grammarly is marking these as errors 😅 )

docs/rules/prefer-query-by-disappearance.md
```diff
@@ -0,0 +1,33 @@
...
+
+```js
+// TestingLibraryElementError: Unable to find an element by: [data-testid="loader"]
+await waitForElementToBeRemoved(screen.getByTestId('loader'));
+
+// The element(s) given to waitForElementToBeRemoved are already removed.
+// waitForElementToBeRemoved requires that the element(s) exist(s) before waiting for removal.
+await waitForElementToBeRemoved(screen.queryByTestId('loader'));
```
gndelia, line 15: I thought `waitForElementToBeRemoved` only accepted callbacks, TIL!

tests/lib/rules/prefer-query-by-disappearance.test.ts
```diff
@@ -0,0 +1,749 @@
...
+      code: `
+        import { render } from '@testing-library/react';
+
+        const { queryByText } = render(<App />)
+        await waitForElementToBeRemoved(function() {
+          return queryByText("hello")
+        })
+      `,
+    },
+  ],
```
gndelia, line 200: would a scenario like (simplified) this work with the rule?

```ts
const callback = () => screen.queryByText('text')
await waitForElementToBeRemoved(callback)
```

If it works, would you mind adding some scenarios to test it?
If it doesn't, if it's not too much trouble we could add them. If it is complex, maybe we could tackle it in another issue. I think some rules check this kind of scenarios

Belco90: This is an interesting case, but I'm happy with the current implementation so it's ok for me if we want to tackle that in a separate PR if it's too much work to implement it now.

lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    function isReportableExpression(node: TSESTree.LeftHandSideExpression) {
+      let argumentProperty;
+
+      if (isMemberExpression(node)) {
+        argumentProperty = getPropertyIdentifierNode(node.property);
+      } else {
+        argumentProperty = getPropertyIdentifierNode(node);
+      }
```
MichaelDeBoey, line 59: This can be simplified

```suggestion
      const argumentProperty = isMemberExpression(node)
        ? getPropertyIdentifierNode(node.property)
        : getPropertyIdentifierNode(node);
```

lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -0,0 +1,184 @@
...
+    function isFunctionExpressionViolation(
+      node: TSESTree.CallExpressionArgument
+    ) {
+      if (!isFunctionExpression(node)) {
+        return false;
+      }
+
+      return node.body.body.some(
+        (statement) =>
+          isReturnViolation(statement) || isNonReturnViolation(statement)
```
MichaelDeBoey, line 118: nit: maybe extract this line into a `isStatementViolation` function?

lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -0,0 +1,184 @@
...
+  isCallExpression,
+  isMemberExpression,
+  isFunctionExpression,
+  isExpressionStatement,
+  isReturnStatement,
+  isBlockStatement,
+} from '../node-utils';
+
+export const RULE_NAME = 'prefer-query-wait-disappearance';
+export type MessageIds = 'preferQueryByDisappearance';
```
MichaelDeBoey, line 16: No need to export these, as they're not used anywhere else

```suggestion
const RULE_NAME = 'prefer-query-wait-disappearance';
type MessageIds = 'preferQueryByDisappearance';
```

lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export default createTestingLibraryRule<[], MessageIds>({
```
MichaelDeBoey, line 18: I think we should extract `Options` type, so it's consistent with all other rules
```suggestion
type Options = [];

export default createTestingLibraryRule<Options, MessageIds>({
```

MichaelDeBoey (changes requested):

lib/rules/prefer-query-by-disappearance.ts
zaicevas: `RULE_NAME` is being used here: https://github.com/testing-library/eslint-plugin-testing-library/pull/402/files#diff-85beea9fec5d5f9c3167a9bb1ca393a7a5c511cf3d4fec7acbe4a5558c1d1a79R2
But `MessageIds` is not - removed `export`.

lib/rules/prefer-query-by-disappearance.ts
zaicevas: Makes sense to me

tests/lib/rules/prefer-query-by-disappearance.test.ts
zaicevas: Seems like the same case as https://github.com/testing-library/eslint-plugin-testing-library/pull/402#discussion_r663524367

I am not sure where to start with this one, considering the callback might be imported from elsewhere 🤔 Anywho, I agree that it shouldn't be a blocker to merge this rule, as both of you mentioned.

gndelia (approved):
🚀

Belco90 (approved):

lib/rules/prefer-query-by-disappearance.ts
MichaelDeBoey: Am I missing something or is this link the wrong one? 🤔

lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -0,0 +1,183 @@
...
+    function isFunctionExpressionViolation(
+      node: TSESTree.CallExpressionArgument
+    ) {
+      if (!isFunctionExpression(node)) {
+        return false;
+      }
+
+      return node.body.body.some((statement) =>
+        isStatementViolation(statement)
+      );
```
MichaelDeBoey, line 119: This can be simplified to

```suggestion
      return node.body.body.some(isStatementViolation);
```

lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -0,0 +1,183 @@
...
+    function isArrowFunctionBodyViolation(
+      node: TSESTree.CallExpressionArgument
+    ) {
+      if (!isArrowFunctionExpression(node) || !isBlockStatement(node.body)) {
+        return false;
+      }
+
+      return node.body.body.some((statement) =>
+        isStatementViolation(statement)
+      );
```
MichaelDeBoey, line 131: Same

```suggestion
      return node.body.body.some(isStatementViolation);
```

MichaelDeBoey (approved):
2 small nit picks, but OK for me! 👍

lib/rules/prefer-query-by-disappearance.ts
zaicevas: I usually try to avoid this notation because of cases like this:
`const a = ['1', '2', '3'].map(parseInt) // [1, NaN, NaN]`

lib/rules/prefer-query-by-disappearance.ts
Belco90: It's the correct link but probably `tests/lib/rules/prefer-query-by-disappearance.test.ts` file is not loading so you need to click `Load diff` to be able to go the linked reference.

Issue #382, opened 2021-05-13
`no-unnecessary-act`: option for strict reporting


Belco90:
Since the current implementation of `no-unnecessary-act` doesn't report `act` usages with react-testing-library calls inside if mixed with non-react-testing-library, it would be interesting to add a new `isStrict` option (`false` by default) which makes the rule to always report react-testing-library calls inside `act` no matter if mixed with non-react-testing-library.

Belco90:
@zaicevas this is the issue I mentioned to tackle the strict reporting in `no-unnecessary-act`.

zaicevas:
@Belco90 do you mind if I'll try to implement it?

Belco90:
> @Belco90 do you mind if I'll try to implement it?

By all means! I won't have much spare time during summer, so I'd really appreciate if you can take care of this one.

PR #404, opened 2021-07-06
feat: isStrict option for no-unnecessary-act

zaicevas:
Closes https://github.com/testing-library/eslint-plugin-testing-library/issues/382

### DIFF
--- a/docs/rules/no-unnecessary-act.md
+++ b/docs/rules/no-unnecessary-act.md
@@ -87,6 +87,30 @@ await act(async () => {
 });
 ```
 
+## Options
+
+This rule has one option:
+
+- `isStrict`: wrapping both things related and not related to Testing Library in `act` is reported
+
+  ```js
+  "testing-library/no-unnecessary-act": ["error", {"isStrict": true}]
+  ```
+
+Incorrect:
+
+```jsx
+// ❌ wrapping both things related and not related to Testing Library in `act` is NOT correct
+
+import { act, screen } from '@testing-library/react';
+import { stuffThatDoesNotUseRTL } from 'somwhere-else';
+
+await act(async () => {
+  await screen.findByRole('button');
+  stuffThatDoesNotUseRTL();
+});
+```
+
 ## Further Reading
 
 - [Inspiration for this rule](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#wrapping-things-in-act-unnecessarily)
--- a/docs/rules/no-unnecessary-act.md
+++ b/docs/rules/no-unnecessary-act.md
@@ -91,7 +91,7 @@ await act(async () => {
 
 This rule has one option:
 
-- `isStrict`: wrapping both things related and not related to Testing Library in `act` is reported
+- `isStrict`: **disabled by default**. Wrapping both things related and not related to Testing Library in `act` is reported
 
   ```js
   "testing-library/no-unnecessary-act": ["error", {"isStrict": true}]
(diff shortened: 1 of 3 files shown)


lib/rules/no-unnecessary-act.ts
```diff
@@ -32,37 +33,71 @@ export default createTestingLibraryRule<[], MessageIds>({
         'Avoid wrapping Testing Library util calls in `act`',
       noUnnecessaryActEmptyFunction: 'Avoid wrapping empty function in `act`',
     },
-    schema: [],
+    schema: [
```
zaicevas, line 36: Added it for consistency, but not sure why `schema` is needed and how and where it's used. Can someone enlighten me? 😄

Belco90: You can find ESLint Options Schemas doc here: https://eslint.org/docs/developer-guide/working-with-rules#options-schemas

TL;DR ESLint will use that schema to validate configuration options and prevent invalid or unexpected inputs before  executing the rules:

ESLint config validation error displayed based on rule schema

docs/rules/no-unnecessary-act.md
```diff
@@ -87,6 +87,30 @@ await act(async () => {
 });
 ```
 
+## Options
+
+This rule has one option:
+
+- `isStrict`: wrapping both things related and not related to Testing Library in `act` is reported
```
Belco90, line 94: Could you specify this is `false` by default?

lib/rules/no-unnecessary-act.ts
Belco90: I just realized the option for `no-dom-import` is specified just as a `string` in the schema, but can be improved to indicate only `"angular"`, `"react"`, or `"vue"` are the valid options.

lib/rules/no-unnecessary-act.ts
```diff
@@ -89,7 +124,16 @@ export default createTestingLibraryRule<[], MessageIds>({
           node: identifierNode,
           messageId: 'noUnnecessaryActEmptyFunction',
         });
-      } else if (!hasSomeNonTestingLibraryCall(blockStatementNode.body)) {
+        return;
+      }
+
+      const isReported =
```
Belco90, line 130: I found this var name confusing! I thought it meant the code violation was already reported, but it actually indicates if it needs to be reported. Could you rename it to something like `shouldBeReported`?

lib/rules/no-unnecessary-act.ts
```diff
@@ -89,7 +124,16 @@ export default createTestingLibraryRule<[], MessageIds>({
...
+        (!isStrict && !hasSomeNonTestingLibraryCall(blockStatementNode.body)) ||
+        (isStrict &&
+          hasSomeNonTestingLibraryCall(blockStatementNode.body) &&
+          hasTestingLibraryCall(blockStatementNode.body));
```
Belco90, line 134: I guess this can be simplified like this? What do you think?

```suggestion
const isReported = isStrict
  ? hasSomeNonTestingLibraryCall(blockStatementNode.body) &&
    hasTestingLibraryCall(blockStatementNode.body)
  : !hasSomeNonTestingLibraryCall(blockStatementNode.body);
```

Belco90 (changes requested):
Thanks for your contribution again! This was a really quick one. Just some minor changes and should be ready.

Belco90 (approved):
Great work again! You are becoming a veteran of this codebase 😅

PR #410, opened 2021-07-18
fix(prefer-query-by-disappearance): fix rule name

MichaelDeBoey:


Belco90 (commented):
Oops, good catch!

Do we need to regenerate the rules in the README tho?

MichaelDeBoey:
@Belco90 These aren't used to generate the README list, so it's safe to merge

Belco90 (approved):
I see it's using the filename to generate the README info, not the rule name.

Issue #417, opened 2021-07-19
Bug: `no-unnecessary-act` reports non-RTL promise


zaicevas:
`no-unnecessary-act` with `isStrict` OFF reports a mocked promise that has nothing to with RTL:

```
        await act(async () => {
          await stuffThatDoesNotUseRTL; // reported
        });
```

PR #418, opened 2021-07-19
fix(no-unnecessary-act): fix false positives when `isStrict` option is set to `false`

zaicevas:
Closes https://github.com/testing-library/eslint-plugin-testing-library/issues/417

### DIFF
--- a/lib/node-utils/is-node-of-type.ts
+++ b/lib/node-utils/is-node-of-type.ts
@@ -49,4 +49,3 @@ export const isReturnStatement = isNodeOfType(AST_NODE_TYPES.ReturnStatement);
 export const isFunctionExpression = isNodeOfType(
   AST_NODE_TYPES.FunctionExpression
 );
-export const isAwaitExpression = isNodeOfType(AST_NODE_TYPES.AwaitExpression);
--- a/lib/node-utils/is-node-of-type.ts
+++ b/lib/node-utils/is-node-of-type.ts
@@ -49,3 +49,4 @@ export const isReturnStatement = isNodeOfType(AST_NODE_TYPES.ReturnStatement);
 export const isFunctionExpression = isNodeOfType(
   AST_NODE_TYPES.FunctionExpression
 );
+export const isAwaitExpression = isNodeOfType(AST_NODE_TYPES.AwaitExpression);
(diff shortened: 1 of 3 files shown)


lib/rules/no-unnecessary-act.ts
```diff
@@ -52,17 +56,30 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
-      if (!identifier) {
-        return null;
+      if (
+        isExpressionStatement(statement) &&
+        isAwaitExpression(statement.expression)
+      ) {
+        return getPropertyIdentifierNode(statement.expression.argument);
+      }
+
+      if (isReturnStatement(statement) && statement.argument) {
```
MichaelDeBoey, line 78:
```suggestion
      if (isReturnStatement(statement) && Boolean(statement.argument)) {
```

MichaelDeBoey (changes requested):

lib/rules/no-unnecessary-act.ts
zaicevas: That doesn't work. We could do `!!statement.argument`, but what's the reasoning behind it?

image

lib/node-utils/is-node-of-type.ts
```diff
@@ -49,3 +49,4 @@ export const isReturnStatement = isNodeOfType(AST_NODE_TYPES.ReturnStatement);
 export const isFunctionExpression = isNodeOfType(
   AST_NODE_TYPES.FunctionExpression
 );
+export const isAwaitExpression = isNodeOfType(AST_NODE_TYPES.AwaitExpression);
```
Belco90, line 52: Same as the other PR: this is available in `@typescript-eslint/experimental-utils`

lib/rules/no-unnecessary-act.ts
Belco90: I guess this is a good practice to force values being evaluated within an if statement to be purely boolean, but since we are not assigning them (just evaluating them within the if statement itself) I'd say it's ok to avoid casting to boolean.

When assigning values it could be dangerous `const foo = isReturnStatement(statement) && statement.argument` since `foo` could be a boolean or something else, but that's not a problem in the code we have.

Belco90 (changes requested):

lib/node-utils/is-node-of-type.ts
zaicevas: Nice :)

Belco90 (approved):
Looks good to me, thanks!

MichaelDeBoey (approved):

Issue #409, opened 2021-07-18
Extend `prefer-explicit-assert` to report `findBy*` and `waitFor`


zaicevas:
With `prefer-explicit-assert` enabled, `getBy*` usages without `expect` are reported, but `findBy*` and `waitFor` are not reported:

```
    screen.getByText('hey') // reported
    await waitFor(() => screen.getByText('hey')) // not reported
    await screen.findByText('hey') // not reported
```

What if we extended `prefer-explicit-assert` to report all of the above cases? 🤔

Belco90:
Hey @zaicevas! Few comments about this improvement idea.

What about `findBy*` usage saving returned element within a var? That should not be reported.

Not really sure about reporting `waitFor` + `getBy*` query, since inside `waitFor` callback the user might have other necessary stuff there.

I'd go just for reporting `findBy*` usages for now. What do you think?

zaicevas:
> What about findBy* usage saving returned element within a var? That should not be reported.

Makes sense to me, since `const foo = screen.getByText('hey')` is not reported and there's no reason the behaviour should be different for `findBy*`

> Not really sure about reporting waitFor + getBy* query, since inside waitFor callback the user might have other necessary stuff there.

Can you give an example to better understand what you have in mind?

> I'd go just for reporting findBy* usages for now. What do you think?

Sure, I'd imagine these 2 cases having separate PRs anyway.

Belco90:
> > Not really sure about reporting waitFor + getBy* query, since inside waitFor callback the user might have other necessary stuff there.
> 
> Can you give an example to better understand what you have in mind?

I don't have a really clear example in mind, to be honest. This is similar to `no-unnecessary-act` rule assuming there could be non-react-testing-library things inside `act`, in which case we need to allow.

This is a stupid example, but I'm not sure it should be reported:

```js
const promiseA = new Promise()
const promiseB = new Promise()
// ...
await waitFor(async () => {
  await promiseA;
  await promiseB;
  screen.getByRole('button')
})
```

It's obvious it should be rewritten to just do `await screen.findByRole('button')` instead of the whole `waitFor`, but could be a legit case were using `waitFor` +  awaiting promises +  asserting with implicit `getBy*`? I think there must be for sure, so I'd say just leave the rule complaining about that standalone `getBy*` query should be enough.

What do you think?

PR #421, opened 2021-07-19
feat(prefer-explicit-assert): report on `findBy*` queries too

zaicevas:
Part of https://github.com/testing-library/eslint-plugin-testing-library/issues/409

Reports standalone `findBy*` queries. Reporting wrong assertions is out of scope for this PR.


lib/rules/prefer-explicit-assert.ts
```diff
@@ -55,14 +97,33 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
           getQueryCalls.push(node);
         }
+
+        if (helpers.isFindQueryVariant(node)) {
+          findQueryCalls.push(node);
+        }
       },
       'Program:exit'() {
+        findQueryCalls.forEach((queryCall) => {
+          if (isVariableDeclaration(queryCall)) {
```
Belco90, line 114: Wouldn't be better to exclude those these nodes when pushing them to `findQueryCalls`?

tests/lib/rules/prefer-explicit-assert.test.ts
```diff
@@ -54,6 +54,24 @@ ruleTester.run(RULE_NAME, rule, {
...
+    })),
+    ...COMBINED_QUERIES_METHODS.map((queryMethod) => ({
+      code: `const quxElement = screen.find${queryMethod}('qux')`,
+    })),
+    ...COMBINED_QUERIES_METHODS.map((queryMethod) => ({
+      code: `
+      async () => {
+        const quxElement = await screen.find${queryMethod}('qux')
+      }`,
+    })),
```
Belco90, line 74: Can you add an extra valid test for a function returning a promise from `findBy*` query?

```js
function findBySubmit() {
  // this is valid
  return screen.findByRole('button', { name: 'submit' })
}
```

Belco90 (changes requested):

tests/lib/rules/prefer-explicit-assert.test.ts
zaicevas: Oh, seems like you've found a bug 😄

zaicevas: Great catch :) Should be good now

lib/rules/prefer-explicit-assert.ts
zaicevas: Not sure about that 🤔 Now there's even more logic in this if statement, so technically we could move everything into `'CallExpression Identifier'(node: TSESTree.Identifier) {`, but I think the current version is more consistent with how `getBy` case is implemented.

Belco90 (approved):
Seems good to go 🚀 

Thanks for your work on the recent PRs!

lib/rules/prefer-explicit-assert.ts
Belco90: Fair enough, we can leave it as it is.

Issue #416, opened 2021-07-19
Bug: `isStrict` option in `no-unnecessary-act` only reports strict violations


zaicevas:
With `isStrict: true` the `no-unnecessary-act` rule only reports mixed RTL+non RTL calls, but does not report `act` with only RTL calls, which is the behaviour without `isStrict` enabled.

```
      await act(async () => {
        userEvent.click(contactSupport) // not reported when isStrict is true
      })
```

I'll take a look at this one when I have the time.

Belco90:
This should have been caught by some test 😞

PR #423, opened 2021-07-20
fix(no-unnecessary-act): report everything that's reported in non-strict

zaicevas:
Closes https://github.com/testing-library/eslint-plugin-testing-library/issues/416

Everything that is reported with `isStrict` disabled should also be reported with `isStrict` enabled. That's what this PR fixes.

With regards to tests - went with the quickest approach. Shamelessly copy-pasted test cases and enabled `isStrict` 😄 If we have a better way to deal with this duplication - let me know.

### DIFF
--- a/lib/rules/no-unnecessary-act.ts
+++ b/lib/rules/no-unnecessary-act.ts
@@ -14,7 +14,7 @@ export const RULE_NAME = 'no-unnecessary-act';
 export type MessageIds =
   | 'noUnnecessaryActEmptyFunction'
   | 'noUnnecessaryActTestingLibraryUtil';
-type Options = [{ isStrict: boolean }];
+export type Options = [{ isStrict: boolean }];
 
 export default createTestingLibraryRule<Options, MessageIds>({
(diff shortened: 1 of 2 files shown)


Belco90 (dismissed):
I'm happy duplicating those tests to enable `isStrict` mode.

MichaelDeBoey:
What about instead of duplicating them, we create a helper function that takes rules and add the `strict: true` option, like I did in `consistent-data-testid`?

https://github.com/testing-library/eslint-plugin-testing-library/blob/4bf6d7621fe5872f0794dfd8c9939248abb50ce4/tests/lib/rules/consistent-data-testid.test.ts#L12-L23

https://github.com/testing-library/eslint-plugin-testing-library/blob/4bf6d7621fe5872f0794dfd8c9939248abb50ce4/tests/lib/rules/consistent-data-testid.test.ts#L357-L363

Belco90 (approved):

MichaelDeBoey (approved):

Issue #420, opened 2021-07-19
feat(prefer-find-by): report `waitFor` with presence matchers


zaicevas:
Originally mentioned in https://github.com/testing-library/eslint-plugin-testing-library/issues/411#issuecomment-882109271

`prefer-find-by` could report presence matchers (`.toBeInTheDocument()`, `toBeDefined()`, etc) that are used in conjunction with RTL queries. Currently, `prefer-find-by` does not report such cases:

```
    await waitFor(() => screen.getByText('button')) // reported
    await waitFor(() => expect(screen.getByText('button')).toBeInTheDocument()) // not reported
```

PR #450, opened 2021-08-09
feat(prefer-find-by): report presence assertions

zaicevas:
## Checks

## Changes

- Enhances `prefer-find-by` by reporting presence assertions

## Context

Closes https://github.com/testing-library/eslint-plugin-testing-library/issues/420
Not gonna lie, took much more effort and AST parsing than I initially thought 😄


Belco90:
@zaicevas oh wow, it seems complex indeed!

I'll try to take a look at the end of the week, thanks ✨

lib/rules/prefer-find-by.ts
```diff
@@ -111,6 +111,226 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+    ) {
+      if (
+        !isCallExpression(node.body) ||
+        !isMemberExpression(node.body.callee)
+      ) {
+        return null;
+      }
+
+      // expect(getByText).toBeInTheDocument() shape
+      if (
```
Belco90, line 125: Wouldn't `isPresenceAssert` and `isAbsenceAssert` be helpful here for some of the current checks?

lib/rules/prefer-find-by.ts
```diff
@@ -111,6 +111,226 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+        !isCallExpression(node.body) ||
+        !isMemberExpression(node.body.callee)
+      ) {
+        return null;
+      }
+
+      if (ASTUtils.isIdentifier(node.body.callee.object)) {
+        // () => screen.getByText
+        return node.body.callee.object.name;
+      } else if (
```
Belco90, line 206: These `else if` can be simplified to just `if` since you are returning inside the statement body right?

lib/rules/prefer-find-by.ts
```diff
@@ -166,17 +391,21 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
-        if (
-          !ASTUtils.isIdentifier(argument.body.callee) ||
-          !helpers.isSyncQuery(argument.body.callee)
-        ) {
+
+        if (!isSyncQuery(argument)) {
           return;
         }
+
         // shape of () => getByText
```
Belco90, line 399: Does this comment make sense anymore? Maybe this one (and others across the file) must be updated.

Belco90 (changes requested):
It looks great in general! I just requested an additional test.

The implementation to detect the `waitFor` + presence/absence assertion seems a bit brittle since it's too tight to a particular tree. But I guess this is what it is, right?

I'm surprised the auto-fix works for this new invalid usage reported 😮

lib/rules/prefer-find-by.ts
zaicevas: Hm, I am not sure how they'd fit in here 🤔

lib/rules/prefer-find-by.ts
zaicevas: Well, it's still accurate in the sense that it covers `() => getByText`, but it also includes `getByText` wrapped in a presence matcher, so I am not sure if I should update the comment to reflect that

tests/lib/rules/prefer-find-by.test.ts
```diff
@@ -120,7 +118,16 @@ ruleTester.run(RULE_NAME, rule, {
       code: `
         import {waitFor} from '@testing-library/foo';
         it('tests', async () => {
-          await waitFor(() => expect(${queryMethod}('baz')).toBeInTheDocument());
+          await waitFor(() => expect(screen.${queryMethod}('baz')).not.toBeInTheDocument());
```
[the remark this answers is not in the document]
zaicevas: I think there's already a test that covers this case https://github.com/testing-library/eslint-plugin-testing-library/blob/845a2287f451e65c829472ca40f6a4ac20d012d7/tests/lib/rules/prefer-find-by.test.ts#L113

However, I can't find an equivalent with a destructured query, so I've added that.

zaicevas:
> The implementation to detect the waitFor + presence/absence assertion seems a bit brittle since it's too tight to a particular tree. But I guess this is what it is, right?

Yeah :S I guess one can't escape that.

> I'm surprised the auto-fix works for this new invalid usage reported 😮

Yep, that's because it relies on a few variables (such as `queryMethod`, `queryVariant`) so it was enough to fill these properly.

---

I've noticed github actions are failing because of what seems to be unrelated reasons. Would you mind helping with that?

lib/rules/prefer-find-by.ts
Belco90: Ah sorry, I saw them used later. It wasn't sure what was the purpose of this function the first time I went through the PR.

Belco90 (approved):
All good from my side. Thanks again!

PR #452, opened 2021-08-16
feat: Add an option to remove the matching of findBy queries by the prefer-explicit-assert rule

### LINKED ISSUE #449: Please provide an option to remove the matching of findBy queries by the prefer-explicit-assert rule

prefer-explicit-assert

Fewer warnings

An option `matchFindByQueries` will be added, true by default.
When it's set to `false` the rule won't match these queries.

### Example code

```
await findByText(/no profile/i);
```

The current rule will report an error for such code, but I'd like to be able to exempt it easily.

When the new option is set to `false`, this code won't be flagged.
When the new option is absent or set to `true` the code will still be flagged as it is now, so it wouldn't be a breaking change.

### Anything else?

My 2 cents: it's common I use `findBy` assertions in my test code, not to assert that they're here, but rather to ensure we're waiting for something to be displayed before moving forward, so that the test isn't flaky. Therefore adding `expect` calls in these cases brings distractions :/

References: #409 #421

Yes

julienw:
A recent patch implemented matching `findBy*` queries by the
`prefer-explicit-assert` rule. This patch makes it possible to disable it
in cases where it's too distracting.

## Checks

## Changes

This adds an option so that a user can globally disable that findBy queries are checked by the role.

## Context

Fixes #449

### DIFF
--- a/docs/rules/prefer-explicit-assert.md
+++ b/docs/rules/prefer-explicit-assert.md
@@ -13,7 +13,7 @@ elements in their tests rather than just use `getBy*` queries and expect
 it doesn't throw an error so it's easier to understand what's the
 expected behavior within the test.
 
-Examples of **incorrect** code for this rule:
+Examples of **incorrect** code for this rule with the default configuration:
 
 ```js
 // just calling `getBy*` query expecting not to throw an error as an
@@ -23,9 +23,13 @@ getByText('foo');
 
 const utils = render(<Component />);
 utils.getByText('foo');
+
+// This is an incorrect code when `includeFindQueries` is `true`, which is the
+// default. Set it to `false` to shut off all warnings about find* queries.
+await findByText('foo');
 ```
 
-Examples of **correct** code for this rule:
+Examples of **correct** code for this rule with the default configuration:
 
 ```js
 // wrapping the get query within a `expect` and use some matcher for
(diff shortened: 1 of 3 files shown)


tests/lib/rules/prefer-explicit-assert.test.ts
```diff
@@ -60,6 +60,26 @@ ruleTester.run(RULE_NAME, rule, {
         const quxElement = await find${queryMethod}('qux')
       }`,
     })),
+    ...COMBINED_QUERIES_METHODS.map((queryMethod) => ({
+      code: `
+      async () => {
+        expect(await find${queryMethod}('qux')).toBeInTheDocument();
+      }`,
+    })),
```
julienw, line 68: I found this was missing so I added it here as well.

lib/rules/prefer-explicit-assert.ts
```diff
@@ -112,26 +114,28 @@ export default createTestingLibraryRule<Options, MessageIds>({
         }
       },
       'Program:exit'() {
-        findQueryCalls.forEach((queryCall) => {
-          const memberExpression = isMemberExpression(queryCall.parent)
-            ? queryCall.parent
-            : queryCall;
+        if (includeFindQueries) {
```
julienw, line 117: Probably a good idea to look at these changes without the whitespace changes.

docs/rules/prefer-explicit-assert.md
```diff
@@ -43,25 +47,26 @@ expect(queryByText('foo')).toBeInTheDocument();
...
-  assertion. This rule only allows defining a presence matcher
-  (`toBeInTheDocument`, `toBeTruthy`, or `toBeDefined`), but checks for both
-  presence and absence matchers (`not.toBeFalsy` and `not.toBeNull`). This means
-  other assertions such as `toHaveValue` or `toBeDisabled` will not trigger this
-  rule since these are valid uses with `getBy*`.
-
-  ```js
-  "testing-library/prefer-explicit-assert": ["error", {"assertion": "toBeInTheDocument"}],
-  ```
+| Option               | Required | Default | Details                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | Example               |
```
julienw, line 57: I've seen other rules were using a table, so I did it here too.

docs/rules/prefer-explicit-assert.md
```diff
@@ -43,25 +47,26 @@ expect(queryByText('foo')).toBeInTheDocument();
 await waitFor(() => getByText('foo'));
 fireEvent.click(getByText('bar'));
 const quxElement = getByText('qux');
+
+expect(await findbyText('foo')).toBeTruthy();
+const myButton = await screen.findByRole('button', { name: /Accept/ });
```
julienw, line 52: I also added this in the documentation, related to the previous patch.

julienw:
I believe this is ready for review now. Also I guess somebody needs to approve the test workflows :-)

Belco90 (approved):
Hey @julienw! Thanks for your PR, it looks good to me!

Belco90:
The step for publishing the new package version failed, I'm checking why.

Belco90:
@all-contributors please add @julienw for bug, code, test

PR #456, opened 2021-08-21
chore(devDeps): update dependencies

MichaelDeBoey:
CI is failing because `ts-jest` changed their `peerDependencies` (which should have been a major release)


Belco90 (approved):
3rd time this year we get CI broken because of new package version released introduced unexpected breaking changes.

Thinking seriously now about fixing versions in package.json

MichaelDeBoey:
I don't think we should fix versions in our `package.json` as CI failing is happening really rarely.
`ts-jest` should have made https://github.com/kulshekhar/ts-jest/pull/2816 a breaking change.

Fixing versions will break the `dedupe` feature too.

PR #458, opened 2021-08-22
chore(devDeps): make `ts-jest`'s version exact

MichaelDeBoey:
As said in https://github.com/testing-library/eslint-plugin-testing-library/pull/456#issuecomment-903276903, `ts-jest` doesn't apply SemVer, so we should make it an exact version

https://github.com/kulshekhar/ts-jest#versioning


Belco90 (approved):
Thank you

PR #468, opened 2021-09-04
feat(no-unnecessary-act): add to React config by default

### LINKED ISSUE #466: [no-unnecessary-act]: add to React config by default

MichaelDeBoey:
Closes #466

BREAKING CHANGE: `no-unnecessary-act` is now enabled by default in the React config (with `isStrict` set to `true`)


lib/configs/react.ts
```diff
@@ -13,6 +13,7 @@ export = {
     'testing-library/no-dom-import': ['error', 'react'],
     'testing-library/no-node-access': 'error',
     'testing-library/no-promise-in-fire-event': 'error',
+    'testing-library/no-unnecessary-act': 'error',
```
Belco90, line 16: I think we should enable it with `isStrict` option set to true. What do you think?

Belco90 (approved):

PR #472, opened 2021-09-08
fix(prefer-find-by): stop prefer-find-by fixer to break code when no …

dyatko:
…spaces before bracket

## Checks

## Changes

Add check for spaces before brackets for `prefer-find-by` fixer

## Context

The rule fixer didn't assume that some codebases may not separate brackets with spaces, so it's braking the code:

```
-        const {getByLabelText, getByTestId, getByText} = await renderComponent();
+        const {getByLabelText, getByTestId, getByTex, findByText} = await renderComponent();
```


lib/rules/prefer-find-by.ts
```diff
@@ -464,10 +464,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
               const textDestructuring = sourceCode.getText(
                 allVariableDeclarations
               );
+              const hasSpaceBeforeBracket = textDestructuring.endsWith(' }');
+              const symbolsToRemove = hasSpaceBeforeBracket ? 2 : 1;
```
Belco90, line 468: This could potentially be any number of spaces right? Would be worth it to implement this in a generic way so we calculate how many spaces there are between the last util and the closing bracket?

What do you think?

Belco90 (commented):
Hi @dyatko!

Good catch! And thank you very much for opening a PR with the fix, I really appreciate it.

It looks good to me, just a small question about if we could make it even more generic. Don't sweat it if it seems an overengineering solution, the current version is better than before!

lib/rules/prefer-find-by.ts
dyatko: @Belco90 yes, regex will do it best. Fixed in 41a8264

Belco90 (approved):
Great! Thanks again for your contribution.

Belco90:
Getting really weird errors in the pipeline when running the tests. I'll have to take a look, no idea what's happening at first sight.

dyatko:
@Belco90 yeah, same issues in another PR https://github.com/testing-library/eslint-plugin-testing-library/pull/465

Belco90:
> @Belco90 yeah, same issues in another PR #465

That's a different issue actually: you can see how it's failing to try to install some dependencies.

Our error is caused by Node v16.9.0: https://github.com/nodejs/node/issues/40030

We need to avoid this version meanwhile and stick to v16.8.0. I'm gonna create a PR to stick that Node version our GitHub Actions, so we can get that change here.

Belco90:
Pull request created: #473

Belco90:
@all-contributors please add @dyatko for bug and code

Issue #478, opened 2021-09-20
Breaking changes for next v5


Belco90:
I have a couple of breaking changes I think we should introduce in the new major version taking advantage of having to support ESLint 8

This project is maintained by volunteers. Support will be added as time permits.

If you're not willing or able to help with adding support:

- Please do not raise issues.
- Please do not comment on threads.
- Please be patient and subscribe to this issue for updates.

This update will require breaking changes on our side as well, so it will take some time to support it.
There is no ETA on this work. Again _please be patient_.

---

✅ = done
🕐 = in progress
🔲 = not yet started

### ✅ `no-render-in-setup`

Enable for `angular`, `react` & `vue` configs (`dom` doesn't have a render method)

- [x] https://github.com/testing-library/eslint-plugin-testing-library/pull/481

### ✅ `no-unnecessary-act`

Enable for `react` configs

- [x] https://github.com/testing-library/eslint-plugin-testing-library/pull/468

### ✅ `no-wait-for-multiple-assertions`

Enable for all configs

- [x] https://github.com/testing-library/eslint-plugin-testing-library/pull/483

### ✅ `no-wait-for-side-effects`

Enable for all configs

- [x] https://github.com/testing-library/eslint-plugin-testing-library/pull/484

### ✅ `no-wait-for-snapshot`

Enable for all configs

- [x] https://github.com/testing-library/eslint-plugin-testing-library/pull/485

### ✅ `prefer-presence-queries`

Enable for all configs

- [x] https://github.com/testing-library/eslint-plugin-testing-library/pull/486

### ✅ `prefer-query-by-disappearance`

Enable for all configs

- [x] https://github.com/testing-library/eslint-plugin-testing-library/pull/487

## ✅ Make `isStrict` option `true` by default for `no-unnecessary-act`

We have enabled this rule with `isStrict` to true in the shared configs, this option is still false by default. We should change it so it's true by default.

- [x] https://github.com/testing-library/eslint-plugin-testing-library/pull/479

## 🕐 Rules rename

We have some inconsistencies in naming some of our rules regarding using singular/plural.

### ✅ `no-debug` -> `no-debugging-utils`

- [x] https://github.com/testing-library/eslint-plugin-testing-library/pull/469

MichaelDeBoey:
> ### Enable more rules in the shared configs
>
> I think we should enable these rules too:
> - testing-library/no-render-in-setup
> - testing-library/no-unnecessary-act (✅ done already)
[...]

PR #479, opened 2021-09-23
feat(no-unnecessary-act): make `isStrict` option `true` by default

MichaelDeBoey:
Part of #478

BREAKING CHANGE: `isStrict` option is now `true` by default


Belco90:
Tests failing here! Since `isStrict` is enabled by default now, some tests were assuming it was disabled by default so this PR has to update tests to:

- [ ] explicitly set `isStrict` to false in test cases without the option
- [ ] remove the `isStrict` option in test cases where it was explicitly set to true

PR #480, opened 2021-09-23
feat(no-unnecessary-act): add to Angular, Dom & Vue configs by default

MichaelDeBoey:
Part of #478

BREAKING CHANGE: `no-unnecessary-act` is now enabled by default in the Angular, Dom & Vue configs


Belco90:
`act` util is from React, so it doesn't exist in Angular, Vue or "vanilla" DOM.

PR #529, opened 2022-01-06
Use disjunction to separate allowed render result names

diegocasmo:
## Checks

## Changes

- Use disjunction to separate allowed render result names

## Context

Separate allowed render result names by comma and an "or" before the last one.

Before:
```
`wrapper` is not a recommended name for `render` returned value. Instead, you should destructure it, or name it using one of: `view`, `utils`
```

After:
```
`wrapper` is not a recommended name for `render` returned value. Instead, you should destructure it, or name it using one of: `view`, or `utils`
```


Belco90 (approved):

Belco90:
@all-contributors please add @diegocasmo for code

PR #530, opened 2022-01-07
fix: update dependencies

MichaelDeBoey:
Closes #499

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -72,8 +72,6 @@ jobs:
           useLockFile: false
 
       - name: Install ESLint v${{ matrix.eslint }}
-        # force installation for now until we get ESLint and all plugins updated
-        # in dev dependencies
         run: npm install --no-save --force eslint@${{ matrix.eslint }}
 
       - name: Run tests
--- a/package.json
+++ b/package.json
@@ -42,37 +42,37 @@
     "prepare": "is-ci || husky install"
   },
   "dependencies": {
-    "@typescript-eslint/experimental-utils": "^5.5.0"
+    "@typescript-eslint/experimental-utils": "^5.9.0"
   },
   "devDependencies": {
-    "@babel/eslint-plugin": "^7.14.5",
-    "@commitlint/cli": "^16.0.1",
+    "@babel/eslint-plugin": "^7.16.5",
+    "@commitlint/cli": "^16.0.2",
     "@commitlint/config-conventional": "^16.0.0",
-    "@types/jest": "^27.0.3",
-    "@types/node": "^16.11.11",
-    "@typescript-eslint/eslint-plugin": "^5.5.0",
-    "@typescript-eslint/parser": "^5.5.0",
+    "@types/jest": "^27.4.0",
+    "@types/node": "^16.11.19",
+    "@typescript-eslint/eslint-plugin": "^5.9.0",
+    "@typescript-eslint/parser": "^5.9.0",
     "cpy-cli": "^3.1.1",
-    "eslint": "^7.32.0",
-    "eslint-config-kentcdodds": "^19.2.0",
+    "eslint": "^8.6.0",
+    "eslint-config-kentcdodds": "^20.0.1",
     "eslint-config-prettier": "^8.3.0",
-    "eslint-plugin-import": "^2.25.3",
-    "eslint-plugin-jest": "^25.3.0",
+    "eslint-plugin-import": "^2.25.4",
+    "eslint-plugin-jest": "^25.3.4",
     "eslint-plugin-jest-formatting": "^3.1.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-promise": "^6.0.0",
-    "eslint-remote-tester": "^2.0.1",
+    "eslint-remote-tester": "^2.1.1",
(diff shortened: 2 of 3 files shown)


Belco90:
With this PR, #499 isn't necessary anymore, is it?

MichaelDeBoey:
@Belco90 That's indeed correct 👍

package.json
```diff
@@ -42,37 +42,37 @@
     "prepare": "is-ci || husky install"
   },
   "dependencies": {
-    "@typescript-eslint/experimental-utils": "^5.5.0"
+    "@typescript-eslint/experimental-utils": "^5.9.0"
```
Belco90, line 45: Can you avoid upgrading this one? That would trigger a new patch version of the plugin.

package.json
```diff
@@ -42,37 +42,37 @@
...
     "@commitlint/cli": "^16.0.1",
     "@commitlint/config-conventional": "^16.0.0",
-    "@types/jest": "^27.0.3",
-    "@types/node": "^16.11.11",
-    "@typescript-eslint/eslint-plugin": "^5.5.0",
-    "@typescript-eslint/parser": "^5.5.0",
+    "@types/jest": "^27.4.0",
+    "@types/node": "^16.11.19",
+    "@typescript-eslint/eslint-plugin": "^5.9.0",
+    "@typescript-eslint/parser": "^5.9.0",
```
Belco90, line 54: Since `@typescript-eslint/experimental-utils` shouldn't be upgraded, these two either for consistency within the ecosystem.

package.json
```diff
@@ -42,37 +42,37 @@
...
     "cpy-cli": "^3.1.1",
-    "eslint": "^7.32.0",
-    "eslint-config-kentcdodds": "^19.2.0",
+    "eslint": "^8.6.0",
+    "eslint-config-kentcdodds": "^20.0.0-alpha.3",
```
Belco90, line 57: Can't we use the stable v20 already?

Belco90 (changes requested):
An additional change:

Can you remove the `--force` flag from the `pipeline` GitHub action [as in the other PR](https://github.com/testing-library/eslint-plugin-testing-library/pull/499/files#diff-fe700949ba86403864aa7b6fdcab2c00c55e1047f73a6a282135a35e8a21e03b)? Thanks

package.json
MichaelDeBoey: Since `v5.0.2` is released earlier today, we technically already use this version 🤷‍♂️

Belco90: We releasing a new version of the plugin doesn't mean the final user using our plugin will update to the very latest version of `@typescript-eslint/experimental-utils` when updating our plugin in their codebase.

We are defining a range of valid versions for `@typescript-eslint/experimental-utils` that the final user can match. The current range is equal to or greater than 5.5.0 before. A user with the current version of our plugin could have version 5.6.0 of `@typescript-eslint/experimental-utils`. They might have fixed their dependencies, so updating `eslint-plugin-testing-library` doesn't install all the latest versions of its dependencies (`@typescript-eslint/experimental-utils` in this case). They are free to remain using `@typescript-eslint/experimental-utils` 5.6.0 under the hood then.

Now you want to bump the valid range up to equal to or greater than 5.9.0, which means the previous user would get the sub dependency `@typescript-eslint/experimental-utils` forced to get bumped from 5.6.0 to 5.9.0. That's a change from the point of view of our Semantic Versioning Policy: there is a change for the final user, a sub dependency must be updated.

That's the reason why bumping `dependencies` for our plugin causes a new release of the plugin. Since this is a chore release, we don't need to upgrade our main dependencies.

.github/workflows/pipeline.yml
```diff
@@ -72,9 +72,7 @@ jobs:
           useLockFile: false
 
       - name: Install ESLint v${{ matrix.eslint }}
-        # force installation for now until we get ESLint and all plugins updated
-        # in dev dependencies
-        run: npm install --no-save --force eslint@${{ matrix.eslint }}
```
Belco90, line 77: Sorry, I just saw that [`eslint-config-kentcdodds` is compatible only with ESLint v8](https://github.com/kentcdodds/eslint-config-kentcdodds/blob/f28506ef913486c19fd4c69d8e1eac80cc5e4ee4/package.json#L63), so we can't remove the `--force` flag here.

Can you put it back? The comment is not up to date tho, it should be something like "force installation to skip dev dependencies not compatible with ESLint v7"

Belco90 (changes requested):

package.json
MichaelDeBoey: Let's make this a `fix` then 👍

Belco90: But why that rush to always set the very latest version?

I'm a bit worried about this, we are always mixing updating production dependencies and dev dependencies. We need to be a bit flexible with prod dependencies. Our plugin doesn't need the last version of that dependency, there is no benefit in doing this right now (and most of the times).

MichaelDeBoey: I'm updating all the dependencies, so why not use the latest for normal dependencies too?

There's no harm in using the latest minor, as it's a not breaking anything at all 🤷‍♂️

Belco90: There is no harm for us, but there might be for some final users. Not the end of the world, of course. But in the same way that there is no harm, there is no benefit.

I'll approve this one just to get everything related to ESLint updated to v8, but please let's stick to the proper scope in future PRs when updating things.

Belco90 (approved):

PR #535, opened 2022-01-13
docs(no-unnecessary-act): isStrict is enabled by default

zaicevas:
## Checks

## Changes

Tweak docs to reflect `isStrict` option on `no-unnecessary-act` being enabled by default

## Context

I've noticed https://github.com/testing-library/eslint-plugin-testing-library/pull/479 didn't change docs to reflect default `isStrict` option.


Belco90 (approved):
Good catch!

PR #538, opened 2022-01-20
fix(prefer-find-by): fix report with querySelector

### LINKED ISSUE #501: `prefer-find-by` reports `querySelector`
### Plugin version

4.12.4

### ESLint version

7.24.0

### Node.js version

14.16.0

### npm/yarn version

6.14.11

### Operating system

macOS Big Sur

### Bug description

I expect this code snippet NOT to be reported. Furthermore, error message and auto-fix have `undefined` instead of a meaningful message/fix.

```
    it('renders', async () => {
      const { container } = render(<App />)

      await waitFor(() =>
        expect(container.querySelector("[data-icon-name='bookmark-filled']")).toBeInTheDocument(), // reported
      )
    })
```

Auto-fixed to: `await container.findByundefined("[data-icon-name='bookmark-filled']")`

### Steps to reproduce

1. Setup `eslint-plugin-testing-library` 4.12.4
2. Enable `prefer-find-by` rule
3. Create a test with the above code snippet
4. Run eslint with `--fix`

### Error output/screenshots

```
error  Prefer `findByundefined` query over using `waitFor` + `querySelector`  testing-library/prefer-find-by
```

### ESLint configuration

```
    {
      files: ['*.test.ts', '*.test.tsx'],
      parser: '@typescript-eslint/parser',
      plugins: ['testing-library'],
      rules: {
        '@typescript-eslint/no-var-requires': 'off',
        'testing-library/prefer-find-by': 'error',
      },
    },
```

### Rule(s) affected

`prefer-find-by`

### Anything else?

_No response_

No

bpinto:
## Checks

## Changes

- No longer report invalid usages with a missing query method.
- Add a test covering the code changes.

## Context

Fixes #501

### DIFF
--- a/lib/rules/prefer-find-by.ts
+++ b/lib/rules/prefer-find-by.ts
@@ -373,6 +373,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
           const callArguments = getQueryArguments(argument.body);
           const queryMethod = fullQueryMethod.split('By')[1];
 
+          if (!queryMethod) {
+            return;
+          }
+
           reportInvalidUsage(node, {
             queryMethod,
             queryVariant,
--- a/tests/lib/rules/prefer-find-by.test.ts
+++ b/tests/lib/rules/prefer-find-by.test.ts
@@ -149,6 +149,14 @@ ruleTester.run(RULE_NAME, rule, {
         })
       `,
     },
+    {
+      code: `
+        import {screen, waitFor} from '@testing-library/foo';
+        it('tests', async () => {
+          await waitFor(() => expect(screen.querySelector('baz')).toBeInTheDocument());
+        })
+      `,
+    },
   ],
   invalid: [
     ...createScenario((waitMethod: string, queryMethod: string) => ({
--- a/tests/lib/rules/prefer-find-by.test.ts
+++ b/tests/lib/rules/prefer-find-by.test.ts
@@ -157,6 +157,15 @@ ruleTester.run(RULE_NAME, rule, {
         })
       `,
     },
+    {
+      code: `
+        import {waitFor} from '@testing-library/foo';
+        it('tests', async () => {
+          const { container } = render()
+          await waitFor(() => expect(container.querySelector('baz')).toBeInTheDocument());
+        })
+      `,
+    },
   ],
   invalid: [
     ...createScenario((waitMethod: string, queryMethod: string) => ({


bpinto:
It's my first contribution to the project so if my fix is bad/wrong let me know how I can improve this.

lib/rules/prefer-find-by.ts
```diff
@@ -373,6 +373,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
           const callArguments = getQueryArguments(argument.body);
           const queryMethod = fullQueryMethod.split('By')[1];
 
+          if (!queryMethod) {
+            return;
```
Belco90, line 377: This partially fixes the issue. This `return` is wrapped within `isScreenSyncQuery` condition, so it will be only applied to `screen.foo` scenarios.

You need to do the same check + `return` a few lines below here, where the other `fullQueryMethod.split('By')[1]` is located.

tests/lib/rules/prefer-find-by.test.ts
```diff
@@ -149,6 +149,14 @@ ruleTester.run(RULE_NAME, rule, {
         })
       `,
     },
+    {
+      code: `
+        import {screen, waitFor} from '@testing-library/foo';
+        it('tests', async () => {
+          await waitFor(() => expect(screen.querySelector('baz')).toBeInTheDocument());
```
Belco90, line 156: If you check the original issue #501, the problematic code was a bit different:

```js
    it('renders', async () => {
      const { container } = render(<App />)

      await waitFor(() =>
        expect(container.querySelector("[data-icon-name='bookmark-filled']")).toBeInTheDocument(), // reported
      )
    })
```

Can you add a test for this one too, please?

Belco90 (changes requested):
Hi @bpinto! Thanks for your first contribution. The fix itself seems fine but handles the original issue partially. I've pointed out somewhere else where you need to apply the fix.

lib/rules/prefer-find-by.ts
bpinto: Thanks for catching that! I missed that this would not work for `container` queries.

tests/lib/rules/prefer-find-by.test.ts
bpinto: I'll add this test, I missed that the original issue was slightly different to what I was experiencing on my side. :+1:

lib/rules/prefer-find-by.ts
bpinto: Hmm actually, I've added the test however it is not failing. `isScreenSyncQuery` is returning true.

```js
function isScreenSyncQuery(node: TSESTree.ArrowFunctionExpression) {
...
  return (
    helpers.isSyncQuery(node.body.callee.property) ||
    isWrappedInPresenceAssert || // this is evaluating to true
    isWrappedInNegatedPresenceAssert
  );
}

Belco90: Ooh, you are right! My bad, sorry: I assumed the isScreenSyncQuery was coming from our helpers, but it's a function from the rule itself. This is fine then!

Belco90 (approved):
All good now, thank you very much!

Belco90:
@all-contributors please add @bpinto for code and test.

PR #539, opened 2022-01-20
test: wait for async process completion

bpinto:
## Checks

## Changes

`generateConfigs()` function executes a nodejs [exec] that takes many
seconds to complete. As the test was not waiting for the process to
terminate, Jest was complaining with the following error message:

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

Additionally, this test could have a false positive expectation since it
was likely to run the snapshot comparison with previously generated
plugin configurations.

By default Jest tests will timeout if they take longer than 5 seconds to
complete, since this test takes more than 10 seconds to complete on my
local machine, this test timeout configuration has been increased to 20
seconds.

[exec]: https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback

### DIFF
--- a/tests/index.test.ts
+++ b/tests/index.test.ts
@@ -1,10 +1,12 @@
 import { exec } from 'child_process';
 import { existsSync } from 'fs';
 import { resolve } from 'path';
+import util from 'util';
 
 import plugin from '../lib';
 
-const generateConfigs = () => exec(`npm run generate:configs`);
+const execAsync = util.promisify(exec);
+const generateConfigs = () => execAsync(`npm run generate:configs`);
 
 const numberOfRules = 26;
 const ruleNames = Object.keys(plugin.rules);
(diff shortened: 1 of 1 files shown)


tests/index.test.ts
```diff
@@ -1,10 +1,12 @@
 import { exec } from 'child_process';
 import { existsSync } from 'fs';
 import { resolve } from 'path';
+import util from 'util';
 
 import plugin from '../lib';
 
-const generateConfigs = () => exec(`npm run generate:configs`);
+const execAsync = util.promisify(exec);
```
Belco90, line 8: TIL about `promisify`!

Belco90 (commented):
Hi again @bpinto, looks good to me! I'm gonna wait for @MichaelDeBoey's review tho since he built this and has more context.

Belco90 (approved):
Let's go ahead with this. @MichaelDeBoey can request/address any changes later if needed.

bpinto:
@MichaelDeBoey ill be happy to address any comments you may have in the future related to this change.

PR #550, opened 2022-02-01
fix: move to non-experimental TS-ESLint utils

MichaelDeBoey:
@typescript-eslint has renamed `experimental-utils` to just `utils` in https://github.com/typescript-eslint/typescript-eslint/pull/4172

Some other plugins have already started migrating to the newer name too:
- https://github.com/jest-community/eslint-plugin-jest/pull/1035

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -36,7 +36,7 @@
     "no-restricted-imports": [
       "error",
       {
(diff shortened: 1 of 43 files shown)


Belco90 (changes requested):
@MichaelDeBoey amazing! Thanks for taking care of this. Just a small request about dependencies version and should be good to go.

Belco90 (approved):

PR #556, opened 2022-03-03
fix: use `ASTUtils`'s `isNodeOfType` instead of own version

MichaelDeBoey:
Now that https://github.com/typescript-eslint/typescript-eslint/pull/4503 is merged & released, we can get rid of our own version of `isNodeOfType`


Belco90:
Nice! Let's get this merged.

Belco90 (approved):

Issue #518, opened 2021-12-13
False positives with testing-library/prefer-presence-queries


domarmstrong:
### Plugin version

v5.0.0

### ESLint version

v8.2.0

### Node.js version

14.18.0

### npm/yarn version

1.22.17

### Operating system

Big Sur

### Bug description

Generally I like this rule but we have to disable it in quite a few places due to false positives with toBeNull.

### Steps to reproduce

Examples:

```typescript
expect(queryTooltip(screen.getByText('Some text'))).toBeNull();
```

```typescript
expect(within(screen.getByRole('dialog')).queryByText('Some text')).toBeNull();
```

```typescript
expect(screen.getByText('Some text').getAttribute('disabled')).toBeNull();
```

### ESLint configuration

```js
    {
      files: ['src/**/*.test.[tj]s?(x)', 'src/test/**/*.[tj]s?(x)'],
      env: {
        jest: true,
        browser: true,
      },
      plugins: ['eslint-plugin-jest', 'eslint-plugin-testing-library'],
      extends: ['plugin:jest/recommended', 'plugin:testing-library/react'],
      rules: {
        // warning are under migration
        'testing-library/prefer-screen-queries': 'warn',
        'testing-library/no-render-in-setup': 'warn',
        'testing-library/render-result-naming-convention': 'warn',
        'testing-library/no-wait-for-multiple-assertions': 'warn',
        'testing-library/no-node-access': 'off'
      },
    },
```

### Rule(s) affected

testing-library/prefer-presence-queries

### Do you want to submit a pull request to fix this bug?

No

Belco90:
Thanks for reporting this @domarmstrong. Interesting issue, it seems it's reporting false positives when the Testing Library query is nested into something else. We will have to take a look at this at some point.

josias-r:
Another case of a false positive:
https://github.com/facebook/react/issues/23093

Belco90:
Hey @josias-r, this is a different problem. Could you move it to a different issue? Thanks.

themagickoala:
I will try and make time for this if I can, but it's unlikely as I've come down with Covid and still having to look after my 2 kids/try and work when I can! I've done a little investigation and I'm going to make some notes here in case somebody else finds it and wants to have a go (this is from the perspective of somebody who has never written anything to do with eslint before):

[...]

PR #557, opened 2022-03-11
feat(prefer-presence-queries): add `absence` & `presence` options

themagickoala:
## Checks

## Changes

- Added configuration options to `testing-library/prefer-presence-queries` to enable or disable just the presence/absence checks by themselves;
- I started out trying to implement the check for `within`, but I think I'd have to do a lot more reading about eslint before I can tackle that, whereas this configuration-based change was a lot simpler. I do have a stash with some changes based on that, but I'm unlikely to get to them soon.

## Context

Provides a patch for #518 (without entirely fixing it).

### DIFF
--- a/lib/rules/prefer-presence-queries.ts
+++ b/lib/rules/prefer-presence-queries.ts
@@ -5,7 +5,12 @@ import { findClosestCallNode, isMemberExpression } from '../node-utils';
 
 export const RULE_NAME = 'prefer-presence-queries';
 export type MessageIds = 'wrongAbsenceQuery' | 'wrongPresenceQuery';
-type Options = [];
+export type Options = [
+  {
+    presence?: boolean;
+    absence?: boolean;
+  }
+];
 
 export default createTestingLibraryRule<Options, MessageIds>({
   name: RULE_NAME,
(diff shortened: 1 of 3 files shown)


lib/rules/prefer-presence-queries.ts
```diff
@@ -26,12 +31,31 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+  defaultOptions: [
+    {
+      presence: true,
+      absence: true,
+    },
+  ],
 
-  create(context, _, helpers) {
+  create(context, [options], helpers) {
+    const { presence, absence } = options;
```
MichaelDeBoey, line 58:
```suggestion
  create(context, [{ absence = true, presence = true }], helpers) {
```

MichaelDeBoey (dismissed):

tests/lib/rules/prefer-presence-queries.test.ts
```diff
@@ -16,38 +17,93 @@ const queryAllByQueries = ALL_QUERIES_METHODS.map(
...
+}: Omit<AssertionFnParams, 'messageId'>): RuleValidTestCase[] => {
+  const finalQuery = shouldUseScreen ? `screen.${query}` : query;
+  const code = `expect(${finalQuery}('Hello'))${matcher}`;
+  return [
+    {
+      code,
+    },
+    {
+      code,
+      options: [
```
Belco90, line 45: You are now passing the `options` property to all the test cases involving the presence or absent matchers. Could you add some additional tests not setting any `options` to check that by default both options are enabled?

Belco90 (changes requested):
@themagickoala thanks for your contribution! The PR looks great, just needs some extra tests and should be ready.

tests/lib/rules/prefer-presence-queries.test.ts
themagickoala: I've actually adapted the `getValidAssertion` method to return an array of valid test cases rather than a single test case, including the case with no options set, the case with neither option set, and the case with just the intended option set. By contrast, `getInvalidAssertion` has been adapted to return an array of invalid test cases, one with no options set, and one with only the intended option set (as having both options disabled cannot result in an invalid test case).

Belco90: Aah sorry, my bad. I missed the case with no options set. All good then!

Belco90 (approved):

Belco90 (changes requested):
I think there are some weird test cases now, let me check.

Belco90 (approved):
I got confused by the new methods to generate the assertions. Definitely all good!

Belco90:
@all-cotributors please add @themagickoala for code and test

Belco90:
@all-contributors please add @themagickoala for code and test

themagickoala:
Thanks @Belco90! I agree that the new methods were slightly confusing. I feel like there are some optimisations that could be made to the test suite as a whole, essentially mapping over a few dimensions of things to generate all the tests that are currently being created manually, but I didn't want to do too much and overstep the purpose of the PR!

PR #560, opened 2022-03-29
feat: add `no-global-regex-flag-in-query` rule

### LINKED ISSUE #559: New rule: no-global-regex-flag-to-query
### Name for new rule

no-global-regex-flag-to-query

The rule would warn when a user uses the global regex flag to search for text/name.
A regex that uses the global flag holds the `lastIndex` state when it found a match, due to this it could be that a new match can't be found.

See https://github.com/testing-library/dom-testing-library/pull/1116 for the fix within DTL, where we added a log when we detect this.

### Testing Library feature

Queries that use a regex

All

_No response_

### Code examples

```ts
screen.getAllByText(/hello world/g)
screen.findByRole('button', {name: /log in/ig})
```

### Anything else?

_No response_

Yes

timdeschryver:
## Checks

## Changes

- adds the new rule `no-global-regex-flag-in-query`

## Context

Resolves #559 

It's been a long time since I've contributed, feel free to comment on this PR :)


MichaelDeBoey (dismissed):

lib/rules/no-global-regexp-flag-in-query.ts
```diff
@@ -0,0 +1,83 @@
...
+  name: RULE_NAME,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Disallow the use of the global RegExp flag (/g) in queries',
+      recommendedConfig: {
+        dom: 'error',
+        angular: 'error',
+        react: 'error',
+        vue: 'error',
```
Belco90, line 26: Per our [Semantic Versioning Policy](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/semantic-versioning-policy.md), reporting more errors from our recommended config would be a major version. Let's avoid setting this rule in any recommended config for now, and we will add it to the list of changes for the next major of the plugin.

lib/node-utils/is-node-of-type.ts
```diff
@@ -59,3 +59,4 @@ export const isReturnStatement = ASTUtils.isNodeOfType(
 export const isFunctionExpression = ASTUtils.isNodeOfType(
   AST_NODE_TYPES.FunctionExpression
 );
+export const isIdentifier = ASTUtils.isNodeOfType(AST_NODE_TYPES.Identifier);
```
Belco90, line 62: The `isIdentifier` util already exists in the `ASTUtils` exported from `@typescript-eslint/utils`.

It's misleading to find these AST utils imported from 2 different modules, so we should re-export all the default ones from our custom module, and forbid the `@typescript-eslint/utils`. Until then, let's used the one exposed from `@typescript-eslint/utils`

lib/rules/no-global-regexp-flag-in-query.ts
```diff
@@ -0,0 +1,83 @@
...
+      ) {
+        if (
+          isCallExpression(node.parent) &&
+          isMemberExpression(node.parent.callee) &&
+          isIdentifier(node.parent.callee.property)
+        ) {
+          lint(node, node.parent.callee.property);
+        }
+      },
+      [`CallExpression[callee.type=Identifier] > Literal[regex.flags=/g/].arguments`](
```
Belco90, line 61: Any specific reason to select only those `CallExpression` whose callee is `Identifier` or `MemberExpression`?

tests/lib/rules/no-global-regexp-flag-in-query.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        import { screen } from '@testing-library/dom'
+        screen.findByRole('button', {name: 'hello'})
+    `,
+    `
+        const utils = render(<Component/>)
+        utils.findByRole('button', {name: /hello/i})
+    `,
+    `
+        const {queryAllByPlaceholderText} = render(<Component/>)
+        queryAllByPlaceholderText(/hello/i)
```
Belco90, line 41: Could you add an extra valid test checking something not related to Testing Library can be called with a regexp with the `g` option?

tests/lib/rules/no-global-regexp-flag-in-query.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    `,
+  ],
+  invalid: [
+    {
+      code: `
+        import { screen } from '@testing-library/dom'
+        screen.getByText(/hello/g)`,
+      errors: [
+        {
+          messageId: 'noGlobalRegExpFlagInQuery',
```
Belco90, line 51: Could you check in which row and column the error is displayed? For all invalid test cases.

tests/lib/rules/no-global-regexp-flag-in-query.test.ts
Belco90, line 41: Could you add an extra valid test checking that something related to Testing Library different from a query can be called with a regexp with the `g` option? Even if it's an option which doesn't really exist in the real method.

Belco90 (changes requested):
@timdeschryver Thanks for implementing this so quickly! I added a couple of comments and asked for some small changes, but should be easy to address.

timdeschryver:
@Belco90 and @MichaelDeBoey  thanks for quick review.
I'll address these points tomorrow.
I also was thinking about a fixer that removes the regexp, thoughts?

lib/rules/no-global-regexp-flag-in-query.ts
timdeschryver: AFAIK these two are the ones that can invoke a method `getByText()` and `something.getByText()`.
To get `getByText()` to verify it's on of our methods, I had to traverse the AST to find that identifier.
Maybe that there's a better way though 😅 (I hope that there is one)

Belco90: Oh, I see. You should be able to traverse the AST to find the identifier with [`getDeepestIdentifierNode`](https://github.com/testing-library/eslint-plugin-testing-library/blob/7bc2b9c10b2b39ded5135295e9cfa1472211cbf9/lib/node-utils/index.ts#L340):

> Gets the deepest identifier node in the expression from a given node.
> 
> An example:
> Having `const a = rtl.within('foo').getByRole('button')`:
> if we call `getDeepestIdentifierNode` with `rtl` node,
> it will return `getByRole` identifier

Belco90: Actually, seeing this example, I just remembered some issues with chained `within` in another rule. Could you add an extra test to make sure the rule reports correctly a query chained with `within`?

Belco90:
> I also was thinking about a fixer that removes the regexp, thoughts?

I'd love to have the autofix, and should be fairly simple. However, I didn't want to request including this feature to merge this PR, so I'm happy about merging this one first and then add the autofix in a later PR if that's more convenient to you.

tests/lib/rules/no-global-regexp-flag-in-query.test.ts
Belco90: It seems this one doesn't have the line and column yet?

timdeschryver (commented):
@Belco90 the last commit refactors the implementation with `getDeepestIdentifierNode`. It does seem more robust to me, let me know what you think :)

lib/rules/no-global-regexp-flag-in-query.ts
```diff
@@ -0,0 +1,94 @@
...
+
+        const [firstArg, secondArg] = isCallExpression(identifierNode.parent)
+          ? identifierNode.parent.arguments
+          : isMemberExpression(identifierNode.parent) &&
+            isCallExpression(identifierNode.parent.parent)
+          ? identifierNode.parent.parent.arguments
+          : [];
+
+        if (!report(firstArg)) {
+          if (isObjectExpression(secondArg)) {
```
Belco90, line 80: I guess this two can be combined into a single if statement.

lib/rules/no-global-regexp-flag-in-query.ts
```diff
@@ -0,0 +1,94 @@
...
+        if (!identifierNode || !helpers.isQuery(identifierNode)) {
+          return;
+        }
+
+        const [firstArg, secondArg] = isCallExpression(identifierNode.parent)
+          ? identifierNode.parent.arguments
+          : isMemberExpression(identifierNode.parent) &&
+            isCallExpression(identifierNode.parent.parent)
+          ? identifierNode.parent.parent.arguments
+          : [];
```
Belco90, line 77: This is not really readable, could you transform it into regular if/else-if/else statements?

I know you won't be able to declare the variables as const, but the code will be much cleaner.

tests/lib/rules/no-global-regexp-flag-in-query.test.ts
```diff
@@ -0,0 +1,189 @@
...
+        },
+      ],
+      output: `
+        const {queryAllByLabelText} = render(<Component/>)
+        queryAllByLabelText(/hello/i)`,
+    },
+    {
+      code: `
+        import { within } from '@testing-library/dom'
+        within(element).findByRole('button', {name: /hello/ig})`,
```
Belco90, line 161: Could you swap the `ig` regexp as to `gi` in one of these tests, so we also make sure the order of the params don't affect the autofix?

Belco90 (commented):
@timdeschryver Much better now, including the autofix function!

Just a couple of small tweaks and should be ready to go.

tests/lib/rules/no-global-regexp-flag-in-query.test.ts
timdeschryver: `gi` and `ig` are already used in the invalid cases, I'll convert this to a regexp that uses 3 flags.

lib/rules/no-global-regexp-flag-in-query.ts
timdeschryver: 💯

timdeschryver: Thanks for calling me out 😅

tests/lib/rules/no-global-regexp-flag-in-query.test.ts
Belco90: Ooh my bad! I didn't realize about that. Good idea about including the other format.

tests/lib/rules/no-global-regexp-flag-in-query.test.ts
```diff
@@ -0,0 +1,189 @@
...
+      errors: [
+        {
+          messageId: 'noGlobalRegExpFlagInQuery',
+          line: 3,
+          column: 44,
+        },
+      ],
+      output: `
+        import { screen } from '@testing-library/dom'
+        screen.findByRole('button', {name: /hellogg/})`,
```
Belco90, line 111: Great test case 💯

Belco90 (approved):
This LGTM!

PR #566, opened 2022-04-03
fix(no-global-regexp-flag-in-query): check if empty name property node

### LINKED ISSUE #565: no-global-regexp-flag-in-query: TypeError 'value' of undefined
### Plugin version

5.2.0

### ESLint version

8.12.0

### Node.js version

16.8.0

### npm/yarn version

npm 7.21.0

### Operating system

Windows 10

### Bug description

I enabled the `no-global-regexp-flag-in-query` rule and received an error when linting

### Steps to reproduce

Create the following test file:

```
import { screen } from "@testing-library/react";

describe("App", () => {
    test("is rendered", async () => {
        // The line below is the only one that matters here
        await screen.findByText("Hello World", {
            exact: false,
        });
    });
});
```

Enable the rule and lint

### Error output/screenshots

```
TypeError: Cannot read property 'value' of undefined
Rule: "testing-library/no-global-regexp-flag-in-query"
    at Object.CallExpression (node_modules\eslint-plugin-testing-library\rules\no-global-regexp-flag-in-query.js:76:45)
    at enhancedRuleInstructions.<computed> (node_modules\eslint-plugin-testing-library\create-testing-library-rule\detect-testing-library-utils.js:518:105)    at ruleErrorHandler (node_modules\eslint\lib\linter\linter.js:1114:28)
    at node_modules\eslint\lib\linter\safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (node_modules\eslint\lib\linter\safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (node_modules\eslint\lib\linter\node-event-generator.js:297:26)
    at NodeEventGenerator.applySelectors (node_modules\eslint\lib\linter\node-event-generator.js:326:22)
    at NodeEventGenerator.enterNode (node_modules\eslint\lib\linter\node-event-generator.js:340:14)
    at CodePathAnalyzer.enterNode (node_modules\eslint\lib\linter\code-path-analysis\code-path-analyzer.js:795:23)
```

### ESLint configuration

```
{
    plugins: ["testing-library"],
    env: {
        es6: true,
        browser: true,
    },
    parser: "@babel/eslint-parser",
    parserOptions: {
        ecmaFeatures: {
            jsx: true,
        },
        sourceType: "module",
        requireConfigFile: false,
    },
    settings: {
        react: {
            pragma: "React",
            version: "detect",
        },
    },
    overrides: [
        {
            files: ["*.test.jsx", "*.test.tsx"],
            rules: {
                "testing-library/no-global-regexp-flag-in-query": "error",
            },
        },
    ],
}
```

### Rule(s) affected

`no-global-regexp-flag-in-query`

### Anything else?

_No response_

No

Belco90:
## Checks

## Changes

- check if the `namePropertyNode` var exists since it can be `undefined`.

## Context

Closes #565

### DIFF
--- a/tests/lib/rules/no-global-regexp-flag-in-query.test.ts
+++ b/tests/lib/rules/no-global-regexp-flag-in-query.test.ts
@@ -80,8 +80,7 @@ ruleTester.run(RULE_NAME, rule, {
     `,
 
     // issue #565
-    {
-      code: `
+    `
       import { screen } from "@testing-library/react"
 
       describe("App", () => {
@@ -90,7 +89,6 @@ ruleTester.run(RULE_NAME, rule, {
         })
       })
     `,
-    },
   ],
   invalid: [
     {
--- a/tests/lib/rules/no-global-regexp-flag-in-query.test.ts
+++ b/tests/lib/rules/no-global-regexp-flag-in-query.test.ts
@@ -78,6 +78,19 @@ ruleTester.run(RULE_NAME, rule, {
         const utils = render(<Component/>)
         utils.notAQuery(/hello/i)
     `,
+
+    // issue #565
+    {
+      code: `
+      import { screen } from "@testing-library/react"
+
+      describe("App", () => {
+        test("is rendered", async () => {
+          await screen.findByText("Hello World", { exact: false });
+        })
+      })
+    `,
+    },
   ],
   invalid: [
     {
--- a/lib/rules/no-global-regexp-flag-in-query.ts
+++ b/lib/rules/no-global-regexp-flag-in-query.ts
@@ -96,8 +96,11 @@ export default createTestingLibraryRule<Options, MessageIds>({
               ASTUtils.isIdentifier(p.key) &&
               p.key.name === 'name' &&
               isLiteral(p.value)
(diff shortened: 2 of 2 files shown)


tests/lib/rules/no-global-regexp-flag-in-query.test.ts
```diff
@@ -78,6 +78,19 @@ ruleTester.run(RULE_NAME, rule, {
         const utils = render(<Component/>)
         utils.notAQuery(/hello/i)
     `,
+
+    // issue #565
+    {
+      code: `
```
timdeschryver, line 84: The other valid snippets only contain the code block.

timdeschryver (dismissed):
Oh wow, how did I miss that 😅
Thanks for fixing my mistake!

tests/lib/rules/no-global-regexp-flag-in-query.test.ts
Belco90: Ah yes, I left this way because I was using the `only` property for debugging. I'll change it later.

timdeschryver (approved):

PR #569, opened 2022-04-11
feat(no-await-sync-events): add eventModules option

### LINKED ISSUE #567: `testing-library/no-await-sync-events` reporting false positives for `@testing-library/user-event` v14
### Plugin version

5.0.5

### ESLint version

8.12.0

### Node.js version

16.4.2

### npm/yarn version

8.5.0

### Operating system

macOS Monterey, 12.3

### Bug description

In version 14, `@testing-library/user-event` changed all APIs to always return a Promise.

See https://github.com/testing-library/user-event/releases/tag/v14.0.0

### Steps to reproduce

Use the async APIs in `@testing-library/user-event` v14 and see all the false positives reported by this rule

### Error output/screenshots

```
error  `userEvent.click` is sync and does not need `await` operator  testing-library/no-await-sync-events
```

### ESLint configuration

{
  "testing-library/no-await-sync-events": "error",
}

### Rule(s) affected

testing-library/no-await-sync-events

### Anything else?

_No response_

No

Belco90:
## Checks

## Changes

- Adding `eventModules` option to be able to disable reporting "fire-event" or "user-event"

## Context

Closes #567

### DIFF
--- a/lib/rules/no-await-sync-events.ts
+++ b/lib/rules/no-await-sync-events.ts
@@ -49,7 +49,9 @@ export default createTestingLibraryRule<Options, MessageIds>({
   },
   defaultOptions: [{ eventModules: VALID_EVENT_MODULES }],
 
-  create(context, _, helpers) {
+  create(context, [options], helpers) {
+    const { eventModules = VALID_EVENT_MODULES } = options;
+
     // userEvent.type() and userEvent.keyboard() are exceptions, which returns a
     // Promise. But it is only necessary to wait when delay option other than 0
     // is specified. So this rule has a special exception for the case await:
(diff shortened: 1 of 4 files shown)


timdeschryver (commented):
This LGTM (if the test cases are passing).

Does this also take the `setup` into account?
Because with v14, it isn't recommended to directly use userEvents from the imported object, but you should rather create your own userEvent instance.

https://testing-library.com/docs/user-event/setup#starting-a-session-per-setup

Belco90:
> Does this also take the setup into account?
> Because with v14, it isn't recommended to directly use userEvents from the imported object, but you should rather create your own userEvent instance.
> 
> https://testing-library.com/docs/user-event/setup#starting-a-session-per-setup

That's a good point. It's not addressed in this PR, but we should update all the internal detection helpers to consider this new way of setting up the user-event module.

timdeschryver (approved):
LGTM

lib/rules/no-await-sync-events.ts
```diff
@@ -32,11 +35,23 @@ export default createTestingLibraryRule<Options, MessageIds>({
       noAwaitSyncEvents:
         '`{{ name }}` is sync and does not need `await` operator',
     },
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          eventModules: {
+            enum: VALID_EVENT_MODULES,
```
MichaelDeBoey, line 43: @Belco90 Is it an `enum` or an array of enum values? 🤔

`schema` says enum, default values says array of enum values.

Belco90: If I understood it correctly, the enum allows to specify an array? Maybe I'm wrong and I ser the incorrect type.

MichaelDeBoey: Tests are also using a value instead of array btw

PR #570, opened 2022-04-12
fix(no-await-sync-events): define option schema correctly

Belco90:
## Checks

## Changes

- Define `eventModules` option schema correctly (array of enum instead of just enum)

## Context

Relates to https://github.com/testing-library/eslint-plugin-testing-library/pull/569#discussion_r847850631

### DIFF
--- a/lib/rules/no-await-sync-events.ts
+++ b/lib/rules/no-await-sync-events.ts
@@ -40,7 +40,11 @@ export default createTestingLibraryRule<Options, MessageIds>({
         type: 'object',
         properties: {
           eventModules: {
-            enum: VALID_EVENT_MODULES,
+            type: 'array',
+            minItems: 1,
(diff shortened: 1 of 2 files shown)


lib/rules/no-await-sync-events.ts
```diff
@@ -40,7 +40,11 @@ export default createTestingLibraryRule<Options, MessageIds>({
         type: 'object',
         properties: {
           eventModules: {
-            enum: VALID_EVENT_MODULES,
+            type: 'array',
+            minItems: 1,
+            items: {
+              enum: VALID_EVENT_MODULES,
+            },
```
Belco90, line 47: You were absolutely @MichaelDeBoey! Good catch 👀

MichaelDeBoey (approved):

PR #582, opened 2022-05-11
test: refactor `@marko/testing-library` tests

MichaelDeBoey:
Follow-up of #572.

All rules that are included in the `marko` config should test that the rule is working for `@marko/testing-library` as well.

To do so for #572, @PrashantAshok copied over some tests and replaced the import, but didn't include tests for the following rules:
- `await-async-utils`
- `no-promise-in-fire-event`
- `no-render-in-setup`
- `no-wait-for-snapshot`
- `prefer-find-by`

The way I presented in the PR, is to map over an array of the already tested framework + `@marko/testing-library`.
This PR is only doing that.

I'll create a separate PR to test against all testing framework that the rules support.

### DIFF
--- a/tests/create-testing-library-rule.test.ts
+++ b/tests/create-testing-library-rule.test.ts
@@ -442,40 +442,26 @@ ruleTester.run(RULE_NAME, rule, {
         },
       ],
     },
-    {
-      code: `
+    ...['@testing-library/react', '@marko/testing-library'].map(
+      (testingFramework) =>
+        ({
+          code: `
       // case: render imported from Testing Library module
-      import { render } from '@testing-library/react'
+      import { render } from '${testingFramework}'
       import { somethingElse } from 'another-module'
       const foo = require('bar')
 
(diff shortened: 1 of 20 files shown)


timdeschryver (dismissed):

Belco90:
What happened with the previous review? 🤔

timdeschryver (approved):

timdeschryver:
I think more code got refactored 😁

MichaelDeBoey:
All should be fine now 🙈

PrashantAshok:
> I think more code got refactored 😁

Exactly. 

Good work @MichaelDeBoey. (btw wonder what's up with specific name calling)

MichaelDeBoey:
> btw wonder what's up with calling out my name specifically

@PrashantAshok I just wanted to provide context & let you know about this PR, that's all 🤷‍♂️

Belco90 (approved):

PR #584, opened 2022-05-13
test: refactor `@testing-library/angular` tests

MichaelDeBoey:
Follow-up of #582 & #583

Added explicit tests for `@testing-library/angular`

### DIFF
--- a/lib/rules/no-dom-import.ts
+++ b/lib/rules/no-dom-import.ts
@@ -12,6 +12,14 @@ const DOM_TESTING_LIBRARY_MODULES = [
   '@testing-library/dom',
 ];
 
+const correctModuleNameByFramework = {
+  angular: '@testing-library/angular', // ATL is *always* called `@testing-library/angular`
+  marko: '@marko/testing-library', // Marko TL is called `@marko/testing-library`
+};
+const getCorrectModuleName = (moduleName: string, framework: string): string =>
+  correctModuleNameByFramework[framework] ||
+  moduleName.replace('dom', framework);
+
 export default createTestingLibraryRule<Options, MessageIds>({
   name: RULE_NAME,
   meta: {
@@ -33,11 +41,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
         'import from DOM Testing Library is restricted, import from {{module}} instead',
     },
     fixable: 'code',
-    schema: [
-      {
-        type: 'string',
-      },
-    ],
+    schema: [{ type: 'string' }],
   },
   defaultOptions: [''],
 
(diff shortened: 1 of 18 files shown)


lib/rules/no-dom-import.ts
```diff
@@ -49,8 +49,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
       if (framework) {
         // marko TL is called @marko/testing-library
         const correctModuleName =
-          framework === 'marko'
-            ? moduleName.replace('dom-', `@${framework}/`)
+          framework === 'angular'
```
timdeschryver, line 52: I'm not sure why this is needed though 😅

timdeschryver (approved):

lib/rules/no-dom-import.ts
MichaelDeBoey: Because `dom-testing-library` would be replaced with `dom-testing-library` otherwise, which never existed.

lib/rules/no-dom-import.ts
```diff
@@ -49,8 +49,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+            ? '@testing-library/angular'
+            : framework === 'marko'
+            ? '@marko/testing-library'
             : moduleName.replace('dom', framework);
```
Belco90, line 56: Can you switch this nested ternaries to a more readable if-else statement?

tests/lib/rules/no-dom-import.test.ts
```diff
@@ -4,29 +4,50 @@ import { createRuleTester } from '../test-utils';
...
   valid: [
     'import { foo } from "foo"',
     'import "foo"',
-    ...SUPPORTED_TESTING_FRAMEWORKS.flatMap((testingFramework) => [
-      `import { fireEvent } from "${testingFramework}"`,
-      `import * as testing from "${testingFramework}"`,
-      `import "${testingFramework}"`,
-    ]),
+    ...SUPPORTED_TESTING_FRAMEWORKS.flatMap(({ oldName, newName }) =>
+      [oldName, newName].flatMap((testingFramework) => [
```
Belco90, line 34: I don't quite understand the purpose of the `oldName` and `newName` from supported testing frameworks.

Belco90 (changes requested):

lib/rules/no-dom-import.ts
Belco90: I don't get it. Is this an issue only for Angular TL? If so, why?

Belco90: Ah, I bet you meant _Because dom-testing-library would be replaced with **angular**-testing-library_

Belco90: Can you add a comment explaining this? Also, the Marko comment is not in the right place now.

tests/lib/rules/no-dom-import.test.ts
MichaelDeBoey: `oldName` is like `react-testing-library`, so before the packages were under the `@testing-library` scope.

lib/rules/no-dom-import.ts
```diff
@@ -12,6 +12,14 @@ const DOM_TESTING_LIBRARY_MODULES = [
   '@testing-library/dom',
 ];
 
+const correctModuleNameByFramework = {
+  angular: '@testing-library/angular', // ATL is *always* called `@testing-library/angular`
+  marko: '@marko/testing-library', // Marko TL is called `@marko/testing-library`
+};
+const getCorrectModuleName = (moduleName: string, framework: string): string =>
+  correctModuleNameByFramework[framework] ||
+  moduleName.replace('dom', framework);
```
Belco90, line 21: Great 😎

Belco90 (approved):

PR #588, opened 2022-05-17
test: cleanup tests

MichaelDeBoey:
Follow-up of #582, #583, #584, #585 & #587

### DIFF
--- a/tests/lib/rules/no-await-sync-events.test.ts
+++ b/tests/lib/rules/no-await-sync-events.test.ts
@@ -88,6 +88,13 @@ const FIRE_EVENT_FUNCTIONS = [
   'gotPointerCapture',
   'lostPointerCapture',
 ];
+const SUPPORTED_TESTING_FRAMEWORKS = [
+  '@testing-library/dom',
+  '@testing-library/angular',
+  '@testing-library/react',
+  '@testing-library/vue',
+  '@marko/testing-library',
+];
 const USER_EVENT_SYNC_FUNCTIONS = [
   'clear',
   'click',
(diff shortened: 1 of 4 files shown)


tests/lib/rules/no-await-sync-events.test.ts
```diff
@@ -168,15 +175,17 @@ ruleTester.run(RULE_NAME, rule, {
...
-        test('should not report fireEvent.${func} sync event awaited', async() => {
-          await fireEvent.${func}('foo');
-        });
-      `,
-      options: [{ eventModules: ['user-event'] }],
-    })),
+    ...SUPPORTED_TESTING_FRAMEWORKS.flatMap((testingFramework) =>
+      FIRE_EVENT_FUNCTIONS.map((func) => ({
+        code: `
+          import { fireEvent } from '@testing-library/${testingFramework}';
```
timdeschryver, line 181: Shouldn't this be the following?
Otherwise, we end up with `@testing-library/@testing-library/dom`

```suggestion
          import { fireEvent } from '${testingFramework}';
```

Belco90:
Thinking about all the changes we did to test all rules for the corresponding Testing Library frameworks, I think it makes sense to add some improvements to our `TestingLibraryRuleMeta` and `createRuleTester` to automatically generate the test cases variations for indicated frameworks.

In my mind, it could look like this:

```js

// rules/foo.js
export default createTestingLibraryRule({
  // ...
  meta: {
    // ...
    supportedFrameworks: ['react', 'vue']
    // ^ we extend our TestingLibraryRuleMeta so it has this new required property,
    // which can be used also to autogenerate a table of compatible frameworks for each rule doc
  }
})

// test/lib/rules/foo.js
ruleTester.run(RULE_NAME, rule, {
  valid: [
    {
      code: `
      import { fireEvent } from '{TESTING_LIBRARY_FRAMEWORK}'
      //                                          ^ this placeholder gets replaced automatically,
      //                                             iterating over each value from supportedFrameworks
    
      await fireEvent.click(element)
      `
    }
  ]
})
```

I don't want to hijack the PR, so if you think this could be useful to avoid repetition, improve test maintenance, and make it easier to contribute in the future, I'll move the discussion to a new issue.

MichaelDeBoey:
@Belco90 We could even derive `supportedFrameworks` from the `recommendedConfig` meta.

We should even be able to run all rules on all testing frameworks though.
I currently only did this when the rule is available in the config for that specific testing framework.

Belco90 (approved):
LGTM!

Issue #590, opened 2022-05-19
Bug: `await-fire-event` doesn't show the fireEvent method correctly in the error message


Belco90:
### Have you read the Troubleshooting section?

Yes

### Plugin version

v5.5.0

### ESLint version

v8.14.0

### Node.js version

v16.15.0

### package manager and version

npm v8.9.0

### Operating system

macOS v12.3.1

### Bug description

Errors reported by `await-fire-event` rule are not indicating the `fireEvent` method name correctly in the message.

### Steps to reproduce

```js
test('increments value on click', async () => {
  render(App);

  expect(screen.getByText('Times clicked: 0')).toBeDefined();

  const button = screen.getByText('increment');

  fireEvent.click(button);

  expect(screen.getByText('Times clicked: 1')).toBeDefined();
});
```

### Error output/screenshots

![CleanShot 2022-05-20 at 00 14 50](https://user-images.githubusercontent.com/2677072/169413320-39596c6b-27af-44f7-8d4f-8f248ef35cea.png)

### ESLint configuration

N/A

### Rule(s) affected

`await-fire-event`

### Do you want to submit a pull request to fix this bug?

No

IvanAprea:
Hello, can i propose a solution for this?

Belco90:
@IvanAprea yes please! PRs are more than welcome.

PR #591, opened 2022-05-20
[ISSUE-509] show wait-fire-event correctly

IvanAprea:
## Checks

## Changes

Issue link: https://github.com/testing-library/eslint-plugin-testing-library/issues/590

- I changed the waitFireEvent prop from "methodName" to "name"

image
I am showing the error with gitLens

### DIFF
--- a/lib/rules/await-fire-event.ts
+++ b/lib/rules/await-fire-event.ts
@@ -29,7 +29,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
     },
     messages: {
       awaitFireEvent:
-        'Promise returned from `fireEvent.{{ methodName }}` must be handled',
+        'Promise returned from `fireEvent.{{ name }}` must be handled',
       fireEventWrapper:
         'Promise returned from `fireEvent.{{ wrapperName }}` wrapper over fire event method must be handled',
     },


IvanAprea:
@Belco90 :eyes:

lib/rules/await-fire-event.ts
```diff
@@ -29,7 +29,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
     },
     messages: {
       awaitFireEvent:
-        'Promise returned from `fireEvent.{{ methodName }}` must be handled',
+        'Promise returned from `fireEvent.{{ name }}` must be handled',
       fireEventWrapper:
         'Promise returned from `fireEvent.{{ wrapperName }}` wrapper over fire event method must be handled',
```
Belco90, line 34: I think we have to do the same change here, right?

Belco90 (changes requested):
Thanks for taking care of this!

I think we need to do the change in the other message. Additionally, can you update the tests for this rule so they check that the correponding `name` property is passed to the messages?

lib/rules/await-fire-event.ts
IvanAprea: I think the same, but change wrapperName to name breaks the tests, i will try to find the problem and fix that :D

Belco90: Let me know if you need help for this!

Belco90: You are right! It should be the name of the wrapper 👌

IvanAprea:
@Belco90 I did another commit with the changes for the messageId: fireEventWrapper. I changed the var to name and fixed the test in order to use the wrapper function name. I saw the rest of tests and i think that they are well. The tests didnt fail before because the var name in the messages was wrong. Please let me know if you have any doubt or if you need that i test anymore

Belco90:
@all-contributors please add @IvanAprea for code and tests

PR #601, opened 2022-06-06
docs(prefer-presence-queries): add a note to discourage usage of options

zaicevas:
## Checks

## Changes

- Added a note in `prefer-presence-queries` that discourages the use of options

## Context

In https://github.com/testing-library/eslint-plugin-testing-library/pull/557 2 `prefer-presence-queries` options were added due to false positives described in https://github.com/testing-library/eslint-plugin-testing-library/issues/518.

As I understand, these options are a short-term workaround. They will be removed when false positives are fixed. If that's the case, I think it's worth outlining this in rule docs.

### DIFF
--- a/docs/rules/prefer-presence-queries.md
+++ b/docs/rules/prefer-presence-queries.md
@@ -61,10 +61,10 @@ test('some test', async () => {
 
 ## Options
 
-| Option     | Required | Default | Details                                                                                                                                                      |
-| ---------- | -------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `presence` | No       | `true`  | If enabled, this rule will ensure `getBy*` is used to validate whether an element is present. If disabled, `queryBy*` will be accepted for presence queries. |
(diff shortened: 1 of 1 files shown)


docs/rules/prefer-presence-queries.md
```diff
@@ -61,10 +61,10 @@ test('some test', async () => {
...
-| `absence`  | No       | `true`  | If enabled, this rule will ensure `queryBy*` is used to validate whether an element is absent. If disabled, `getBy*` will be accepted for absence queries.   |
+| Option     | Required | Default | Details                                                                                                                                                                                                                                                                                                                                                                            |
+| ---------- | -------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
```
zaicevas, line 65: Auto-formatting did some magic 🙄

Belco90:
Thanks for adding this clarification!

PR #611, opened 2022-07-22
feat(no-node-access): add `allowContainerFirstChild` option

### LINKED ISSUE #593: Allow exceptions for "no-node-access"

no-node-access

Fewer warnings

`render` method returns container which has content rendered inside it (`div` element by default - [docs](https://testing-library.com/docs/react-testing-library/api#:~:text=Tip%3A%20To%20get%20the%20root%20element%20of%20your%20rendered%20element%2C%20use%20container.firstChild.) )

The easiest way to access content itself is by using `firstChild` property
This is hardly avoidable in some cases, such as:
* snapshot testing
* testing if rendering simply has/has not happened
* counting child elements

The rule could ignore `firstChild`(only?) in some cases.

### Example code

```
const MyComponent = () => <div>Any kind of stuff</div>
```

```
describe('some test', () => {
  it('test case', () => {
    const { container } = render(<MyComponent />)

    expect(container.firstChild).toMatchSnapshot()
  })
})
```

Currently rule highlights `container.firstChild` part as a mistaken direct Node access

Won't highlight it

### Anything else?

_No response_

No

Semigradsky:
## Checks

## Changes

- Add `allowContainerFirstChild` option to `no-node-access` rule

## Context

Fixes #593

### DIFF
--- a/docs/rules/no-node-access.md
+++ b/docs/rules/no-node-access.md
@@ -51,6 +51,25 @@ within(signinModal).getByPlaceholderText('Username');
 document.getElementById('submit-btn').closest('button');
 ```
 
+## Options
+
+This rule has one option:
+
+- `allowContainerFirstChild`: **disabled by default**. When we have container
+  with rendered content then the easiest way to access content itself is [by using
+  `firstChild` property](https://testing-library.com/docs/react-testing-library/api/#container-1). Use this option in cases when this is hardly avoidable.
+
+  ```js
+  "testing-library/no-node-access": ["error", {"allowContainerFirstChild": true}]
+  ```
+
+Correct:
+
+```jsx
+const { container } = render(<MyComponent />);
+expect(container.firstChild).toMatchSnapshot();
+```
+
 ## Further Reading
 
 ### Properties / methods that return another Node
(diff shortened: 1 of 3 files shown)


docs/rules/no-node-access.md
```diff
@@ -51,6 +51,25 @@ within(signinModal).getByPlaceholderText('Username');
 document.getElementById('submit-btn').closest('button');
 ```
 
+## Options
+
+This rule has one option:
+
+- `allowContainerFirstChild`: **disabled by default**. When we have container
+  with rendered content then the easiest way to access content itself is by using
+  `firstChild` property. Use this option in cases when this is hardly avoidable.
```
Belco90, line 60: Could you add a link to the Testing Library docs where this is recommended? https://testing-library.com/docs/react-testing-library/api/#container-1

Belco90 (changes requested):
Thanks for your contribution @Semigradsky!

LGTM, just a small improvement in the rule doc and should be ready to go.

docs/rules/no-node-access.md
Semigradsky: Added

Belco90 (approved):

Belco90:
@all-contributors please add @Semigradsky for code, test and docs

PR #633, opened 2022-08-18
style: indent lines with tabs instead of spaces

Belco90:
## Checks

## Changes

- Configure prettier to indent lines with tabs rather than spaces
- Set up lint-staged using a js config file, so ignored files can be excluded properly

## Context

There was no ticket for this. However, [Prettier is considering switching to tabs by default in v3](https://twitter.com/rich_harris/status/1541761871585464323?s=12&t=Pz1PyMfaUUv8tP2HxYX65w). This is a great change, you can find the motivation behind it here: https://github.com/prettier/prettier/issues/7475

I think we should switch to tabs now to improve the accessibility of the code. The tab size can be customized in GitHub here: https://github.com/settings/appearance


lint-staged.config.js
```diff
@@ -0,0 +1,19 @@
...
+const removeIgnoredFiles = async (files) => {
+	const eslint = new ESLint();
+	const ignoredFiles = await Promise.all(
+		files.map((file) => eslint.isPathIgnored(file))
+	);
+	const filteredFiles = files.filter((_, i) => !ignoredFiles[i]);
+	return filteredFiles.join(' ');
+};
+
+module.exports = {
```
Belco90, line 13: I had to move lint-staged config from `.lintstagedrc` to `lint-staged.config.js` so files from `.eslintignore` can be ignored properly. [This is the official way to do so from lint-staged docs](https://github.com/okonet/lint-staged#how-can-i-ignore-files-from-eslintignore).

lint-staged.config.js
```diff
@@ -0,0 +1,19 @@
...
+	'*.{js,ts}': async (files) => {
+		const filesToLint = await removeIgnoredFiles(files);
+		return [`eslint --max-warnings=0 ${filesToLint}`];
+	},
+	'*': 'prettier --write --ignore-unknown',
```
timdeschryver, line 18: TIL about `ignore-unknown`, thanks!

lint-staged.config.js
```diff
@@ -0,0 +1,19 @@
...
+	const eslint = new ESLint();
+	const ignoredFiles = await Promise.all(
+		files.map((file) => eslint.isPathIgnored(file))
+	);
+	const filteredFiles = files.filter((_, i) => !ignoredFiles[i]);
+	return filteredFiles.join(' ');
+};
+
+module.exports = {
+	'*.{js,ts}': async (files) => {
```
timdeschryver, line 14: Previously `jest --findRelatedTests` was also run, was it intended to remove that?

Belco90: I forgot to mention this! Yes, I did that on purpose indeed. That way, the pre-commit tasks are as fast as possible.

MichaelDeBoey (approved):
LGTM!

Not sure if we already need to do it before the Prettier discussion is resolved though. 🤔
I would personally wait for it & do it together with the Prettier v3 update.

Belco90:
> Not sure if we already need to do it before the Prettier discussion is resolved though. 🤔
> I would personally wait for it & do it together with the Prettier v3 update.

Not sure what Prettier discussion you are referring to, but [the original issue for v3 is still open](https://github.com/prettier/prettier/issues/7475).

Since [the first alpha for Prettier v3 was released a few days ago](https://github.com/prettier/prettier/releases/tag/3.0.0-alpha.0) with no change related to using tabs, I'd just switch to tabs in our codebase. If they do the change in v3, we are ready for it. If not, we made our codebase more accessible, even if Prettier didn't change its defaults!

timdeschryver (approved):

PR #639, opened 2022-09-01
fix: `prefer-query-by-disappearance` error line and column to the beginning of `get` or `find` queries

### LINKED ISSUE #600: Bug: `prefer-query-by-disappearance` reports error in an incorrect part of the code

Yes

### Plugin version

v5.5.0

### ESLint version

v8.14.0

### Node.js version

v16.15.0

npm v8.9.0

### Operating system

macOS v12.3.1

### Bug description

When there is an error reported by `prefer-query-by-disappearance`, it's displayed at the very beginning of the callback passed to `waitForElementToBeRemoved`. Instead, the error should be placed in the query within the callback.

### Steps to reproduce

```js
it('should do something', async () => {
  render(<App />)

  fireEvent.click(screen.getByRole('button'))

  await waitForElementToBeRemoved(() => screen.getByText('Loading'))
})
```

### Error output/screenshots

CleanShot 2022-06-03 at 01 23 52

### ESLint configuration

N/A

### Rule(s) affected

`prefer-query-by-disappearance`

### Anything else?

_No response_

No

sjarva:
## Checks

## Changes
- refactor reporting an error deeper, where the `get` or `find` query variants are found
- refactor some function names to convey that they also check/report the error, not just return boolean based on if an error should be reported
- update unit tests to have correct error line and column

## Context
Closes #600

### DIFF
--- a/lib/rules/prefer-query-by-disappearance.ts
+++ b/lib/rules/prefer-query-by-disappearance.ts
@@ -51,14 +51,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
 		}
 
 		/**
-		 * Checks if a node is a query node and starts with "get" or "find".
+		 * Checks if node is reportable (starts with "get" or "find") and if it is, reports it with `context.report()`.
 		 *
 		 * @param {TSESTree.LeftHandSideExpression} node - Node to be tested
-		 * @returns {Boolean} True if node should be reported and reports it with `context.report()`. False if node should not be reported.
+		 * @returns {Boolean} Boolean indicating if expression was reported
 		 */
-		function isReportableExpression(
-			node: TSESTree.LeftHandSideExpression
-		): boolean {
+		function reportExpression(node: TSESTree.LeftHandSideExpression): boolean {
 			const argumentProperty = isMemberExpression(node)
 				? getPropertyIdentifierNode(node.property)
 				: getPropertyIdentifierNode(node);
@@ -92,7 +90,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
 				return false;
 			}
 
-			return isReportableExpression(node.callee);
+			return reportExpression(node.callee);
 		}
 
 		function isReturnViolation(node: TSESTree.Statement) {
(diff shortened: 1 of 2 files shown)


lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -50,7 +50,15 @@ export default createTestingLibraryRule<Options, MessageIds>({
 			return helpers.isAsyncUtil(identifierNode, ['waitForElementToBeRemoved']);
 		}
 
-		function isReportableExpression(node: TSESTree.LeftHandSideExpression) {
+		/**
+		 * Checks if a node is a query node and starts with "get" or "find".
+		 *
+		 * @param {TSESTree.LeftHandSideExpression} node - Node to be tested
+		 * @returns {Boolean} True if node should be reported and reports it with `context.report()`. False if node should not be reported.
+		 */
```
Belco90, line 58: Thanks for adding the jsdoc!

lib/rules/prefer-query-by-disappearance.ts
```diff
@@ -59,13 +67,20 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
-			return (
+			if (
 				helpers.isGetQueryVariant(argumentProperty) ||
 				helpers.isFindQueryVariant(argumentProperty)
-			);
+			) {
+				context.report({
+					node: argumentProperty,
+					messageId: 'preferQueryByDisappearance',
+				});
```
Belco90, line 77: I find it quite confusing that the function is just checking if the expression is reportable, but if so it reports it in here. Could you rename the function, so it's called something like `reportExpression`, and then indicate the behavior in the jsdoc? Something simple, like _"Checks if node is reportable and report it if so. Returns boolean indicating if was reported"_.

Belco90 (changes requested):
Hi @sjarva! Thanks for your first contribution.

LGTM, just a small change request so the purpose of one of the function is clearer.

sjarva:
Thanks for the review @Belco90 😊 I replied to your change request, hope that the function name and JSDoc comment are better now 🤞

lib/rules/prefer-query-by-disappearance.ts
sjarva: Good point. I re-worded both a bit and to the direction you suggested. Is it any better now?

Belco90: Looks great now!

Belco90 (approved):
Amazing, merging now!

Belco90:
@all-contributors please add @sjarva for code and test

Issue #626, opened 2022-08-04
Transform `await-fire-event` into `await-async-events`


Belco90:
Transform `await-fire-event` into `await-async-events`.

Since user-events v14 is async, we should transform this rule to consider both fire and user events.

It needs a new option to decide which modules are enabled or disabled:
- fire-event: disabled by default since it's only used in Vue (enabled in Vue shareable config tho)
- user-event: enabled by default since user-event 14 is async

Explain in the rule's doc that the user-event option must be disabled when working with user-event prior to v14

skovy:
@Belco90 @MichaelDeBoey assuming this was supposed to be plural? I can open a PR to fix it since it's currently `await-async-event`

Belco90:
> @Belco90 @MichaelDeBoey assuming this was supposed to be plural? I can open a PR to fix it since it's currently `await-async-event`

I didn't realize it wasn't plural! Please, open a small PR to rename it 🙏

PR #641, opened 2022-09-07
fix: `no-await-sync-events` false positive

### LINKED ISSUE #403: False positive for testing-library/no-await-sync-events
Using `await` with `userEvent.type`'s `delay` option is correct, but is incorrectly flagged as a lint error.

```js
async () => {
  const delay = 10
  // error  `userEvent.type` is sync and does not need `await` operator  testing-library/no-await-sync-events
  await userEvent.type(element, text, {delay})
}
  ```

sjarva:
## Checks

## Changes

- Check if `delay` is declared or assigned a value elsewhere than in the call expression's arguments
- Add test cases that have declarations and assignments

## Context

Closes #403

### DIFF
--- a/lib/rules/no-await-sync-events.ts
+++ b/lib/rules/no-await-sync-events.ts
@@ -56,6 +56,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
 
 	create(context, [options], helpers) {
 		const { eventModules = VALID_EVENT_MODULES } = options;
+		let hasDelayDeclarationOrAssignmentGTZero: boolean;
 
 		// userEvent.type() and userEvent.keyboard() are exceptions, which returns a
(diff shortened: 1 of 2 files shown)


sjarva:
Oh wow, that must have been the fastest review in my life @Belco90 🥳 🎈 Thank you so much!

Belco90:
It was a coincidence to be honest! I opened GitHub to check something else, and I saw this short PR 🚀

ph-fritsche:
With `user-event@14` this rule should probably be removed as all APIs are asynchronous.

Belco90:
> With `user-event@14` this rule should probably be removed, as all APIs are asynchronous.

It won't be removed yet, but we planned to adapt another rule to handle async user event method properly https://github.com/testing-library/eslint-plugin-testing-library/issues/626

Issue #500, opened 2021-10-17
no-wait-for-side-effects erroring in promise


Brianzchen:
### Plugin version

v4.10.1

### ESLint version

v7.31.0

### Node.js version

v12.17.0

### npm/yarn version

yarn 1.22.10

### Operating system

macOS 10.14.6

### Bug description

When I use `fireEvent` within a `.then()` of `waitFor` it throws an error

### Steps to reproduce

Basically what's shown in the screenshot. I prefer to use `.then()` as opposed to async await but it throws errors when it shouldn't

<img width="1109" alt="Screen Shot 2021-10-18 at 10 13 09 am" src="https://user-images.githubusercontent.com/12436524/137648599-cc21bd0e-4329-4026-a5cf-d37d6bb8f668.png">

### ESLint configuration

A few private stuff but the rule sets are 
```
    'testing-library/no-dom-import': 2,
    'testing-library/no-manual-cleanup': 2,
    'testing-library/no-wait-for-empty-callback': 2,
    'testing-library/no-wait-for-side-effects': 2,
    'testing-library/prefer-screen-queries': 2,
    'testing-library/prefer-wait-for': 2,
```

### Rule(s) affected

testing-library/no-wait-for-side-effects

### Do you want to submit a pull request to fix this bug?

No

sjarva:
Thanks @Brianzchen for reporting this! This was nice and easy to fix in #645 🛠️ 🤗

PR #645, opened 2022-09-12
fix(no-wait-for-side-effects): false positive inside `.then()`

sjarva:
## Checks

## Changes

- when reporting side efdects, check if the node is a `.then()`, and if it is, not report it
- added unit tests
- adds a mention of edge case to rule documentation

## Context

Closes: #500

### DIFF
--- a/lib/rules/no-wait-for-side-effects.ts
+++ b/lib/rules/no-wait-for-side-effects.ts
@@ -1,4 +1,4 @@
-import { ASTUtils, TSESTree } from '@typescript-eslint/utils';
+import { TSESTree } from '@typescript-eslint/utils';
 
 import { createTestingLibraryRule } from '../create-testing-library-rule';
 import {
@@ -8,7 +8,7 @@ import {
 	isAssignmentExpression,
 	isCallExpression,
 	isSequenceExpression,
-	isMemberExpression,
+	hasThenProperty,
 } from '../node-utils';
 
 export const RULE_NAME = 'no-wait-for-side-effects';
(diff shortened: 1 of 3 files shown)


tests/lib/rules/no-wait-for-side-effects.test.ts
```diff
@@ -722,6 +735,41 @@ ruleTester.run(RULE_NAME, rule, {
       `,
 				errors: [{ line: 4, column: 11, messageId: 'noSideEffectsWaitFor' }],
 			} as const,
+			{
+				// Issue #500, https://github.com/testing-library/eslint-plugin-testing-library/issues/500
+				code: `
+        import { waitFor } from '${testingFramework}';
+        await waitFor(function() {
```
Belco90, line 742: Could you remove the `await` operator from the `waitFor` utils chained with `then`? So we can make sure the example is valid because thenable promises are handled properly, and not because it was a side-effect from the `wait` operator.

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -56,6 +57,27 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+
+			const callExpressionNode = node.parent.parent as TSESTree.CallExpression;
+
+			const test =
+				isMemberExpression(callExpressionNode.callee) &&
+				ASTUtils.isIdentifier(callExpressionNode.callee.property) &&
+				callExpressionNode.callee.property.name === 'then';
+
+			return test;
+		}
```
Belco90, line 79: I think you can reuse our [`hasChainedThen`](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/lib/node-utils/index.ts#L128) to simplify the chained then detection.

Belco90 (changes requested):
Thanks for fixing this one too! Just a couple of small changes and should be ready to go.

tests/lib/rules/no-wait-for-side-effects.test.ts
sjarva: Good catch! Done.

lib/rules/no-wait-for-side-effects.ts
sjarva: Good point this too. I couldn't get to work with `hasChainedThen`, but `hasThenProperty` did the job.

tests/lib/rules/no-wait-for-side-effects.test.ts
sjarva: Removed the `await` also from the rule documentation example.

Belco90 (approved):
All good!

Issue #627, opened 2022-08-04
Remove `prefer-wait-for` rule


Belco90:
Remove `prefer-wait-for` rule, since `wait`, `waitForElement` and `waitForDomChange` have been removed already

Belco90:
I just realized there are other rules like `prefer-find-by` that are still checking these async methods. I'm going to add another ticket to check other places and remove such checks for removed async utils.

PR #648, opened 2022-09-17
feat: remove `prefer-wait-for` rule

skovy:
BREAKING CHANGE: remove prefer-wait-for rule since the deprecated helpers were removed

## Checks

## Changes

- Removes the `prefer-wait-for` rule since the deprecated helpers were removed https://github.com/testing-library/dom-testing-library/pull/966

Not sure if there is a `v6` or `alpha` branch?

## Context

Resolves https://github.com/testing-library/eslint-plugin-testing-library/issues/627


Belco90:
Hey @skovy, thanks for your contribution!

I'm going to create v6 so we can target this PR to that branch.

MichaelDeBoey:
@Belco90 `alpha` should be created instead so we can download latest changes from npm to test them out locally in our own projects

MichaelDeBoey (approved):

Belco90:
> @Belco90 `alpha` should be created instead so we can download latest changes from npm to test them out locally in our own projects

Even better, thanks for that!

Belco90 (approved):

Issue #625, opened 2022-08-04
Rename `no-render-in-setup`


Belco90:
Rename `no-render-in-setup` to `no-render-in-lifecycle` so it's not confused with preventing [renders nested in a setup function](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing#apply-aha-avoid-hasty-abstractions).

PR #649, opened 2022-09-20
fix: rename no-render-in-setup to no-render-in-lifecycle

skovy:
BREAKING CHANGE: rename no-render-in-setup to no-render-in-lifecycle

## Checks

## Changes

- Rename `no-render-in-setup` to `no-render-in-lifecycle` so it's not confused with preventing [renders nested in a setup function](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing#apply-aha-avoid-hasty-abstractions).

## Context

Resolves https://github.com/testing-library/eslint-plugin-testing-library/issues/625.


Belco90 (approved):
LGTM, thanks again!

PR #652, opened 2022-09-29
feat: rename `await-fire-event` to `await-async-event` and support `user-event`

skovy:
BREAKING CHANGE: rename `await-fire-event` to `await-async-event` and support `user-event`

## Checks

## Changes

- Update the existing `await-fire-event` rule to become the `await-async-event` rule
- Support both user event and fire event via configuration, with the default being user event
- Update recommended configs to have `userEvent` for most, but `fireEvent` for those that have async fire event methods
- Update related documentation

## Context

Resolves https://github.com/testing-library/eslint-plugin-testing-library/issues/626


lib/rules/await-fire-event.ts
```diff
@@ -10,8 +10,13 @@ import {
 } from '../node-utils';
 
 export const RULE_NAME = 'await-fire-event';
+const EVENT_MODULES = ['fireEvent', 'userEvent'] as const;
```
Belco90, line 13: We have `EVENTS_SIMULATORS` exported in `utils`. Any problem with reusing that one?

lib/rules/await-fire-event.ts
```diff
@@ -33,11 +38,28 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
 		},
-		schema: [],
+		schema: [
+			{
+				type: 'object',
+				default: {},
+				additionalProperties: false,
+				properties: {
+					eventModule: {
+						type: 'string',
```
Belco90, line 48: Sorry, I think I didn't specify this really well in the original ticket. The idea is that this is a list accepting either "fire-event", "user-event", or both!

Frameworks like Vue Testing Library will need to report both, so they'll need to set this option to `["fire-event", "user-event"]`.

By default, it should be just `["user-event"]`

It should be fairly easy to adapt your current code tho.

Belco90 (commented):
Thanks for taking care of this! I left some comments.

skovy:
Thanks for the feedback, I'll take a look at those comments! Forgot about draft PRs, thanks for updating that too 🤦‍♂️

README.md
```diff
@@ -206,9 +206,9 @@ To enable this configuration use the `extends` property in your
 
 | Name                                                                                 | Description                                                                                  | 🔧  | Included in configurations                                                         |
 | ------------------------------------------------------------------------------------ | -------------------------------------------------------------------------------------------- | --- | ---------------------------------------------------------------------------------- |
+| [`await-async-event`](./docs/rules/await-async-event.md)                             | Enforce promises from `fireEvent` methods to be handled                                      |     | ![vue-badge][] ![marko-badge][]                                                    |
```
Belco90, line 209: Have you generated this with the scripts provided in the project? They are mentioned in the PR description, so this table is automatically generated from rules meta info. The description and badges for this one doesn't seem correct.

docs/rules/await-async-event.md
```diff
@@ -0,0 +1,143 @@
...
+// wrap a userEvent method within a function...
+function triggerEvent() {
+	return userEvent.click(button);
+}
+await triggerEvent(); // ...and handling promise from it is correct also
+
+// using `Promise.all` or `Promise.allSettled` with an array of promises is valid
+await Promise.all([
+	await userEvent.click(getByText('Click me'));
+	await userEvent.tripleClick(getByText('Click me'));
```
Belco90, line 93: When handling several promises with `Promise.all`, it's not necessary to handle the inner ones, but only the outer one. Not sure if this is inherited from the previous docs, but it's incorrect. Could you fix this, please?

docs/rules/await-async-event.md
```diff
@@ -0,0 +1,143 @@
...
+]);
+```
+
+## Options
+
+- `eventModule`: which event module should be linted for async event methods (`userEvent` and/or `fireEvent`). Defaults to `userEvent` which should be used after v14. `fireEvent` should only be used with frameworks that have async fire event methods.
```
Belco90, line 99: Could you specify the expected type? Which basically is string or array of strings.

lib/rules/await-async-event.ts
```diff
@@ -0,0 +1,163 @@
...
+	meta: {
+		type: 'problem',
+		docs: {
+			description: 'Enforce promises from async event methods are handled',
+			recommendedConfig: {
+				dom: ['error', { eventModule: 'userEvent' }],
+				angular: ['error', { eventModule: 'userEvent' }],
+				react: ['error', { eventModule: 'userEvent' }],
+				vue: ['error', { eventModule: 'fireEvent' }],
+				marko: ['error', { eventModule: 'fireEvent' }],
```
Belco90, line 35: Vue and Marko actually need both by default: we are aiming for userEvent v14 by default, and fireEvent is async in those frameworks:

```suggestion
				vue: ['error', { eventModule: ['fireEvent', 'userEvent'] }],
				marko: ['error', { eventModule: ['fireEvent', 'userEvent'] }],
```

Belco90 (changes requested):
Great job @skovy! I added a few comments, but should be easy to fix. I think after that it should be ready to go 🚀

Belco90 (changes requested):
Last request! Can you add some tests without setting the option so it takes the default value? This way we explicitly test the default options. Duplicating existing tests just without the option property is fine.

Belco90 (approved):

Issue #202, opened 2020-07-27
Support eslint fix command in fixable rules


renatoagds:
As suggested in #186 we could start to allow fix for rules that's possible to do that.

Rules that we could implement fix:

- [x] `await-async-query` -> #1077 
- [x] `no-await-sync-query` -> #1079 

**I'm working to map more rules that's open for that**

Belco90:
Are we sure we can implement the fix code for those with ease? They seem quite complex or even ambiguous. How do you plan to fix both of them?

I was thinking about adding fix code for some other rules that are more simple than those, like `await-async-query`, `await-async-utils`, `await-fire-event`or `no-await-sync-query`.

renatoagds:
>  How do you plan to fix both of them?

- In `no-side-effects-wait-for` I was thinking about moving only side effects before `waitFor`, since they probably do some action that's expected in `waitFor`
- In `no-multiple-assertions-wait-for` I was thinking about keep the first `expect` and move the others after `waitFor`.

Makes sense?

> I was thinking about adding fix code for some other rules that are more simple than those, like `await-async-query`, `await-async-utils`, `await-fire-event` or `no-await-sync-query`.

Good, will add them in the current list here

Belco90:
Those fix strategies sound fine! Let us know if you find any edge case when implementing any of them.

renatoagds:
Sorry guys .. Have been out for a long time, but will return helping the project, starting for this issue 🤘

renatoagds:
Did we had some change into the scenario exposed here? cc/ @Belco90

Belco90:
I don't think so, but these rules are gonna be refactor for v4 pretty soon. I don't know if it would be better to wait for v4 then? Not a major issue, but will cause conflicts on v4 of course.

renatoagds:
@Belco90 Makes sense!! That's something more that I could help? Maybe the rules refactor

Belco90:
I guess we have two options:
- implement this fixable code within v4 itself, so we implement as much as we can while finishing the v4. Then, we continue with fixable code for remaining rules after v4 release
- put more effort on v4 rules refactor and after that we go back to this ticket

Ideally I'd prefer not stopping adding/fixing things in current v3, but all modifications necessary for this issue will slow me down in v4 since I have to update v4 branches, adapt the code, check everything works as expected etc. And there would be **a lot** of conflicts when updating v4 with master.

[...]

PR #656, opened 2022-10-01
feat(await-async-event): add basic fixer

skovy:
## Checks

## Changes

- Add fixer for `await-async-event` rule

I was thinking if this fixer should update the `it`/`test` to also be `aysnc` if not, but that felt like out of the scope of this rule/fixer? It cares if it's awaited, not managing if it's an async function wrapping?

## Context

Branched off and a follow-up to https://github.com/testing-library/eslint-plugin-testing-library/pull/652 to make the migration easier to `user-event` v14.

Related to one of the rules mentioned in https://github.com/testing-library/eslint-plugin-testing-library/issues/202.


Belco90:
Waiting until #652 gets merged (you could point this PR to the branch from that PR to make explicit it's branched off from there).

I'm happy without managing the async operator from the outer function. I think other autofixes from this plugin work that way.

skovy:
> you could point this PR to the branch from that PR to make explicit it's branched off from there

The only branch options are the ones in this repo, I think since that pull request is a branch from my fork I can't base it on that branch? I can wait for that one to be merged and I can clean up this PR after.

Belco90 (approved):
@skovy thank you for adding the autofix implementation for this rule!

epmatsw:
Fwiw, I think this fixer should either 1) convert the function to async or 2) refuse to fix the issue if the function isn't async. As-is, it just produces syntactically invalid code and requires manual attention in those cases anyways.

Belco90:
> Fwiw, I think this fixer should either 1) convert the function to async or 2) refuse to fix the issue if the function isn't async. As-is, it just produces syntactically invalid code and requires manual attention in those cases anyways.

Actually… I checked again our current fixable rules and there is none adding the `await` operator (I don't know why, but I assumed there was one at least already). So I agree with @epmatsw: the fixer should add the async to the outer function (if found).

It should be fairly easy to do this. What do you think @skovy?

Belco90:
I've created #674 for improving the fixer. @epmatsw thanks for your thoughts on this!

skovy:
Thanks @Belco90 - I'll try to take a look at the issue over the next few days/weeks if someone doesn't get to it before that 👍

epmatsw:
No problem! Thanks @skovy and @Belco90 for the hard work, can’t wait to get started using this!

PR #657, opened 2022-10-01
fix(no-dom-imports): false negatives with several testing library imports

### LINKED ISSUE #586: Bug: `no-dom-import` false negatives with several Testing Library imports

Yes

### Plugin version

5.3.3

### ESLint version

v8.14.0

### Node.js version

v16.15.0

npm v8.9.0

### Operating system

macOS v12.3.1

### Bug description

If there are several Testing Library imports, `no-dom-import` doesn't report `@testing-library/dom` imports if they are other than the first occurrence between Testing Library imports.

### Steps to reproduce

This should be reported but it's not:
```js
import { screen } from '@testing-library/dom'
// ^ this import is not reported since it's not the first Testing Library import

import { App } from '../App'

it('should render a basic demo', () => {
  render(<App />)
  expect(screen.getByText('Hello Parcel + React!')).toBeInTheDocument()
})
```

### Error output/screenshots

N/A

### ESLint configuration

N/A

### Rule(s) affected

`no-dom-import`

### Anything else?

This is related to how we store the imported Testing Library module in the internal helpers. This only allows one Testing Library module, so the first one found it's the chosen one for later checks.

I remember mentioning this in another issue/PR, but I can't remember which one. This is not that simple to fix since we need to store a list of Testing Library modules used rather than a single one, updating all the helpers provided internally too.

After we allow storing several Testing Library modules in our core helpers, we will be able to fix the actual issue reported.

Yes

sjarva:
## Checks

## Changes
- adds a new util function `getAllTestingLibraryImportNodes` to get all `testing-library` related imports
- in `no-dom-import` loop through all the `testing-library` imports, and report an error on that import(s) that imports from `testing-library/dom`

## Context

Closes #586 

P.S. If you feel that this PR is following the rules of [Hactoberfest](https://hacktoberfest.com/) and this gets merged, it would be wonderful if this PR would be tagged with a `hacktoberfest-accepted` tag/label. Thanks!

### DIFF
--- a/lib/create-testing-library-rule/detect-testing-library-utils.ts
+++ b/lib/create-testing-library-rule/detect-testing-library-utils.ts
@@ -54,6 +54,7 @@ export type EnhancedRuleCreate<
 
 // Helpers methods
 type GetTestingLibraryImportNodeFn = () => ImportModuleNode | null;
+type GetTestingLibraryImportNodesFn = () => ImportModuleNode[];
 type GetCustomModuleImportNodeFn = () => ImportModuleNode | null;
 type GetTestingLibraryImportNameFn = () => string | undefined;
 type GetCustomModuleImportNameFn = () => string | undefined;
@@ -95,6 +96,7 @@ type IsNodeComingFromTestingLibraryFn = (
 
 export interface DetectionHelpers {
 	getTestingLibraryImportNode: GetTestingLibraryImportNodeFn;
+	getAllTestingLibraryImportNodes: GetTestingLibraryImportNodesFn;
 	getCustomModuleImportNode: GetCustomModuleImportNodeFn;
 	getTestingLibraryImportName: GetTestingLibraryImportNameFn;
 	getCustomModuleImportName: GetCustomModuleImportNameFn;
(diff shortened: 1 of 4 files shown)


lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -299,15 +301,20 @@ export function detectTestingLibraryUtils<
 
 		// Helpers for Testing Library detection.
 		const getTestingLibraryImportNode: GetTestingLibraryImportNodeFn = () => {
-			return importedTestingLibraryNode;
+			return importedTestingLibraryNodes[0];
```
Belco90, line 304: So clever! I was struggling to implement this change since I didn't want to refactor all this logic. Leaving it pointing to the first module found is a great workaround!

Belco90 (approved):
Great approach for fixing this! Label for hacktoberfest added.

lib/create-testing-library-rule/detect-testing-library-utils.ts
sjarva: Thank you so much! I thought that this is the easiest workaround and hoped that it's enough for the reviewer(s) too 🤗  I am happy to look at those other rules that still use the old util function, because I got stuck with them too 😅 But maybe together we will untangle that mess 💪

Belco90: We can tackle that in a separate PR for sure, thanks!

PR #658, opened 2022-10-01
fix(no-node-access): false positives with `props.children`

### LINKED ISSUE #386: Library incorrectly report no-node-access when referencing props.children
Library complains `Avoid direct Node access. Prefer using the methods from Testing Library.` for the following code `props.children`:

I want to be able to insert children, or a default node.

```
return (
  <>
    { // Other elements here. }
    {'children' in props ? (
      props.children
    ) : (
      <MyComponent {...props} disabled={disabled} />
    )}
  </>
)
```

See [https://stackoverflow.com/q/67099835/481207](https://stackoverflow.com/q/67099835/481207)

sjarva:
## Checks

## Changes
- checks if the node accessing property (e.g. `children`, `firstChild` or `previousSibling`) is called by `props`, so if there's a `props.children`, `props.firstChild` or `props.previousSibling`, then that code won't be reported as an error

## Context
Closes #386 (at least the most common case, not all of the examples given in that issue).
This PR does not close/fix the deconstructed props case (mentioned by @Belco90 in #386):

```
function ComponentB({ children }) {
  // this should NOT be reported
  if (children) {
    // ...
  }

  // this should NOT be reported
  return <div>{children}</div>
}
```

But I am happy to make another PR about this case too (I just haven't figured out how to identify with AST).

P.S. If you feel that this PR is following the rules of [Hactoberfest](https://hacktoberfest.com/) and this gets merged, it would be wonderful if this PR would be tagged with a hacktoberfest-accepted tag/label. Thanks!

### DIFF
--- a/docs/rules/no-node-access.md
+++ b/docs/rules/no-node-access.md
@@ -46,7 +46,7 @@ within(signinModal).getByPlaceholderText('Username');
 ```
 
 ```js
-import { screen } from '${testingFramework}';
+import { screen } from '@testing-library/react';
 
 function ComponentA(props) {
 	// props.children is not reported
--- a/docs/rules/no-node-access.md
+++ b/docs/rules/no-node-access.md
@@ -46,6 +46,17 @@ within(signinModal).getByPlaceholderText('Username');
 ```
 
 ```js
+import { screen } from '${testingFramework}';
+
+function ComponentA(props) {
+	// props.children is not reported
+	return <div>{props.children}</div>;
+}
+
+render(<ComponentA />);
+```
+
+```js
 // If is not importing a testing-library package
 
 document.getElementById('submit-btn').closest('button');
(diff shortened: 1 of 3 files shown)


docs/rules/no-node-access.md
```diff
@@ -45,6 +45,17 @@ const signinModal = getByLabelText('Sign In');
 within(signinModal).getByPlaceholderText('Username');
 ```
 
+```js
+import { screen } from '${testingFramework}';
```
Belco90, line 49: Can you put an actual testing library module in this import?

Belco90 (changes requested):
Just a small change in the docs and will be ready to merge.

docs/rules/no-node-access.md
sjarva: Ah, this is a good catch. This only related to `react`, so I'll put that there.

sjarva: Done!

Belco90 (approved):

Issue #621, opened 2022-08-04
Enable reporting all methods by default in `no-debugging-utils`


Belco90:
Enable reporting all methods by default in `no-debugging-utils`

PR #663, opened 2022-10-02
feat(no-debugging-utils): enable all methods

skovy:
BREAKING CHANGE: enable all debug methods by default for `no-debugging-utils`

## Checks

## Changes

- Enable all of the debug methods by default in the `no-debugging-utils` rule

## Context

Resolves https://github.com/testing-library/eslint-plugin-testing-library/issues/621

### DIFF
--- a/lib/rules/no-debugging-utils.ts
+++ b/lib/rules/no-debugging-utils.ts
@@ -13,13 +13,21 @@ import {
 } from '../node-utils';
 import { DEBUG_UTILS } from '../utils';
 
-type DebugUtilsToCheckFor = Partial<
-	Record<typeof DEBUG_UTILS[number], boolean>
->;
+type DebugUtilsToCheckForConfig = Record<typeof DEBUG_UTILS[number], boolean>;
+type DebugUtilsToCheckFor = Partial<DebugUtilsToCheckForConfig>;
 
 export const RULE_NAME = 'no-debugging-utils';
 export type MessageIds = 'noDebug';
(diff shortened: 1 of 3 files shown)


lib/rules/no-debugging-utils.ts
```diff
@@ -13,14 +13,22 @@ import {
...
-	Record<typeof DEBUG_UTILS[number], boolean>
->;
+type DebugUtilsToCheckForConfig = Record<typeof DEBUG_UTILS[number], boolean>;
+type DebugUtilsToCheckFor = Partial<DebugUtilsToCheckForConfig>;
 
 export const RULE_NAME = 'no-debugging-utils';
 export type MessageIds = 'noDebug';
 type Options = [{ utilsToCheckFor?: DebugUtilsToCheckFor }];
 
+const defaultUtilsToCheckFor: DebugUtilsToCheckForConfig = {
```
skovy, line 23: Intentionally defined a variable with this type so if a debug method is removed/added a type error will be thrown for this default config so it remains in sync

Belco90: Good idea!

Belco90 (changes requested):
Can you remove the options for all the valid/invalid tests manually enabling to report certain debugging methods? They are enabled by default so we don't need to set them explicitly.

skovy:
Ah, yea! I think I removed all of the ones that weren't explicitly part of the test. Let me know if I missed any

Belco90 (approved):

PR #667, opened 2022-10-04
feat(no-await-sync-events): add to DOM, Angular & React configs by default

### LINKED ISSUE #619: Enable no-await-sync-events in proper shareable configs
Enable `no-await-sync-events` in DOM, Angular and React shareable configs

MichaelDeBoey:
BREAKING CHANGE: `no-await-sync-events` is now enabled by default in the DOM, Angular & React configs

---

Closes #619


Belco90:
> should user event be removed from this rule since all methods are async (so there are no sync methods to incorrectly await)? I think the rule as it is may create false positives/negatives in user event v14?

That's right. I think this was mentioned in another issue. Let me create a ticket for it in the v6 project.

Belco90:
Issue for updating `no-await-sync-events` accordingly: #669

timdeschryver:
> should user event be removed from this rule since all methods are async (so there are no sync methods to incorrectly await)? I think the rule as it is may create false positives/negatives in user event v14?

Makes sense to me.
Also, are the sync/async rules needed? There are specific linting rules that verify that promises are awaited and that non-promises are not awaited.

Belco90:
> Also, are the sync/async rules needed? There are specific linting rules that verify that promises are awaited and that non-promises are not awaited.

I guess you mean rules from [TypeScript ESLint](https://typescript-eslint.io/)? If your codebase is written in TS, definitely the sync/async rules can be skipped. But if your codebase is written in JS, then those rules make sense.

Belco90 (approved):

PR #670, opened 2022-10-05
fix(await-async-event): pluralize to await-async-events

skovy:
## Checks

## Changes

- I originally goofed on the naming of this rule when renaming, it should be plural.

## Context

Fully resolves https://github.com/testing-library/eslint-plugin-testing-library/issues/626.


Belco90 (commented):
We need to update the migration doc, right?

Belco90 (approved):
You are right! The migration docs were fine already.

PR #672, opened 2022-10-09
fix: filtered out queries ending with 'ByType'

### LINKED ISSUE #671: Un-report react-test-renderer's findByType & findAllByType queries as async.

await-async-query

Fewer warnings

Check that given node does not end with 'ByType'.

### Example code

	const elements = root.findAllByType('div')

	promise returned from `findAllByType` query must be handled

This rule will not be applied if the query ends with 'ByType'.

### Anything else?

_No response_

Yes

Mrblackey:
from await-async-query rule.

## Checks

- [X] If some rule is added/updated/removed, I've regenerated the rules list (`npm run generate:rules-list`)

## Changes
Added a test case, based on personal usage.
Added a check for early return in the identifier name of the call ends with 'ByType' in await-async-query.ts file.

## Context
Fixes #671

### DIFF
--- a/lib/rules/await-async-query.ts
+++ b/lib/rules/await-async-query.ts
@@ -52,7 +52,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
 			CallExpression(node) {
 				const identifierNode = getDeepestIdentifierNode(node);
 
-				if (!identifierNode || identifierNode.name.endsWith('Type')) {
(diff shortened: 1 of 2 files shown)


Belco90:
Closing because this is related to the Aggressive Reporting feature and the desired behaviour can be achieved through settings.

Issue #674, opened 2022-10-11
Improve `await-async-events` fixer


Belco90:
As discussed in https://github.com/testing-library/eslint-plugin-testing-library/pull/656#issuecomment-1270702921, we need to improve the fixer for `await-async-events` rule. The fixer must add the `async` operator to the outer function if any and not still async.

PR #675, opened 2022-10-12
fix(await-async-events): improve fixer

skovy:
## Changes

- Improve the fixer for `await-async-events` to also add missing `async` keyword to wrapping function expression

## Context

Resolves https://github.com/testing-library/eslint-plugin-testing-library/issues/674

### DIFF
--- a/lib/node-utils/index.ts
+++ b/lib/node-utils/index.ts
@@ -13,6 +13,8 @@ import {
 	isBlockStatement,
 	isCallExpression,
 	isExpressionStatement,
+	isFunctionExpression,
+	isFunctionDeclaration,
 	isImportDeclaration,
 	isImportNamespaceSpecifier,
 	isImportSpecifier,
@@ -93,6 +95,28 @@ export function findClosestVariableDeclaratorNode(
 	}
 
 	return findClosestVariableDeclaratorNode(node.parent);
+}
+
+export function findClosestFunctionExpressionNode(
+	node: TSESTree.Node | undefined
+):
+	| TSESTree.ArrowFunctionExpression
+	| TSESTree.FunctionExpression
+	| TSESTree.FunctionDeclaration
+	| null {
+	if (!node) {
+		return null;
+	}
+
+	if (
+		isArrowFunctionExpression(node) ||
+		isFunctionExpression(node) ||
+		isFunctionDeclaration(node)
+	) {
+		return node;
+	}
+
+	return findClosestFunctionExpressionNode(node.parent);
 }
 
 /**
--- a/lib/rules/await-async-events.ts
+++ b/lib/rules/await-async-events.ts
@@ -3,6 +3,7 @@ import { ASTUtils, TSESLint, TSESTree } from '@typescript-eslint/utils';
 import { createTestingLibraryRule } from '../create-testing-library-rule';
 import {
 	findClosestCallExpressionNode,
+	findClosestFunctionExpressionNode,
 	getFunctionName,
 	getInnermostReturningFunction,
 	getVariableReferences,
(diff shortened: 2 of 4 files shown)


Belco90 (commented):
Whoa, that was quick! Everything looks fine, but I would like to ask for some more tests:

- An invalid test case where the outer test already has the `async` operator, so we test is respected
- An invalid test case where there is no outer function (i.e. just a separate async event method not awaited), so we test the fixer doesn't crash if no outer function is found

lib/node-utils/index.ts
```diff
@@ -95,6 +96,20 @@ export function findClosestVariableDeclaratorNode(
...
 }
 
+export function findClosetFunctionExpression(
+	node: TSESTree.Node | undefined
+): TSESTree.ArrowFunctionExpression | TSESTree.FunctionExpression | null {
+	if (!node) {
+		return null;
+	}
+
+	if (isArrowFunctionExpression(node) || isFunctionExpression(node)) {
```
epmatsw, line 106: Should this also check FunctionDeclarations? It's a little silly, but someone could do something like:

```js
function a () {
  userEvent.click(blah);
}

test("blah", a);
```

lib/rules/await-async-events.ts
```diff
@@ -3,6 +3,7 @@ import { ASTUtils, TSESLint, TSESTree } from '@typescript-eslint/utils';
 import { createTestingLibraryRule } from '../create-testing-library-rule';
 import {
 	findClosestCallExpressionNode,
+	findClosetFunctionExpression,
```
epmatsw, line 6: Typo in the function name: s/Closet/Closest

epmatsw: Also a bit of an inconsistency here, seems like this should be `findClosestFunctionExpressionNode` to match `findClosestCallExpressionNode`?

skovy:
> An invalid test case where the outer test already has the async operator, so we test is respected

This should exist in several of the tests. I noticed the diff is a bit hard to read the way it's summarized so it might need to be expanded.

> An invalid test case where there is no outer function (i.e. just a separate async event method not awaited), so we test the fixer doesn't crash if no outer function is found

Good point. Will do.

Belco90 (approved):
Awesome! Thanks for sorting it out so quick!

lib/rules/await-async-events.ts
Belco90: I think we have covered these cases, right? @epmatsw

epmatsw: Oops, wrong comment. But yes, also addressed

lib/rules/await-async-events.ts
```diff
@@ -142,7 +143,25 @@ export default createTestingLibraryRule<Options, MessageIds>({
 							closestCallExpression,
 							fix: (fixer) => {
 								if (isMemberExpression(node.parent)) {
-									return fixer.insertTextBefore(node.parent, 'await ');
+									const memberExpressionFixer = fixer.insertTextBefore(
+										node.parent,
+										'await '
```
epmatsw, line 148: We're firmly into edge-case territory now, so feel free to disregard.

This can still produces syntactically invalid code in some cases. If there's no FunctionExpressionNode found, that means (I think) that this is inserting a top-level `await`, which is only valid in eslint@8 with `es2022` or higher enabled. It's also not clear to me that adding a top-level await to a single file isn't just going to break everything that depends on it anyways.

Approaches I could think of:

* Don't insert `await` if we can't find & validate the async-ness of the containing function, just leave it as an unfixable error (no top-level await insertion, no chance of doing something unexpected)
* Assume that no one is actually going to have weird dangling top-level `userEvent` calls and skip the complexity of handling that (the current approach)
* Check the eslint context to see if top-level await is supported before inserting in that case (not sure what the exact way to do that would be, probably over-engineering)

Personally I'd prefer the first approach (give up when we hit a seemingly nonsensical case rather than produce potentially invalid code), but since the second is already written and requires no code changes that's fine by me too since I'm not the one writing the code 😅 (Just figured I'd throw out the only remaining edge case I could spot)

skovy:
I updated the fixer logic to only add an `await` before `userEvent` if the containing function expression can be found. The first approach you suggested.

> Don't insert await if we can't find & validate the async-ness of the containing function, just leave it as an unfixable error

This should never happen in practice, and if it does, we'll ignore it instead of being potentially syntactically incorrect.

Belco90 (approved):

epmatsw (approved):

PR #677, opened 2022-10-17
fix(no-wait-for-side-effects): false negatives in variables declarations

### LINKED ISSUE #368: no-wait-for-side-effects: not reporting VariableDeclaration nodes
> Also, I've noticed this rule is NOT reporting this case. Is that intentional? 🤔 
> 
> 
> 
> ```
> 
>     await waitFor(() => {
> 
>       userEvent.click(button) // reported
> 
>       const a = userEvent.click(button) // NOT reported
> 
>     })
> 
> ```

Good catch. It's not intentional, but I detected it when implementing `no-unnecessary-act`, so it will be addressed there. I'll make sure to include this case in the tests.

_Originally posted by @Belco90 in https://github.com/testing-library/eslint-plugin-testing-library/issues/363#issuecomment-830643800_

This will be finally addressed separately, after #259 gets released.

sjarva:
## Checks

## Changes

- adds a new helper function that checks if there are side effects (`userEvent`, `fireEvent` or `render`) in variable declarations
- adds a new invalid unit test

## Context
Closes #368 

P.S. If you feel that this PR is following the rules of [Hactoberfest](https://hacktoberfest.com/) and this gets merged, it would be wonderful if this PR would be tagged with a hacktoberfest-accepted tag/label. Thanks!

### DIFF
--- a/lib/rules/no-wait-for-side-effects.ts
+++ b/lib/rules/no-wait-for-side-effects.ts
@@ -120,6 +120,40 @@ export default createTestingLibraryRule<Options, MessageIds>({
 			return node.expressions.some(isRenderInAssignmentExpression);
 		}
 
+		/**
+		 * Checks if there are side effects in variable declarations.
+		 *
+		 * For example, these variable declarations have side effects:
+		 * const a = userEvent.doubleClick(button);
+		 * const b = fireEvent.click(button);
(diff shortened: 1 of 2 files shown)


Belco90 (approved):
LGTM, thanks!!

PR #684, opened 2022-10-23
test(no-node-access): add more valid test cases

sjarva:
These examples were gotten from discussions in issue #386

## Checks

## Changes

- add a valid test case to `no-node-access`

## Context
Not closing an existing issue, just adding one test that should have been added in #658. I commented in that PR that it wouldn't be fixing an error from a code like this (component with destructured props), but turns out, that this code hasn't produced an error, it is totally valid from this rule's perspective.


Belco90 (approved):
Thanks for this extra valid test 🙏

PR #690, opened 2022-11-19
docs: automate docs with `eslint-doc-generator`

bmish:
## Checks

## Changes

I built this CLI tool [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) for automating the generation of the README rules list and rule doc title/notices for ESLint plugins. It follows common documentation conventions from this and other top ESLint plugins and will help us standardize documentation across ESLint plugins (and generally improve the usability of our custom rules through better documentation and streamline the process of adding new rules).

eslint-doc-generator is significantly more robust compared to the previous documentation generator script which I have now removed. It has much more functionality (for example, the rules list legend is auto-generated, the rules list will show additional columns of information when applicable, etc) and 100% test coverage. It is used in the top 15+ ESLint plugins already.

I also added lint scripts to ensure that no one can forget to run the generate scripts anymore.


MichaelDeBoey (dismissed):

MichaelDeBoey:
ESLint errors will be fixed by #691

package.json
```diff
@@ -30,9 +30,12 @@
 		"format": "prettier --write .",
 		"format:check": "prettier --check .",
 		"generate:configs": "ts-node tools/generate-configs",
-		"generate:rules-list": "ts-node tools/generate-rules-list",
-		"lint": "eslint . --max-warnings 0 --ext .js,.ts",
-		"lint:fix": "npm run lint -- --fix",
+		"generate:rules-list": "npm run build && eslint-doc-generator --path-rule-list \"../README.md\" --path-rule-doc \"../docs/rules/{name}.md\" dist/",
+		"lint": "npm-run-all \"lint:*\"",
```
Belco90, line 34: This will run both `lint:js` and `lint:js:fix`, right? Ideally we only need the former. Can we rework the scripts to avoid running both?

Belco90 (commented):
Many thanks for this PR! I wasn't aware of this tool, it's really useful for us to auto generate those docs based on rules data.

Just leaving a comment about the new lint scripts.

Belco90:
> ESLint errors will be fixed by #691

Why do we have new lint errors out of the blue? Why not fixing them here?

package.json
bmish: `lint:*` will only run `lint:js` but not `lint:js:fix`. That's the convention of npm-run-all:

https://github.com/mysticatea/npm-run-all/blob/master/docs/npm-run-all.md#glob-like-pattern-matching-for-script-names

bmish:
The lint errors aren't related to this PR so I think it's a good idea to fix them in the separate PR.

package.json
Belco90: Ooh I see. Great!

Belco90:
Lint errors fixed in #691, can you pull the main branch in this PR to get the fixes, please?

Belco90 (approved):
LGTM, thanks again!

MichaelDeBoey (approved):

Belco90 (approved):

Belco90:
@all-contributors please add @bmish for doc and tool

MichaelDeBoey:
@bmish Rules now to https://github.com/testing-library/eslint-plugin-testing-library/blob/docs/rules/await-async-query.md instead of https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/await-async-query.md
This is probably a bug in `eslint-doc-generator`? 🤔

bmish:
Away from keyboard now but I'll be sure to investigate and fix this soon hopefully in a few hours.

bmish:
@MichaelDeBoey fixed in https://github.com/testing-library/eslint-plugin-testing-library/pull/693. Please test it, thanks for finding it.

PR #691, opened 2022-11-20
chore: update dependencies

MichaelDeBoey:
Will fix ESLint errors in #690

### DIFF
--- a/lib/node-utils/index.ts
+++ b/lib/node-utils/index.ts
@@ -67,7 +67,7 @@ export function findClosestCallExpressionNode(
 		return node;
 	}
 
-	if (!node || !node.parent) {
+	if (!node?.parent) {
 		return null;
 	}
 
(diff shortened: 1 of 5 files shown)


Belco90:
I don't get why we have new lint errors out of the blue. I suspect it's because unpinned dependencies. This happened many times before already, so I'm in favor of pinning dependencies once and for all.

MichaelDeBoey:
@Belco90 It seems that this is because there was a bug in the `prefer-optional-chain` rule, which is now fixed

bmish (approved):

Belco90:
@MichaelDeBoey so it's definitely caused by not pinning our dependencies. I'll merge this to avoid blocking other PRs and create an issue to tackle the pinning dependencies topic.

Belco90 (approved):

PR #693, opened 2022-11-21
docs: fix rule links in README

bmish:
## Checks

## Changes

Follow-up to:
* #690

Fixes broken rule links as reported [here](https://github.com/testing-library/eslint-plugin-testing-library/pull/690#issuecomment-1321200320).

Normally, [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) would never have any trouble with this. The reason for the trouble this time is that eslint-plugin-testing-library has some unusual setup:

* The package root as specified by package.json (`"main": "index.js"`) doesn't actually exist in that location, it exists in `dist/` because of `"pkgRoot": "dist"` specified by semantic release.
  https://github.com/testing-library/eslint-plugin-testing-library/blob/3b278d02eb30d7743511a43a44373f6adc44f071/package.json#L27
  https://github.com/testing-library/eslint-plugin-testing-library/blob/3b278d02eb30d7743511a43a44373f6adc44f071/.releaserc.json#L2
* Because of that, we have to point eslint-doc-generator to `dist/` as the plugin root.
* But then we have to correct the file paths with `--path-rule-list "../README.md"` and `--path-rule-doc "../docs/rules/{name}.md"` since these still live outside `dist/`.
* And finally, we have to update the URL to the rule docs, since these aren't affected by the weirdness on disk, with `--url-rule-doc "docs/rules/{name}.md"`.


Belco90 (approved):

Belco90:
@bmish Thanks for the extra context about the package root. It would be a good idea to make the dist location more explicit.

PR #694, opened 2022-11-21
refactor: rework rule docs scripts

Belco90:
## Checks

## Changes

- Separate scripts for doc generation from linting
- Run doc checks in CI separately
- Remove unnecessary checks from PR template

## Context

Follow-up to #690


.github/pull_request_template.md
```diff
@@ -1,8 +1,6 @@
 ## Checks
 
 - [ ] I have read the [contributing guidelines](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/CONTRIBUTING.md).
-- [ ] If some rule is added/updated/removed, I've regenerated the rules list (`npm run generate:rules-list`)
-- [ ] If some rule meta info is changed, I've regenerated the plugin shared configs (`npm run generate:configs`)
```
Belco90, line 5: These aren't needed anymore since this is checked on CI.

.github/workflows/pipeline.yml
```diff
@@ -45,6 +45,9 @@ jobs:
       - name: Check format
         run: npm run format:check
 
+      - name: Check autogenerated docs
+        run: npm run docs:gen && git diff --exit-code
```
Belco90, line 49: Docs checks are run in CI. Should we split each code check on its own job, so they are run in parallel?

Belco90, line 49: This command has been removed from `scripts` under package.json since it doesn't really make sense to be run locally, as any other change pending to be committed can make this check fail.

tests/index.test.ts
```diff
@@ -1,13 +1,8 @@
-import { exec } from 'child_process';
-import { existsSync } from 'fs';
-import { resolve } from 'path';
-import util from 'util';
+import { existsSync } from 'node:fs';
+import { resolve } from 'node:path';
```
MichaelDeBoey, line 2: This is only available as of Node v18

```suggestion
import { existsSync } from 'fs';
import { resolve } from 'path';
```

package.json
```diff
@@ -28,15 +28,15 @@
...
-		"lint:js:fix": "npm run lint:js -- --fix",
+		"prettier-base": "prettier . --ignore-unknown --cache --loglevel warn",
+		"format": "npm run prettier-base -- --write",
+		"format:check": "npm run prettier-base -- --check",
+		"lint": "eslint . --max-warnings 0 --ext .js,.ts",
+		"lint:fix": "npm run lint -- --fix",
+		"rule-doc-generator": "eslint-doc-generator --path-rule-list \"../README.md\" --path-rule-doc \"../docs/rules/{name}.md\" --url-rule-doc \"docs/rules/{name}.md\" dist/",
+		"docs:gen": "npm-run-all \"docs:gen:*\"",
+		"docs:gen:configs": "ts-node tools/generate-configs && npm run format",
+		"docs:gen:rules-list": "npm run build && npm run rule-doc-generator && npm run format",
```
MichaelDeBoey, line 39: Can we run `npx sort-package-json` again, like it was?

MichaelDeBoey (changes requested):

tests/index.test.ts
Belco90: I think it's available since v16, but still out of the range we support 🤦 . I'm reverting it!

package.json
Belco90: I didn't know it was sorted following that package. Sure, I'll run it.

Belco90: I'm running it, but it makes no difference 🤔

.github/pull_request_template.md
sjarva: Wonderful! 🤩

package.json
MichaelDeBoey: This is probably because of the `npm-run-all` call
By default it's running these in sequences, it should use the parallel flag instead.

```json
"docs:gen": "run-p \"docs:gen:*\"",
```

Same for `generate` & `lint` script

Belco90: I'm completely lost now. How is this related to `npx sort-package-json`?

MichaelDeBoey: If `npm-run-all` is running in sequence, the order of the scripts matter
`sort-package-json` is taking this into account.

For more info see https://github.com/keithamus/sort-package-json/issues/220 & https://github.com/keithamus/sort-package-json/pull/232

Temporary solution would be to remove `npm-run-all` from `devDependencies`, run `npx sort-package-json` & add `npm-run-all` to `devDependencies` again

Belco90: Not sure if I understood this correctly at all.

I was just running `sort-package-json` manually to sort the package.json according to it. But it seems you are referring to running this in one script as before? I don't see where it was executed prior to my changes.

MichaelDeBoey: You indeed need to run it manually, but first need to (temporarily) remove the `npm-run-all` dependency, otherwise it won't change anything

Belco90: Ok, I don't understand any of this, but I finally got it 😂

MichaelDeBoey (approved):

PR #703, opened 2022-12-05
fix(await-async-query): work around false-positives in chained findBy queries

### LINKED ISSUE #673: False positives for `find*` queries from `react-test-renderer`

Yes

### Plugin version

v5.7.2

### ESLint version

v8.25.0

### Node.js version

16.15.0

n/a

### Operating system

n/a

### Bug description

[The `react-test-renderer` library provides several utils](https://reactjs.org/docs/test-renderer.html#testinstance) which match the `find*` query pattern: `findByType`, `findByProps`, `findAll`, `findAllByType`,  `findAllByProps`.

These methods are provided from a `react-test-renderer` instance, or any [React Native Testing Library query](https://callstack.github.io/react-native-testing-library/docs/api-queries#queries). They should be ignored, so they are not treated as Testing Library queries (since they actually aren't).

This should apply to all rules, so it must be implemented in the [`isQuery` internal helper](https://github.com/testing-library/eslint-plugin-testing-library/blob/9eaca4a427fe80ae7e0901069a1c0afb8a89afdf/lib/create-testing-library-rule/detect-testing-library-utils.ts#L356). This function should ignore the queries listed before if they were imported from `react-test-renderer` or `@testing-library/react-native`. Checking the imports from the latter it's important since `react-native-testing-library` queries return nodes with the mentioned utils, so it's possible to do something like `getByRole('modal').findByProps('bar')`, hence ignoring `react-test-renderer` utils if coming from `@testing-library/react-native`

### Steps to reproduce

This example extracted from [React docs](https://reactjs.org/docs/test-renderer.html) to cover the `react-test-renderer` import:

```js
import TestRenderer from 'react-test-renderer';

function MyComponent() {
  return (
    <div>
      <SubComponent foo="bar" />
      <p className="my">Hello</p>
    </div>
  )
}

function SubComponent() {
  return (
    <p className="sub">Sub</p>
  );
}

const testRenderer = TestRenderer.create(<MyComponent />);
const testInstance = testRenderer.root;

expect(testInstance.findByType(SubComponent).props.foo).toBe('bar');
expect(testInstance.findByProps({className: "sub"}).children).toEqual(['Sub']);
```

Example to cover the `@testing-library/react-native` import:

```js
import { render } from '@testing-library/react-native';

test('some test case', () => {
  render(<SomeComponent />);
  const falsePositive = screen.getByText('foo').findByProps({ testID: 'bar' })
})
```

It would be interesting to check the test cases added in [this PR](https://github.com/testing-library/eslint-plugin-testing-library/pull/703/files#diff-2889ef91637a07f60b716ad39b8e9b3c61da2b4433e62b899ea27797116a4463).

### Error output/screenshots

`findByType`, `findByProps`, and other `react-test-renderer` utils are reported by the plugin (e.g., `await-async-queries` complains about they must be awaited).

### ESLint configuration

n/a

### Rule(s) affected

All rules reporting on queries are affected, but `await-async-queries` is the most obvious one.

### Anything else?

This was originated by #671 and #703

No

notjosh:
## Checks

## Changes

- improve `await-async-query` to be aware of (likely) `react-test-renderer` instances so it can suppress false-positive errors requiring `await` on `findBy*()` queries when used with React Native Testing Library

🚧 I'm happy to clean this up before merging, but wanted to get feedback before proceeding.

The approach used is fairly naïve but it seems reliable in my usage. Similar to other rules in this repo, it's easy enough to create false-negatives with enough layers of abstraction (via variables/function calls/etc), but the more common simple cases should be covered.

## Context

Related: #673

### DIFF
--- a/lib/rules/await-async-query.ts
+++ b/lib/rules/await-async-query.ts
@@ -8,6 +8,8 @@ import {
 	getInnermostReturningFunction,
 	getVariableReferences,
 	isPromiseHandled,
+	isMemberExpression,
+	isCallExpression,
 } from '../node-utils';
 
 export const RULE_NAME = 'await-async-query';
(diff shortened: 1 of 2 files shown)


Belco90 (commented):
Hi @notjosh! Thanks for your MR.

I'm afraid this wasn't the approach we needed. As described in #673, the actual fix should be ignoring every query matching Testing Library patterns if they were imported from `react-test-renderer`.

Your fix is interesting, but it's trying to fix false positives in chained `findBy*` queries. However, the actual problem is caused by false positives in `findBy*` queries from `react-test-renderer`. Another issue of solving this exclusively for this rule is that the same false positive might happen in another rule, that's another reason why we require a generic fix in our internal helpers (as mentioned in the original ticket).

I'll think about if we can reuse part of your code for the actual fix or something else, since I don't want to just throw away your code.

notjosh:
> I'm afraid this wasn't the approach we needed. As described in https://github.com/testing-library/eslint-plugin-testing-library/issues/673, the actual fix should be ignoring every query matching Testing Library patterns if they were imported from react-test-renderer.

@Belco90 I was just thinking about this more, and realised this PR doesn't actually solve #673. But it does solve a different problem.

#673 is about directly importing (and using) `react-test-renderer` directly, but this PR isn't relevant to that.

Instead, this PR fixes an issue we're having that comes from [React Native Testing Library](https://callstack.github.io/react-native-testing-library/). For example, this small snippet summarises the problem:

```typescript
import { render } from '@testing-library/react-native';

test('some test case', async () => {
  const screen = render(<SomeComponent />);
  const falsePositive = screen.getByText('foo').findByProps({ testID: 'bar' })
                               ^^^^^^^^^        ^- from react-test-renderer
                               └- from RNTL, returns ReactTestInstance
})
```

In RNTL, the queries (mostly) all return a `ReactTestInstance` ([docs](https://callstack.github.io/react-native-testing-library/docs/api-queries#queries)) from `react-test-renderer`, which can happily chain a (sync) `findBy*()` method (causing the 💥 eslint false positive 💥).

However, as far as I can tell, DOM-based Testing Library queries (usually) return the _element_, so chaining doesn't make sense, and you'd never run into this problem.

So I guess we can think about this PR as being specific to React Native Testing Library, so we can:

- create a `testing-library/react-native` configuration, which extends `testing-library/react` and includes this rule
- ignore RNTL as it's a different library with a different API (and just happens to be mostly compatible)
- include this rule in `testing-library/react` as it is, as it shouldn't interfere with non-React Native usage

Can you take another look at this PR with this context in mind? Fixing this issue is a blocker to us adopting this library, so I'd love to help you resolve it!

Belco90:
@notjosh this changes everything, thank you very much for the context!

I wasn't aware that React Native Testing Library was returning a `[ReactTestInstance](https://reactjs.org/docs/test-renderer.html#testinstance)` from every query.

That basically means the problem you described and the one from #671 are related, and all the `findBy` utils provided by `react-test-renderer` must be ignored as a Testing Library query in general.

The bad news is that your PR won't be necessary. The good news is that the actual fix is pretty simple: we only need to ignore all the [`find*` utils provided by `react-test-renderer`](https://reactjs.org/docs/test-renderer.html#testinstance). That's as simple as excluding those utils name from our [`isFindQueryVariant` internal method](https://github.com/testing-library/eslint-plugin-testing-library/blob/25f73c546c586adf27fbc50b169d5d42afcc2f95/lib/create-testing-library-rule/detect-testing-library-utils.ts#L391).

I'm closing this PR and repurposing #673 to apply this simple change instead. Would you like to submit the PR to change that behavior @notjosh?

PR #705, opened 2022-12-20
chore: refine typescript-eslint config

Belco90:
## Checks

## Changes

- Remove unnecessary config from eslintrc
- Move ESLint config file to JS format
- Enable [linting with type information](https://typescript-eslint.io/linting/typed-linting)
- Fix new linting errors

## Context

n/a


.eslintrc.js
```diff
@@ -0,0 +1,55 @@
...
+		'plugin:jest-formatting/recommended',
+		'prettier',
+	],
+	rules: {
+		// Base
+		'max-lines-per-function': 'off',
+		'no-restricted-imports': [
+			'error',
+			{
+				patterns: ['@typescript-eslint/utils/dist/*'],
```
sjarva, line 15: I'm asking because I'm curious - why aren't we allowing importing from `@typescript-eslint/utils/dist/*`? Could we clarify that with the `message` option for this rule and in that message also guide people to importing from somewhere else (if such library or folder exists)?

tests/index.test.ts
```diff
@@ -67,6 +67,7 @@ it('should export configs that refer to actual rules', () => {
 		expect(rule.startsWith(ruleNamePrefix)).toBe(true);
 		expect(ruleNames).toContain(ruleName);
 
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
sjarva, line 70: I tried to change this code so that we wouldn't need the eslint-disable comment at all, but oh welp, that didn't work 😅 So let's let this be.

.eslintrc.js
Belco90: Good point! Not sure if still possible, but you could import typescript-eslint utils from the `dist/` folder instead of their exported objects. You can see [the original commit here](https://github.com/testing-library/eslint-plugin-testing-library/pull/385/commits/1b2024c8f892cc7b0a5a2b462f0946213edae867).

I'll add a message to clarify it must be just exported from `@typescript-eslint/utils` instead.

Belco90: Fixed! 
imagen

tests/index.test.ts
Belco90: This is a test, so I'd say it's fine having an exception here.

sjarva (approved):
Thanks for doing this!

Also, a very meta PR - configuring the `eslint` config of a `eslint` plugin library 😂

PR #709, opened 2022-12-24
fix: avoid reporting `react-test-renderer` utils as queries

### LINKED ISSUE #671: Un-report react-test-renderer's findByType & findAllByType queries as async.

### LINKED ISSUE #673: False positives for `find*` queries from `react-test-renderer`

Belco90:
## Checks

## Changes

- Avoid reporting `react-test-renderer` utils as `find*` queries
- Avoid reporting `react-test-renderer` utils as custom queries
- Add tests

## Context

Fixes #671
Fixes #673

### DIFF
--- a/lib/create-testing-library-rule/detect-testing-library-utils.ts
+++ b/lib/create-testing-library-rule/detect-testing-library-utils.ts
@@ -419,7 +419,11 @@ export function detectTestingLibraryUtils<
 		};
 
 		const isCustomQuery: IsCustomQueryFn = (node) => {
-			return isQuery(node) && !ALL_QUERIES_COMBINATIONS.includes(node.name);
+			return (
+				isQuery(node) &&
+				!ALL_QUERIES_COMBINATIONS.includes(node.name) &&
+				!EXCLUDED_FIND_BY_QUERIES.includes(node.name)
+			);
 		};
 
 		const isBuiltInQuery = (node: TSESTree.Identifier): boolean => {
--- a/lib/create-testing-library-rule/detect-testing-library-utils.ts
+++ b/lib/create-testing-library-rule/detect-testing-library-utils.ts
@@ -133,6 +133,14 @@ const FIRE_EVENT_NAME = 'fireEvent';
 const CREATE_EVENT_NAME = 'createEvent';
 const USER_EVENT_NAME = 'userEvent';
 const RENDER_NAME = 'render';
+const EXCLUDED_FIND_BY_QUERIES = [
+	//react-test-renderer utils
+	'findByType',
+	'findByProps',
+	'findAll',
+	'findAllByType',
+	'findAllByProps',
+];
 
 export type DetectionOptions = {
 	/**
@@ -389,7 +397,11 @@ export function detectTestingLibraryUtils<
 		 * Determines whether a given node is `find*` query variant or not.
 		 */
 		const isFindQueryVariant: IsFindQueryVariantFn = (node) => {
-			return isQuery(node) && node.name.startsWith('find');
+			return (
+				isQuery(node) &&
+				node.name.startsWith('find') &&
+				!EXCLUDED_FIND_BY_QUERIES.includes(node.name)
+			);
 		};
 
 		/**
--- a/tests/create-testing-library-rule.test.ts
+++ b/tests/create-testing-library-rule.test.ts
@@ -401,6 +401,16 @@ ruleTester.run(RULE_NAME, rule, {
       `,
 		},
 
+		// `react-test-renderer` utils starting with "find" shouldn't be reported as queries
+		// https://github.com/testing-library/eslint-plugin-testing-library/issues/673
+		`// issue #673
+		findByType('foo');
+		findByProps('foo');
+		findAll('foo');
+		findAllByType('foo');
+		findAllByProps('foo');
+		`,
+
 		// Weird edge cases
 		`(window as any).__THING = false;`,
 		`thing.method.lastCall.args[0]();`,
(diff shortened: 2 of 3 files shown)


lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -389,7 +397,11 @@ export function detectTestingLibraryUtils<
 		 * Determines whether a given node is `find*` query variant or not.
 		 */
 		const isFindQueryVariant: IsFindQueryVariantFn = (node) => {
-			return isQuery(node) && node.name.startsWith('find');
+			return (
+				isQuery(node) &&
+				node.name.startsWith('find') &&
+				!EXCLUDED_FIND_BY_QUERIES.includes(node.name)
```
Belco90, line 403: Actually... I could add this to `isQuery` instead.

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -133,6 +133,14 @@ const FIRE_EVENT_NAME = 'fireEvent';
 const CREATE_EVENT_NAME = 'createEvent';
 const USER_EVENT_NAME = 'userEvent';
 const RENDER_NAME = 'render';
+const EXCLUDED_FIND_BY_QUERIES = [
+	//react-test-renderer utils
+	'findByType',
+	'findByProps',
+	'findAll',
+	'findAllByType',
+	'findAllByProps',
```
Mrblackey, line 142: And what if someone uses a custom query with one of these names? We still want them to be reported. no? How can we make sure that we are dealing with a react-test-renderer query?
This relates also to [this comment](https://github.com/testing-library/eslint-plugin-testing-library/issues/673#issuecomment-1281303226).

Belco90: That would be the ideal fix indeed. However, since the utils names seem quite specific and can be chained from built-in queries in React Native, it could be a quick workaround.

I personally don't have the time to implement the ideal fix so we check if queries are coming from `react-test-renderer`, so it's a matter of deciding if we are happy with this workaround (and its trade-offs) or we are only interested in the ideal fix.

Belco90: Another idea could be adding a config for React Native to the plugin, which adds automatically the corresponding query names to the settings to be ignored as custom queries. This would be a workaround too, but perhaps makes more sense.

Belco90:
Thinking about this again... this doesn't make sense. I'll close this PR and update the issue to mention how the ideal fix should look like and a temporary workaround for those React Native users running into false positives.

PR #716, opened 2023-01-13
fix: ignore getBy inside of within for prefer-presence-queries on absence queries

### LINKED ISSUE #518: False positives with testing-library/prefer-presence-queries

nathanmmiller:
## Checks

## Changes

- Ignore getBy* nodes inside of within() for prefer-presence-queries when checking absence queries

## Context

Fixes #518 [Well, partially fixes it.]

### DIFF
--- a/tests/lib/rules/prefer-presence-queries.test.ts
+++ b/tests/lib/rules/prefer-presence-queries.test.ts
@@ -837,6 +837,9 @@ ruleTester.run(RULE_NAME, rule, {
      // right after clicking submit button it disappears
      expect(submitButton).not.toBeInTheDocument()
     `,
+		`// checking absence on getBy* inside a within with queryBy* outside the within
+	 expect(within(screen.getByRole("button")).queryByText("Hello")).not.toBeInTheDocument()
+	`,
 	],
 	invalid: [
 		// cases: asserting absence incorrectly with `getBy*` queries
--- a/lib/rules/prefer-presence-queries.ts
+++ b/lib/rules/prefer-presence-queries.ts
@@ -59,6 +59,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
 		return {
 			'CallExpression Identifier'(node: TSESTree.Identifier) {
 				const expectCallNode = findClosestCallNode(node, 'expect');
+				const withinCallNode = findClosestCallNode(node, 'within');
+
+				if (withinCallNode) {
+					//Ignore false-positives where our query is inside a within
+					return;
+				}
 
 				if (!expectCallNode || !isMemberExpression(expectCallNode.parent)) {
 					return;


tests/lib/rules/prefer-presence-queries.test.ts
```diff
@@ -837,6 +837,9 @@ ruleTester.run(RULE_NAME, rule, {
      // right after clicking submit button it disappears
      expect(submitButton).not.toBeInTheDocument()
     `,
+		`// checking absence on getBy* inside a within with queryBy* outside the within
+	 expect(within(screen.getByRole("button")).queryByText("Hello")).not.toBeInTheDocument()
```
Belco90, line 841: Could you add another test to check that chained queries from `within` are reported as expected tho?

```js
expect(within(screen.getByRole("button")).getByText("Hello")).not.toBeInTheDocument()
```

The example above should be reported, but I don't think it will be by your implementation.

Belco90 (changes requested):
Hi @nathanmmiller! Thanks for your PR. I left a comment asking for more tests, since I think your implementation is not covering all cases as expected.

tests/lib/rules/prefer-presence-queries.test.ts
nathanmmiller: Good call. I think maybe I need to rethink the way I am handling this - it's not sufficient to say "ignore everything inside a within." Lemme rethink and re-PR.

nathanmmiller: Actually I think this does correctly handle this use case. I've added the unit test as requested and will re-raise the PR - it correctly deals with this exact case.

nathanmmiller: @Belco90 reopened at https://github.com/testing-library/eslint-plugin-testing-library/pull/717, with the additional test case

PR #729, opened 2023-02-02
chore: pin dependencies with lockfile

Belco90:
## Checks

## Changes

- Pin dependencies with `package-lock.json`

## Context

I'm tired of getting unexpected errors in CI because our dev dependencies aren't fixed by a lock file. I'm fixing them with the npm lock file.

In a follow-up PR, I'll set up Renovatebot to keep our dev dependencies up to date automatically for us (disabling dependabot too).


.github/workflows/pipeline.yml
```diff
@@ -3,13 +3,14 @@ name: Pipeline
...
     branches:
-      # semantic-release valid branches
+      # semantic-release valid branches and all-contributors branches
       - '+([0-9])?(.{+([0-9]),x}).x'
       - 'main'
       - 'next'
       - 'next-major'
       - 'beta'
       - 'alpha'
+      - 'all-contributors/**'
```
Belco90, line 13: I left this one out in a previos PR. Fixing it here.

package.json
```diff
@@ -88,6 +88,6 @@
 	},
 	"engines": {
 		"node": "^12.22.0 || ^14.17.0 || >=16.0.0",
-		"npm": ">=6"
+		"npm": ">=8"
```
MichaelDeBoey, line 91: This could throw errors in projects that use `eslint-plugin-testing-library` (especially when they use Yarn)

```suggestion
		"npm": ">=6"
```

MichaelDeBoey (changes requested):
Why would you change to Renovatebot?
Dependabot would create PRs for updating as well once we have the lock file

package.json
Belco90: Good point!

Belco90:
@MichaelDeBoey Because Renovatebot is way smarter than Dependabot, has more granular control, and provides [lockfile maintenance](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance).

MichaelDeBoey:
@Belco90 Dependabot will update lock files as well

Belco90:
@MichaelDeBoey Didn't mean just updating a dependency in the lockfile, but updating the lockfile regularly so transitive dependencies are deduped and up to date.

MichaelDeBoey:
@Belco90 Dependabot does that as well

MichaelDeBoey (approved):

Belco90:
> @Belco90 Dependabot does that as well

That's great! I'll check id Dependabot can do what we need then.

PR #730, opened 2023-02-04
feat: add toBeVisible as a presence query

### LINKED ISSUE #728: Create a rule to support enforcing getBy* for toBeVisible
(this issue has been edited to reflect an updated intent; it was originally a proposal to change `prefer-presence-queries`)

### Name for new rule

prefer-query-matchers

The rule should somehow provide a way to enforce that `toBeVisible()` should always be called with a `getBy*` query in place of a `queryBy*`. This works differently than the `prefer-presence-queries` rule, because a negative there expects an element to not be present in the DOM. In my understanding `not.toBeVisible()` means "is present in the DOM but is not visible," and it fails if it receives a null: "received value must be an HTMLElement or an SVGElement".

Rather than coding for this specific matcher only, we currently plan to provide a configurable rule where you can configure which matchers require `get` and which require `query`.

### Testing Library feature

The `getBy*` and `queryBy*` queries. It is prompted by the `toBeVisible` matcher from `jest-dom`.

### Testing Library frameworks

`getBy*` and `queryBy*` are in the Core API and so are not specific to any framework.

Other: a best-practice recommendation. It allows users to configure rules to enforce use of the queries that will produce the best failure messages.

### Code Examples

For the following proposed options schema:

```
{
  "rules": {
    "prefer-query-matchers": ["error", { 
      "validEntries": [{ "query": "get", "matcher": "toBeVisible"}]
     }],
  }
}
```

With the following code:

```
/* 1 */ expect(screen.queryByText("a")).toBeVisible();
/* 2 */ expect(screen.getByText("a")).toBeVisible();
/* 3 */ expect(screen.queryByText("a")).not.toBeVisible();
/* 4 */ expect(screen.getByText("a")).not.toBeVisible();
```

1 and 3 would produce errors, and 2 and 4 would not.

If instead the `query` value was changed:

```
{ 
  "validEntries": [{ "query": "get", "matcher": "toBeVisible"}]
}
```

Then 2 and 4 would produce errors, and 1 and 3 would not.

### Anything else?

We investigated handling this with the existing `prefer-presence-queries` rule, but since that rule has a presence/absence distinction and the motivating case `toBeVisible` does not, it is a better fit to handle it in its own standalone rule. That also allows users to configure handling for any number of matchers.

Yes

CodingItWrong:
## Checks

## Changes

- Adds `toBeVisible` as a presence query, so that `expect(screen.queryBy…).toBeVisible()` will be recommended to be changed to `expect(screen.getBy…).toBeVisible()`

## Context

Fixes #728

## Questions

I am not fully confident if this PR yet handles the proposed behavior in #728 — both of the following should produce an error:

```
/* 5 */ expect(screen.queryByText("a")).toBeVisible();
/* 7 */ expect(screen.queryByText("a")).not.toBeVisible();
```

This is different from other presence queries, where `expect(screen.queryBy…).not.toBeVisible()` is correct.

Could a review please check if that is or isn't working as desired?

Since I just copied the `toBeInTheDocument` tests, I would think that the tests are _not_ specifying that this is the case. But I don't have enough confidence in my understanding of the test suite to be sure. Also, I wasn't able to figure out how to run my local branch in a real React project; I got the error "ESLint couldn't find the plugin" when I tried `yarn publish`, `yalc`, and just specifying a `package.json` path to my local plugin folder.

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -116,7 +116,12 @@ const ALL_RETURNING_NODES = [
 	...METHODS_RETURNING_NODES,
 ];
 
-const PRESENCE_MATCHERS = ['toBeInTheDocument', 'toBeTruthy', 'toBeDefined'];
+const PRESENCE_MATCHERS = [
+	'toBeInTheDocument',
+	'toBeVisible',
+	'toBeTruthy',
+	'toBeDefined',
+];
 const ABSENCE_MATCHERS = ['toBeNull', 'toBeFalsy'];
 
 export {
--- a/tests/lib/rules/prefer-presence-queries.test.ts
+++ b/tests/lib/rules/prefer-presence-queries.test.ts
@@ -140,6 +140,11 @@ ruleTester.run(RULE_NAME, rule, {
 				}),
 				...getValidAssertions({
 					query: queryName,
+					matcher: '.toBeVisible()',
+					assertionType: 'presence',
+				}),
+				...getValidAssertions({
+					query: queryName,
 					matcher: '.toBeTruthy()',
 					assertionType: 'presence',
 				}),
@@ -188,6 +193,12 @@ ruleTester.run(RULE_NAME, rule, {
 				...getValidAssertions({
 					query: queryName,
 					matcher: '.toBeInTheDocument()',
+					shouldUseScreen: true,
+					assertionType: 'presence',
+				}),
+				...getValidAssertions({
+					query: queryName,
+					matcher: '.toBeVisible()',
 					shouldUseScreen: true,
 					assertionType: 'presence',
 				}),
@@ -253,6 +264,11 @@ ruleTester.run(RULE_NAME, rule, {
 				}),
 				...getValidAssertions({
 					query: queryName,
+					matcher: '.toBeVisible()',
+					assertionType: 'presence',
+				}),
+				...getValidAssertions({
+					query: queryName,
 					matcher: '.toBeTruthy()',
 					assertionType: 'presence',
 				}),
@@ -301,6 +317,12 @@ ruleTester.run(RULE_NAME, rule, {
 				...getValidAssertions({
 					query: queryName,
 					matcher: '.toBeInTheDocument()',
+					shouldUseScreen: true,
+					assertionType: 'presence',
+				}),
+				...getValidAssertions({
+					query: queryName,
+					matcher: '.toBeVisible()',
 					shouldUseScreen: true,
 					assertionType: 'presence',
 				}),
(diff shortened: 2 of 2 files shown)


Belco90:
@CodingItWrong I didn't realize about these two cases were supposed to throw an error:

```js
/* 5 */ expect(screen.queryByText("a")).toBeVisible();
/* 7 */ expect(screen.queryByText("a")).not.toBeVisible();
```

> This is different from other presence queries, where expect(screen.queryBy…).not.toBeVisible() is correct.

This is basically the main reason, let me give you some context.

From the point of view of the rule, reporting case 5 but not reporting case 7 is expected. The goal of the rule is to encourage specific matchers for checking presence or absence. If a matcher is related to presence, then using `get*` queries with a positive matcher is correct (e.g. `expect(screen.getByText("a")).toBeInTheDocument()`). Otherwise, using `query*` queries or negated matcher will produce an error (e.g. `expect(screen.queryByText("a")).toBeInTheDocument()` and `expect(screen.getByText("a")).not.toBeInTheDocument()`). This basically means the matchers are always associated with either presence or absence, and the other usage will be considered as invalid.

On the other hand, from the point of view of `toBeVisible` matcher, case 5 and 7 are expected to be reported because we need an existing element to assert if it's visible or not. Then, this matcher can't be aligned with the concept of presence or absence from this rule, since the matcher must find an existing element before doing the assertion. What we want to do is encourage always using `get*` queries with `toBeVisible` matcher, which is different from the behavior described for this rule.

I'd say it doesn't make sense to include `toBeVisible` in `prefer-presence-queries` since the behavior doesn't match, so I'd close this PR. That doesn't mean there is no room for the desired behavior for `toBeVisible` matcher, but it should be placed in a new rule. I'd say we could create a new rule named like `prefer-query-matchers`, that receives through its options those valid entries of queries and matchers that must be used together (it could be also the other way around, so we define invalid query and matcher combinations, this is just an idea):

```json
{
  "rules": {
    "prefer-query-matchers": ["error", { 
      "validEntries": [{ "query": "get", "matcher": "toBeVisible"}]
     }],
  }
}
```

Then, this new rule would report the matchers given if found used with other queries than the valid ones defined.

What do you think?

CodingItWrong:
@Belco90 Your assessment sounds totally right; I think it makes sense for this behavior to be implemented as a separate rule. Especially because there is _so much_ already going on in the existing `prefer-presence-queries` rule. Plus, the new rule will be simpler, since we don't need to check presence vs absence.

I'm closing this PR and I've updated #728 to note the plan to create a new rule. I will shoot to work on that soon.

PR #803, opened 2023-08-22
fix(no-await-sync-events): stop reporting `user-event` by default

### LINKED ISSUE #669: Add option to stop reporting user-event in `no-await-sync-events`
`no-await-sync-events` must not report user-event methods by default since it aims user-event v14, so only fire-event it's reported. However, we need an option to decide which methods to be reported: fire-event, user-event or both.

The idea is to include an option in the same way as #626 with the following values for presets:

- dom: rule enabled with fire-event
- angular: rule enabled with fire-event
- react: rule enabled with fire-event
- vue: rule not enabled
- marko: rule not enabled

### LINKED ISSUE #801: Contradicting issue reports between `async-await-events` and `no-await-sync-events`

Yes

### Plugin version

latest

### ESLint version

latest

### Node.js version

18.16.1

pnpm 8.6.12

### Operating system

Macos 13.5

### Bug description

I upgraded to v6 and was first met with this error in my Vercel build:

image

Easy enough. These had no `await` so I added that before each. Thinking I would be in the clear, I wasn't expecting the build to fail. However, I am now getting the opposite error in Vercel:

image

Unsure of what to do here or if this a bug or I need to do some combination of rules config in my eslint. Thanks for the help in advance!

### Steps to reproduce

See above

### Error output/screenshots

_No response_

### ESLint configuration

No rules overridden in ESLint

### Rule(s) affected

`async-await-events` and `no-await-sync-events`

### Anything else?

_No response_

No

Belco90:
## Checks

## Changes

- Stop reporting `user-event` methods by default, assuming v14 is the standard now.

## Context

Fixes #801
Fixes #669

### DIFF
--- a/lib/rules/no-await-sync-events.ts
+++ b/lib/rules/no-await-sync-events.ts
@@ -10,13 +10,28 @@ import {
 } from '../node-utils';
 
 const USER_EVENT_ASYNC_EXCEPTIONS: string[] = ['type', 'keyboard'];
-const VALID_EVENT_MODULES = ['fire-event', 'user-event'] as const;
+const FIRE_EVENT_OPTION = 'fire-event' as const;
+const USER_EVENT_OPTION = 'user-event' as const;
+const VALID_EVENT_MODULES = [FIRE_EVENT_OPTION, USER_EVENT_OPTION];
 
 export const RULE_NAME = 'no-await-sync-events';
 export type MessageIds = 'noAwaitSyncEvents';
-type Options = [
-	{ eventModules?: readonly (typeof VALID_EVENT_MODULES)[number][] }
-];
+
+type ValidEventModules = (typeof VALID_EVENT_MODULES)[number];
+type EventModulesOptions = ReadonlyArray<ValidEventModules> | ValidEventModules;
+type Options = [{ eventModules?: EventModulesOptions }];
+
+function getEnabledEventModules(
+	eventModulesOption?: EventModulesOptions | undefined
+): ReadonlyArray<ValidEventModules> {
+	if (typeof eventModulesOption === 'undefined') {
+		// This should match the default option
+		return [FIRE_EVENT_OPTION];
+	} else if (typeof eventModulesOption === 'string') {
+		return [eventModulesOption];
+	}
+	return eventModulesOption;
+}
 
 export default createTestingLibraryRule<Options, MessageIds>({
 	name: RULE_NAME,
@@ -25,9 +40,9 @@ export default createTestingLibraryRule<Options, MessageIds>({
 		docs: {
 			description: 'Disallow unnecessary `await` for sync events',
 			recommendedConfig: {
-				dom: 'error',
-				angular: 'error',
-				react: 'error',
+				dom: ['error', { eventModules: 'fire-event' }],
+				angular: ['error', { eventModules: 'fire-event' }],
+				react: ['error', { eventModules: 'fire-event' }],
 				vue: false,
 				marko: false,
 			},
@@ -41,18 +56,22 @@ export default createTestingLibraryRule<Options, MessageIds>({
 				type: 'object',
 				properties: {
 					eventModules: {
-						type: 'array',
-						minItems: 1,
-						items: {
-							enum: VALID_EVENT_MODULES,
-						},
+						default: FIRE_EVENT_OPTION,
+						oneOf: [
+							{ type: 'string', enum: VALID_EVENT_MODULES },
+							{
+								type: 'array',
+								itmes: { type: 'string', enum: VALID_EVENT_MODULES },
+								minItems: 1,
+							},
+						],
 					},
 				},
 				additionalProperties: false,
 			},
 		],
 	},
-	defaultOptions: [{ eventModules: VALID_EVENT_MODULES }],
+	defaultOptions: [{ eventModules: FIRE_EVENT_OPTION }],
 
 	create(context, [options], helpers) {
 		const { eventModules = VALID_EVENT_MODULES } = options;
@@ -90,6 +109,11 @@ export default createTestingLibraryRule<Options, MessageIds>({
 			},
 			'AwaitExpression > CallExpression'(node: TSESTree.CallExpression) {
 				const simulateEventFunctionIdentifier = getDeepestIdentifierNode(node);
+				const enabledEventModules = getEnabledEventModules(eventModules);
+				const isFireEventEnabled =
+					enabledEventModules.includes(FIRE_EVENT_OPTION);
+				const isUserEventEnabled =
+					enabledEventModules.includes(USER_EVENT_OPTION);
 
 				if (!simulateEventFunctionIdentifier) {
 					return;
@@ -107,10 +131,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
 					return;
 				}
 
-				if (isFireEventMethod && !eventModules.includes('fire-event')) {
+				if (isFireEventMethod && !isFireEventEnabled) {
 					return;
 				}
-				if (isUserEventMethod && !eventModules.includes('user-event')) {
+				if (isUserEventMethod && !isUserEventEnabled) {
 					return;
 				}
 
--- a/tests/lib/rules/no-await-sync-events.test.ts
+++ b/tests/lib/rules/no-await-sync-events.test.ts
@@ -219,6 +219,16 @@ ruleTester.run(RULE_NAME, rule, {
       `,
 			options: [{ eventModules: ['fire-event'] }],
 		})),
+
+		// valid tests for user-event with default options (user-event disabled)
+		...USER_EVENT_SYNC_FUNCTIONS.map((func) => ({
+			code: `
+        import userEvent from '@testing-library/user-event';
+        test('should not report userEvent.${func} by default', async() => {
+          await userEvent.${func}('foo');
+        });
+      `,
+		})),
 	],
 
 	invalid: [
@@ -254,6 +264,7 @@ ruleTester.run(RULE_NAME, rule, {
           await userEvent.${func}('foo');
         });
       `,
+					options: [{ eventModules: 'user-event' }],
 					errors: [
 						{
 							line: 4,
@@ -277,7 +288,6 @@ ruleTester.run(RULE_NAME, rule, {
           await fireEvent.${func}('foo');
         });
       `,
-						options: [{ eventModules: ['fire-event'] }],
 						errors: [
 							{
 								line: 4,
@@ -289,8 +299,7 @@ ruleTester.run(RULE_NAME, rule, {
 					} as const)
 			)
 		),
-		// sync userEvent sync methods with await operator are not valid
-		// when only fire-event set in eventModules
+
 		...USER_EVENT_SYNC_FUNCTIONS.map(
 			(func) =>
 				({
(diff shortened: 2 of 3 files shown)


lib/rules/no-await-sync-events.ts
```diff
@@ -41,18 +56,22 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
 				properties: {
 					eventModules: {
-						type: 'array',
-						minItems: 1,
-						items: {
-							enum: VALID_EVENT_MODULES,
-						},
+						default: FIRE_EVENT_OPTION,
+						oneOf: [
+							{ type: 'string', enum: VALID_EVENT_MODULES },
```
MichaelDeBoey, line 61: I would keep this as an array only tbh as the name is representing multiple values

MichaelDeBoey: Since the counterpart rule is singular, I would rename that to multiple in the next major release and only support passing an array (even if it's only 1 value) in both rules tbh

Belco90: Makes sense, I'll do that.

lib/configs/angular.ts
```diff
@@ -11,7 +11,10 @@ export = {
 		],
 		'testing-library/await-async-queries': 'error',
 		'testing-library/await-async-utils': 'error',
-		'testing-library/no-await-sync-events': 'error',
+		'testing-library/no-await-sync-events': [
+			'error',
+			{ eventModules: 'fire-event' },
```
MichaelDeBoey, line 16:
```suggestion
			{ eventModules: ['fire-event'] },
```

lib/configs/dom.ts
```diff
@@ -11,7 +11,10 @@ export = {
 		],
 		'testing-library/await-async-queries': 'error',
 		'testing-library/await-async-utils': 'error',
-		'testing-library/no-await-sync-events': 'error',
+		'testing-library/no-await-sync-events': [
+			'error',
+			{ eventModules: 'fire-event' },
```
MichaelDeBoey, line 16:
```suggestion
			{ eventModules: ['fire-event'] },
```

lib/configs/react.ts
```diff
@@ -11,7 +11,10 @@ export = {
 		],
 		'testing-library/await-async-queries': 'error',
 		'testing-library/await-async-utils': 'error',
-		'testing-library/no-await-sync-events': 'error',
+		'testing-library/no-await-sync-events': [
+			'error',
+			{ eventModules: 'fire-event' },
```
MichaelDeBoey, line 16:
```suggestion
			{ eventModules: ['fire-event'] },
```

lib/rules/no-await-sync-events.ts
```diff
@@ -25,9 +40,9 @@ export default createTestingLibraryRule<Options, MessageIds>({
 		docs: {
 			description: 'Disallow unnecessary `await` for sync events',
 			recommendedConfig: {
-				dom: 'error',
-				angular: 'error',
-				react: 'error',
+				dom: ['error', { eventModules: 'fire-event' }],
+				angular: ['error', { eventModules: 'fire-event' }],
+				react: ['error', { eventModules: 'fire-event' }],
```
MichaelDeBoey, line 45:
```suggestion
				dom: ['error', { eventModules: ['fire-event'] }],
				angular: ['error', { eventModules: ['fire-event'] }],
				react: ['error', { eventModules: ['fire-event'] }],
```

lib/rules/no-await-sync-events.ts
```diff
@@ -41,18 +56,22 @@ export default createTestingLibraryRule<Options, MessageIds>({
 				type: 'object',
 				properties: {
 					eventModules: {
-						type: 'array',
-						minItems: 1,
-						items: {
-							enum: VALID_EVENT_MODULES,
-						},
+						default: FIRE_EVENT_OPTION,
```
MichaelDeBoey, line 59:
```suggestion
						default: [FIRE_EVENT_OPTION],
```

lib/rules/no-await-sync-events.ts
```diff
@@ -41,18 +56,22 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+							},
+						],
 					},
 				},
 				additionalProperties: false,
 			},
 		],
 	},
-	defaultOptions: [{ eventModules: VALID_EVENT_MODULES }],
+	defaultOptions: [{ eventModules: FIRE_EVENT_OPTION }],
```
MichaelDeBoey, line 74:
```suggestion
	defaultOptions: [{ eventModules: [FIRE_EVENT_OPTION] }],
```

tests/lib/rules/no-await-sync-events.test.ts
```diff
@@ -254,6 +264,7 @@ ruleTester.run(RULE_NAME, rule, {
           await userEvent.${func}('foo');
         });
       `,
+					options: [{ eventModules: 'user-event' }],
```
MichaelDeBoey, line 267:
```suggestion
					options: [{ eventModules: ['user-event'] }],
```

MichaelDeBoey (changes requested):

MichaelDeBoey (changes requested):
We should release this in a major release btw as this is breaking behavior for this specific rule and people need to adjust their config if they want the current released functionality (for instance when only enabling this specific rule)

lib/rules/no-await-sync-events.ts
```diff
@@ -10,13 +10,27 @@ import {
...
+
+function getEnabledEventModules(
+	eventModulesOption?: EventModulesOptions | undefined
+): ReadonlyArray<ValidEventModules> {
+	if (typeof eventModulesOption === 'undefined') {
+		// This should match the default option
+		return [FIRE_EVENT_OPTION];
+	}
+
+	return eventModulesOption;
```
MichaelDeBoey, line 32: This can be simplified to

```suggestion
  // This should match the default option
	return eventModulesOption || [FIRE_EVENT_OPTION];
```

MichaelDeBoey (changes requested):

lib/rules/no-await-sync-events.ts
```diff
@@ -90,6 +103,11 @@ export default createTestingLibraryRule<Options, MessageIds>({
 			},
 			'AwaitExpression > CallExpression'(node: TSESTree.CallExpression) {
 				const simulateEventFunctionIdentifier = getDeepestIdentifierNode(node);
+				const enabledEventModules = getEnabledEventModules(eventModules);
+				const isFireEventEnabled =
+					enabledEventModules.includes(FIRE_EVENT_OPTION);
+				const isUserEventEnabled =
+					enabledEventModules.includes(USER_EVENT_OPTION);
```
MichaelDeBoey, line 110: These can be inlined

```suggestion
```

lib/rules/no-await-sync-events.ts
```diff
@@ -107,10 +125,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
 					return;
 				}
 
-				if (isFireEventMethod && !eventModules.includes('fire-event')) {
+				if (isFireEventMethod && !isFireEventEnabled) {
```
MichaelDeBoey, line 128:
```suggestion
				if (isFireEventMethod && !enabledEventModules.includes(FIRE_EVENT_OPTION)) {
```

lib/rules/no-await-sync-events.ts
```diff
@@ -107,10 +125,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
 					return;
 				}
-				if (isUserEventMethod && !eventModules.includes('user-event')) {
+				if (isUserEventMethod && !isUserEventEnabled) {
```
MichaelDeBoey, line 131:
```suggestion
				if (isUserEventMethod && !enabledEventModules.includes(USER_EVENT_OPTION)) {
```

MichaelDeBoey (changes requested):

lib/rules/no-await-sync-events.ts
```diff
@@ -25,9 +39,9 @@ export default createTestingLibraryRule<Options, MessageIds>({
 		docs: {
 			description: 'Disallow unnecessary `await` for sync events',
 			recommendedConfig: {
-				dom: 'error',
-				angular: 'error',
-				react: 'error',
+				dom: ['error', { eventModules: ['fire-event'] }],
+				angular: ['error', { eventModules: ['fire-event'] }],
+				react: ['error', { eventModules: ['fire-event'] }],
```
MichaelDeBoey, line 44: This way we make sure we never have a typo and if we would remove `FIRE_EVENT_OPTION`, we'll get a TS error here

```suggestion
				dom: ['error', { eventModules: [FIRE_EVENT_OPTION] }],
				angular: ['error', { eventModules: [FIRE_EVENT_OPTION] }],
				react: ['error', { eventModules: [FIRE_EVENT_OPTION] }],
```

MichaelDeBoey (changes requested):

lib/rules/no-await-sync-events.ts
MichaelDeBoey: We're not re-using this variable anywhere at all, so this way we keep the code changes as small as possible

lib/rules/no-await-sync-events.ts
```diff
@@ -10,13 +10,23 @@ import {
...
+
+type ValidEventModules = (typeof VALID_EVENT_MODULES)[number];
+type EventModulesOptions = ReadonlyArray<ValidEventModules>;
+type Options = [{ eventModules?: EventModulesOptions }];
+
+function getEnabledEventModules(
+	eventModulesOption?: EventModulesOptions | undefined
+): ReadonlyArray<ValidEventModules> {
+	return eventModulesOption ?? DEFAULT_EVENT_MODULES;
+}
```
MichaelDeBoey, line 29: This is basically just setting a default value now, so we can inline this in the `options` destructuring

```suggestion
```

lib/rules/no-await-sync-events.ts
```diff
@@ -42,17 +52,16 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
 				},
 				additionalProperties: false,
 			},
 		],
 	},
-	defaultOptions: [{ eventModules: VALID_EVENT_MODULES }],
+	defaultOptions: [{ eventModules: DEFAULT_EVENT_MODULES }],
 
 	create(context, [options], helpers) {
 		const { eventModules = VALID_EVENT_MODULES } = options;
```
MichaelDeBoey, line 67:
```suggestion
		const { eventModules = DEFAULT_EVENT_MODULES } = options;
```

lib/rules/no-await-sync-events.ts
```diff
@@ -90,6 +99,11 @@ export default createTestingLibraryRule<Options, MessageIds>({
 			},
 			'AwaitExpression > CallExpression'(node: TSESTree.CallExpression) {
 				const simulateEventFunctionIdentifier = getDeepestIdentifierNode(node);
+				const enabledEventModules = getEnabledEventModules(eventModules);
+				const isFireEventEnabled =
+					enabledEventModules.includes(FIRE_EVENT_OPTION);
+				const isUserEventEnabled =
+					enabledEventModules.includes(USER_EVENT_OPTION);
```
MichaelDeBoey, line 106:
```suggestion
				const isFireEventEnabled = eventModules.includes(FIRE_EVENT_OPTION);
				const isUserEventEnabled = eventModules.includes(USER_EVENT_OPTION);
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

PR #830, opened 2023-10-15
feat(await-async-events): instance of userEvent is recognized as async

### LINKED ISSUE #812: `await-async-events` doesn't seem to check calls from user sessions returned by userEvent.setup()

await-async-events

More warnings

The code needs to detect calls to `userEvent.setup()` and calls to the resulting object's methods.

### Example code

```js
import userEvent from '@testing-library/user-event';
const user = userEvent.setup();
user.click(domElement);
```

`user.click()` and others returns a `Promise` and should be `await`ed.

```js
import userEvent from '@testing-library/user-event';
const user = userEvent.setup();
await user.click(domElement);
```

### Anything else?

see also https://github.com/testing-library/eslint-plugin-testing-library/issues/800

No

Kvanttinen:
## Checks

## Changes

Adds a feature that events from the instance of userEvent are recognized as async events.

Example:
const user = userEvent.setup();
user.click(button);

The 'user' will be flagged in this case.

## Context

Closes #812

My first open source PR, and not used to GitHub that much, so there might be some errors with the process. Changes itself are quite simple, but I didn't find a more elegant way to do that.

### DIFF
--- a/lib/create-testing-library-rule/detect-testing-library-utils.ts
+++ b/lib/create-testing-library-rule/detect-testing-library-utils.ts
@@ -72,7 +72,10 @@ type IsAsyncUtilFn = (
 	validNames?: readonly (typeof ASYNC_UTILS)[number][]
 ) => boolean;
 type IsFireEventMethodFn = (node: TSESTree.Identifier) => boolean;
-type IsUserEventMethodFn = (node: TSESTree.Identifier) => boolean;
+type IsUserEventMethodFn = (
+	node: TSESTree.Identifier,
+	userEventSession?: string
+) => boolean;
 type IsRenderUtilFn = (node: TSESTree.Identifier) => boolean;
 type IsCreateEventUtil = (
 	node: TSESTree.CallExpression | TSESTree.Identifier
@@ -97,6 +100,9 @@ type FindImportedTestingLibraryUtilSpecifierFn = (
 type IsNodeComingFromTestingLibraryFn = (
 	node: TSESTree.Identifier | TSESTree.MemberExpression
 ) => boolean;
+type getUserEventImportIdentifierFn = (
+	node: ImportModuleNode | null
+) => TSESTree.Identifier | null;
 
 export interface DetectionHelpers {
 	getTestingLibraryImportNode: GetTestingLibraryImportNodeFn;
@@ -130,6 +136,7 @@ export interface DetectionHelpers {
 	canReportErrors: CanReportErrorsFn;
 	findImportedTestingLibraryUtilSpecifier: FindImportedTestingLibraryUtilSpecifierFn;
 	isNodeComingFromTestingLibrary: IsNodeComingFromTestingLibraryFn;
+	getUserEventImportIdentifier: getUserEventImportIdentifierFn;
 }
 
 const USER_EVENT_PACKAGE = '@testing-library/user-event';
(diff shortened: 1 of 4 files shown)


ka2jun8:
any updates?

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -326,6 +333,35 @@ export function detectTestingLibraryUtils<
 			return getImportModuleName(importedCustomModuleNode);
 		};
 
+		const getUserEventImportIdentifier = (node: ImportModuleNode | null) => {
+			if (!node) {
+				return null;
+			}
+
+			if (isImportDeclaration(node)) {
```
Belco90, line 341: This function should always receive a `node`. If the node is not available, then avoid calling this function in the corresponding code.
```suggestion
		const getUserEventImportIdentifier = (node: ImportModuleNode) => {
			if (isImportDeclaration(node)) {
```

lib/create-testing-library-rule/detect-testing-library-utils.ts
```diff
@@ -853,35 +895,7 @@ export function detectTestingLibraryUtils<
...
-					const requireNode = importedUserEventLibraryNode.parent;
-					if (!ASTUtils.isIdentifier(requireNode.id)) {
-						return null;
-					}
-
-					return requireNode.id;
-				}
-
-				return null;
+				return getUserEventImportIdentifier(importedUserEventLibraryNode);
```
Belco90, line 898: Here you should call `getUserEventImportIdentifier` only if `importedUserEventLibraryNode` exists.

lib/node-utils/index.ts
```diff
@@ -679,3 +679,37 @@ export function findImportSpecifier(
...
+	for (const [index, token] of tokensAndComments.entries()) {
+		if (
+			token.type === 'Identifier' &&
+			token.value === userEventImport.name &&
+			tokensAndComments[index + 1].value === '.' &&
+			tokensAndComments[index + 2].value === 'setup' &&
+			tokensAndComments[index + 3].value === '(' &&
+			tokensAndComments[index - 1].value === '='
+		) {
+			return tokensAndComments[index - 2].value;
```
Belco90, line 711: This looks a bit brittle and hardcoded… I'm trying to think a better way to implement this in a more generic way with AST selectors.

lib/node-utils/index.ts
```diff
@@ -679,3 +679,37 @@ export function findImportSpecifier(
 		return (property as TSESTree.Property).key as TSESTree.Identifier;
 	}
 }
+
+/**
+ * Finds if the userEvent is used as an instance
+ */
+
+export function getUserEventInstance(
```
Belco90, line 687: I think this one should belong to the `detect-testing-library-utils` too.

tests/lib/rules/await-async-events.test.ts
```diff
@@ -32,7 +32,7 @@ const USER_EVENT_ASYNC_FUNCTIONS = [
 	'upload',
 ] as const;
 const FIRE_EVENT_ASYNC_FRAMEWORKS = [
-	'@testing-library/vue',
+	// '@testing-library/vue',
```
Belco90, line 35: This shouldn't be commented.

Belco90:
I gave it a first round. After the changes requested are applied, I'll give it another round.

ttbui11:
any updates on this please?

MichaelDeBoey:
@Kvanttinen Are you still interested in implementing this feature?

If so, can you please rebase onto latest `main` & fix @Belco90's remarks?

PR #835, opened 2023-10-25
fix: handle multiple imports with `no-manual-cleanup`

maisano:
## Checks

## Changes

Crawl all the import references when sniffing out `no-manual-cleanup` violations.

## Context

There is currently a bug with `no-manual-cleanup` where we only ever check the first import of any `testing-library` package. While duplicative imports of the _same_ package is very likely uncommon, there are many packages within the `testing-library` ecosystem that will trip up the package detection and break this rule when we only look at the first.

### DIFF
--- a/lib/rules/no-manual-cleanup.ts
+++ b/lib/rules/no-manual-cleanup.ts
@@ -111,15 +111,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
 		return {
 			'Program:exit'() {
 				const testingLibraryImportName = helpers.getTestingLibraryImportName();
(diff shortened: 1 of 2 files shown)


maisano-patreon:
👋  Curious if I can get some eyes this, cc @Belco90

Belco90:
I'll try to review this next week.

maisano-patreon:
Hi hi – bumping this again! Should be a very quick review. The PR just accounts for when someone might have multiple testing library imports (which is common in my company's codebase).

Belco90 (approved):
LGTM, thanks for taking care of this!

Belco90:
@all-contributors please add @maisano for code and test

Issue #818, opened 2023-09-25
`await-async-events` reports a problem when `userEvent.setup` is called from another function


dfernandez79:
### Have you read the Troubleshooting section?

Yes

### Plugin version

v6.0.2

### ESLint version

v8.50.0

### Node.js version

18.18.0

### package manager and version

npm 9.8.1

### Operating system

macOS Ventura 13.5.2 (22G91)

### Bug description

The rule `await-async-events` with `Promise returned from `setUpUserEvents` wrapper over async event method must be handled` when I call `userEvent.setup()` from `setUpUserEvents`.

Example:
```tsx
import userEvent, { type UserEvent } from '@testing-library/user-event';

const setUpUserEvents = (): UserEvent =>
  userEvent.setup({ advanceTimers: jest.advanceTimersByTime });

test('Something', async () => {
  const { click } = setUpUserEvents();

  /* ... */
  await click(someElement);
});
```

### Steps to reproduce

1. Create a test that uses `userEvent.setup()`
2. Run lint w/`await-async-events` (all ok)
3. Refactor the call to `userEvent.setup()` into another function
4. Run lint w/`await-async-events` (eslint fails)

### Error output/screenshots

Before the refactor:
<img width="819" alt="image" src="https://github.com/testing-library/eslint-plugin-testing-library/assets/1102266/2b18ad79-1a49-43cd-aed7-31b450260c3b">

After the refactor (eslint error):
<img width="580" alt="image" src="https://github.com/testing-library/eslint-plugin-testing-library/assets/1102266/f2e57e86-177d-4b27-aad0-1b85d6030c5e">

<img width="758" alt="image" src="https://github.com/testing-library/eslint-plugin-testing-library/assets/1102266/61897ef5-c45d-473e-9e78-42d7adbf5490">

Eslint error:
```
error  Promise returned from `setUpUserEvents` wrapper over async event method must be handled  testing-library/await-async-events
```

### ESLint configuration

I'm extending AirBnb rules with:

```
    {
      files: ['**/*.test.ts?(x)'],
      env: {
        jest: true,
      },
      extends: ['plugin:testing-library/react', 'plugin:jest-dom/recommended'],
    },
```

### Rule(s) affected

`testing-library/await-async-events`

### Anything else?

The PR https://github.com/testing-library/eslint-plugin-testing-library/pull/817 is checking for the name `setup`.

Maybe that isn't enough.

### Do you want to submit a pull request to fix this bug?

No

[...]

PR #834, opened 2023-10-24
fix(await-async-events): avoid reporting `userEvent.setup()` in wrappers

doochik:
## Checks

- [X] I have read the [contributing guidelines](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/CONTRIBUTING.md).

## Changes

Skipping all CallExpressions with nodes named "setup" in the await-async-events rule

## Context

Better fix for #800 after #817

PR #840, opened 2023-11-03
fix(await-async-events): setup() incorrectly reported from a wrapper

trungutt:
## Checks

## Changes

- `userEvent.setup()` is [sync](https://testing-library.com/docs/user-event/setup/#starting-a-session-per-setup). So we shouldn't report any async linting error on it

## Context

- Fix https://github.com/testing-library/eslint-plugin-testing-library/issues/818
- I'm not super familiar with the codebase, please let me know if anything goes wrong

### DIFF
--- a/lib/rules/await-async-events.ts
+++ b/lib/rules/await-async-events.ts
@@ -91,9 +91,17 @@ export default createTestingLibraryRule<Options, MessageIds>({
 			messageId?: MessageIds;
 			fix?: TSESLint.ReportFixFunction;
 		}): void {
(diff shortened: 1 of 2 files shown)


trungutt:
@Belco90 it sounds like you would be the right reviewer. Could you PTAL?

Belco90:
Hi @trungutt. Thanks for your contribution! Can you check if this is still relevant after merging #834? If so, please update your branch accordingly.

trungutt:
it should be good now

PR #851, opened 2023-12-01
docs: update eslint pattern matching

NicolasBonduel:
## Checks

## Changes

I believe there is a mistake in the README, where the `s` should not be optional, but the `x` should

    *.[jt]s?(x):
        [jt]: Matches a single character, either 'j' or 't'.
        s?: The 's' is optional.
        (x): This matches an `x`, however, parenthesis are not a standard syntax for file globbing.

    *.[jt]sx?:
        [jt]: Matches a single character, either 'j' or 't'.
        s: This 's' is required.
        x?: The 'x' is optional.
        

## Context

File matching in eslint configuration example in the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ module.exports = {
 	overrides: [
 		{
 			// 3) Now we enable eslint-plugin-testing-library rules or preset only for matching testing files!
-			files: ['**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+(spec|test).[jt]s?(x)'],
+			files: ['**/__tests__/**/*.[jt]sx?', '**/?(*.)+(spec|test).[jt]sx?'],
 			extends: ['plugin:testing-library/react'],
 		},
 	],


Belco90 (approved):
Hi @NicolasBonduel! Thanks for reporting. I got that pattern from [Jest's `testMatch` docs](https://jestjs.io/docs/next/configuration#testmatch-arraystring) directly, but I guess it makes more sense to get the updated pattern you suggest here.

Belco90:
@all-contributors pleased add @NicolasBonduel for docs.

joshunger:
@Belco90 @NicolasBonduel 👋 just ran across this. The new pattern doesn't work and no longer matches. 

> Zero or One Pattern ?(a|b)

See https://www.digitalocean.com/community/tools/glob

Belco90:
@joshunger Interesting, thanks for reporting! I'll revert it for now until we figure out if there is a better pattern. I took the original one from Jest docs, so definitely works.

PR #852, opened 2023-12-06
chore: upgrade `@typescript-eslint`

lesha1201:
`eslint-plugin-testing-library` has an outdated `dependency` - `@typescript-eslint/utils` and because of it, in modern setups, it might result in duplicate `@typescript-eslint/parser` packages.

This PR upgrades all `@typescript-eslint` dependencies as well as TypeScript.

Because `@typescript-eslint@6` uses package.json `exports`, we had to upgrade some dependencies which didn't support `exports` (see https://github.com/typescript-eslint/typescript-eslint/issues/7284). Affected dependencies:

- `jest`

I also dropped support for Node.js versions lower than 16 because they're not maintained anymore (v16 also not maintained but decided to leave it for now).

image

## `@typescript-eslint@6`

- https://typescript-eslint.io/blog/announcing-typescript-eslint-v6

Upgrading `@typescript-eslint` brought new rules. The rules I fixed:

- [`@typescript-eslint/no-redundant-type-constituents`](https://typescript-eslint.io/rules/no-redundant-type-constituents/)
- [`@typescript-eslint/no-confusing-void-expression`](https://typescript-eslint.io/rules/no-confusing-void-expression)


.eslintrc.js
```diff
@@ -1,7 +1,12 @@
 module.exports = {
 	root: true,
+	env: {
+		es6: true,
+		node: true,
+	},
 	extends: [
-		'kentcdodds',
```
lesha1201, line 4: `eslint-config-kentcdodds` uses some outdated packages which block us from upgrading TypeScript dependencies properly. It didn't bring many rules so decided just use `eslint:recommended`, `plugin:import/recommended` and configure some rules manually.

.github/workflows/pipeline.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       fail-fast: false
       matrix:
         # The .x indicates "the most recent one"
-        node: [19.x, 18.x, 17.x, 16.x, 14.x, 14.17.0, 12.x, 12.22.0]
```
lesha1201, line 56: Dropped support for Node.js versions lower than 16. They're not maintained long time ago and doesn't support package.json `exports` which is used at least in `@typescript-eslint`.

.prettierignore
```diff
@@ -0,0 +1,6 @@
+# Files
+package-lock.json
+
+# Folders
+node_modules/
+dist/
```
lesha1201, line 6: Add `.prettierignore` to ignore `dist` folder.

.eslintrc.js
```diff
@@ -44,8 +55,8 @@ module.exports = {
 				project: ['./tsconfig.eslint.json'],
 			},
 			extends: [
-				'plugin:@typescript-eslint/recommended',
-				'plugin:@typescript-eslint/recommended-requiring-type-checking',
+				'plugin:@typescript-eslint/strict-type-checked',
```
lesha1201, line 58: We can also add `plugin:@typescript-eslint/stylistic-type-checked` which brings some useful rules but I wasn't sure what rules should be used in the project.

lib/rules/no-await-sync-events.ts
```diff
@@ -76,6 +76,7 @@ export default createTestingLibraryRule<Options, MessageIds>({
 						property.id.name === 'delay' &&
 						isLiteral(property.init) &&
 						property.init.value &&
+						// @ts-expect-error -- TODO: fix me
 						property.init.value > 0
```
lesha1201, line 80: Previously, we had `suppressImplicitAnyIndexErrors: true` in `tsconfig.json` which allowed such code. But in TypeScript v5, it complains about this configuration so I removed it.

tsconfig.json
```diff
@@ -2,17 +2,16 @@
 	"compilerOptions": {
 		"strict": true,
 		"target": "es6",
-		"module": "commonjs",
-		"moduleResolution": "node",
+		"module": "Node16",
+		"moduleResolution": "Node16",
```
lesha1201, line 6: We need this to support `exports` from `package.json` (https://github.com/typescript-eslint/typescript-eslint/issues/7284).

tsconfig.json
```diff
@@ -2,17 +2,16 @@
...
 		"forceConsistentCasingInFileNames": true,
 		"noImplicitAny": true,
 		"outDir": "./dist",
 		"removeComments": true,
 		"skipLibCheck": true,
-		"sourceMap": false,
-		"suppressImplicitAnyIndexErrors": true
+		"sourceMap": false
 	},
 	"include": ["./lib/**/*.ts"]
```
lesha1201, line 16: I think we should include all TS files by default so IDE can process all TS files and have a separate TS config for building the package. Decided to not do this in this PR.

Belco90:
Hi @lesha1201. Thanks for your contribution, but I'm afraid I have to reject this PR until ESLint v9 is released, and we bump the plugin to the next major:

- We don't want to drop support for Node.js lower than v18 yet. We are aware they are not officially supported, but we want to support the same supported version as ESLint. [They'll drop support for Node prior to v18.18 in ESLint v9](https://eslint.org/blog/2023/11/whats-coming-in-eslint-9.0.0/#dropping-support-for-node.js-%3C-v18.18.0%2C-v19), so we will do it once ESLint v9 is released.
- For the previous reason, we are keeping `@typescript-eslint/utils` in v5. We will upgrade it in the next major together with dropping non-supported Node.js versions.
- Upgrading TS to v5 should be addressed in a separate PR, it's too much for mixing it with other stuff. Same for other dependencies like jest.

.eslintrc.js
Belco90: I like this change! We should apply it anyway.

lesha1201:
@Belco90 

> We don't want to drop support for Node.js lower than v18 yet. We are aware they are not officially supported, but we want to support the same supported version as ESLint. [They'll drop support for Node prior to v18.18 in ESLint v9](https://eslint.org/blog/2023/11/whats-coming-in-eslint-9.0.0/#dropping-support-for-node.js-%3C-v18.18.0%2C-v19), so we will do it once ESLint v9 is released.

We only dropped support for Node.js lower than v16 in this PR. And it might work fine in v12 and v14 because they also support package.json `exports` but I didn't test it (https://nodejs.org/api/packages.html#exports) since `typescript-eslint` dropped official support for v12 and v14.

The main issue that motivated me to open this PR is that a lot of popular ESLint plugins uses `typescript-eslint@6` and the most of TS projects use it as well which results in duplicated `@typescript-eslint/parser` when using it with `eslint-plugin-testing-library`.

ESLint v9 won't be released soon. We can create a major release now that upgrades TS related packages and then create another major release that upgrades ESLint v9 and something else. Users who rely on old Node.js versions can use the current version of `eslint-plugin-testing-library` and if they report some issue that can be patched, we can create a patch release for them. Another option is to create `alpha` release so the users can start using it eliminating the issue with duplicated `@typescript-eslint/parser`.

---

Should I open separate PRs for some changes? If so can you please write a list of changes that you would like to see in separate PRs? Also do you have other tasks I can work on? For example, we can upgrade Prettier.

Belco90:
> We only dropped support for Node.js lower than v16 in this PR

That's already a breaking change, but also it's not in the same page as versions supported by ESLint v8.

> ESLint v9 won't be released soon. We can create a major release now that upgrades TS related packages and then create another major release that upgrades ESLint v9 and something else.

We could indeed. No idea when ESLint will be released, but I want to avoid releasing a major just for the sake of upgrading this typescript-eslint.

I get the motivation for all the updates you did. However, we can't perform them all at once, and some will have to wait a bit. In order to keep discussing it, I'll open a new issue.

PR #856, opened 2023-12-12
chore: remove `eslint-config-kentcdodds`

lesha1201:
## Checks

## Context

These changes were initially introduced in https://github.com/testing-library/eslint-plugin-testing-library/pull/852


.eslintrc.js
```diff
@@ -54,6 +66,24 @@ module.exports = {
...
+				// TypeScript and @typescript-eslint.
+				'import/default': 'off',
+				'import/export': 'off',
+				'import/namespace': 'off',
+				'import/no-unresolved': 'off',
+			},
+			settings: {
+				'import/resolver': {
+					node: {
+						extensions: ['.js', '.jsx', '.ts', '.tsx'],
```
Belco90, line 81: We don't need jsx and tsx in this repo.

Belco90 (changes requested):

.eslintrc.js
lesha1201: Done

lesha1201:
@Belco90 By the way, do you have some conventions/rules in mind that you want to enforce? I would like to improve ESLint setup  later but not very familiar with the codebase yet and its conventions.

lesha1201:
@Belco90 CI will fail for Node v12 because of this https://github.com/testing-library/eslint-plugin-testing-library/pull/857#issuecomment-1855338100. Previously, `eslint-config-kentcdodds` were taking precedence over our `extends: ['plugin:jest/recommended']` and because of that ESLint were using an old version of `eslint-plugin-jest` which support Node.js v12. But now it uses our version of `eslint-plugin-jest` which doesn't support Node.js v12.

image

Belco90:
> @Belco90 By the way, do you have some conventions/rules in mind that you want to enforce? I would like to improve ESLint setup later but not very familiar with the codebase yet and its conventions.

I have something in mind, but let's do that in a follow-up PR.

Belco90 (approved):
Sorry for the late response! The Christmas break was way busier than I thought, so I'm slowly coming back to this now.

Belco90:
@all-contributors please add @lesha1201 for maintenance

PR #863, opened 2024-01-04
docs: add missing import

stefannibrasil:
## Checks

- [ X] I have read the [contributing guidelines](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/CONTRIBUTING.md).

## Changes

I was going through this example to understand a failure I was getting. I noticed that the import methods from the testing-library was missing the `render` method. Not sure if that's intentional since the examples are for any testing library. Thanks!


Belco90 (approved):
Thanks for your contribution! We probably missed this import.

PR #889, opened 2024-03-18
Disallow usages of innerHTML for no-container rule

### LINKED ISSUE #883: Fail no-container on use of innerHTML

no-container

More warnings

The rule will look for usages of `container.innerHTML` in addition to looking for usages of methods like `container.querySelector()`

### Example code

```
const { container } = render(<Greeting />);
expect(container.innerHTML).toContain("Hello");
```

The current rule allows this code. It only errors if a method like `container.querySelector` is accessed.

The updated rule would fail on usages of `container.innerHTML` just as it does with the methods.

### Anything else?

Descriptions of the rule would need to be updated from saying "disallow the user of container methods" to "container methods and properties" (or something)

If we found that anyone has use cases where using innerHTML is commonly needed but container methods are not, we could either make it configurable whether it's allowed, or make it a separate rule very similar to `no-container`

Yes

CodingItWrong:
## Checks

## Changes

- So far, updates the rule to reject when innerHTML is accessed on a destructured `container` proprety
- QUESTION: one test case is not yet covered: when the return value of a render function is assigned to a render result variable like `view`, and then it is accessed as `view.container.innerHTML`. For container methods, this seems to be caught by the code on lines 95-106. But we can't use that code as-is in `MemberExpression()` because of the error `Property 'callee' does not exist on type 'MemberExpression'.`. Looking over our code, do you have an ideas on how we could detect the `view.container.innerHTML` case in `MemberExpression()`? @Belco90 

## Context

Fixes #883


tests/lib/rules/no-container.test.ts
```diff
@@ -232,5 +322,101 @@ ruleTester.run(RULE_NAME, rule, {
...
+		//     view.container.innerHTML;
+		//   `,
+		// 	errors: [
+		// 		{
+		// 			line: 3,
+		// 			column: 29,
+		// 			messageId: 'noContainer',
+		// 		},
+		// 	],
+		// },
```
CodingItWrong, line 420: @Belco90 This comment-out test is the test that is failing now, that shows us that we are not yet handling the `view.container.innerHTML` case

Belco90: I see. I need to go again through the utils for navigating the nodes, I think we had something to fix the problem described in the main description.

MichaelDeBoey:
@CodingItWrong Are you still interested in implementing this feature?

If so, can you please rebase onto latest `main`?

CodingItWrong:
Hi @MichaelDeBoey, I've moved on from the consultancy where I was working when implementing this functionality, and unfortunately I'm no longer able to continue work on it.

PR #890, opened 2024-03-22
fix(await-async-events): false positive reports on awaited expressions evaluating to promise

### LINKED ISSUE #887: False positive for `testing-library/await-async-events` when returning using a ternary

Yes

### Plugin version

v6.2.0

### ESLint version

v8.56.0

### Node.js version

20.11.1

yarn 3.2.2

### Operating system

macOS Ventura 13.6.4

### Bug description

Ternary return of `userEvent.type` or `userEvent.clear` is reporting testing-library/await-async-queries, even though the query will get awaited

### Steps to reproduce

1. Go to test file
2. Add code like
```js
async function changeEmailInput(value) {
	const emailInput = screen.getByRole('textbox', { name: 'Email' });

	return value.length
		? userEvent.type(emailInput, value)
		: userEvent.clear(emailInput);
}
```
3. Observe errors on the last two lines 
> Promise returned from async event method `type` must be handled eslint [testing-library/await-async-events](https://github.com/testing-library/eslint-plugin-testing-library/tree/main/docs/rules/await-async-events.md)
> Promise returned from async event method `clear` must be handled eslint [testing-library/await-async-events](https://github.com/testing-library/eslint-plugin-testing-library/tree/main/docs/rules/await-async-events.md)

### Error output/screenshots

_No response_

### ESLint configuration

```
{
  ...,
  plugins: [
     ...,
     'eslint-plugin-testing-library'
  ],
  rules: {
    ...,
    'testing-library/await-async-queries': ['error', { eventModule: 'userEvent' }]
  }
}

### Rule(s) affected

testing-library/await-async-queries

### Anything else?

_No response_

No

### PR DESCRIPTION
**Pull request by Chamion**
## Checks

## Changes

In `await-async-events`, `await-async-queries` and `await-async-utils` when determining whether a promise is handled the checks are performed on a parent expression that evaluates to the call expression returning a promise instead of always the call expression. In the following cases the parent expression is used instead.

```ts
// sequence expression
(sideEffect(), promise());
// conditional expression
condition ? promise() : otherPromise();
// logical expression
maybePromise() ?? promise();
maybePromise() || promise();
condition && promise();
```

This change does not eliminate false positives from the three `await-async-` rules. The rules are fundamentally written in a brittle way where they look for specific reasons not to report and default to reporting. This means the rules are not robust against changes to the language so even if we defined perfect behaviour for all cases in current JavaScript a new language feature could introduce false positives again.

For the rules to be robust and guaranteed not to report false positives they would have to default to not report unless they find a specific reason to report. In 4d878d1 I took this approach but I noticed the other two rules seem to have a requirement of reporting in some unsafe cases (e.g. `expect(query).toBeInTheDocument()`) so I reverted the changes and opted to carve out exceptions for specific cases instead. Whether the rules ought to be safe with regard to false positives or if they're intentionally opinionated is another conversation. I left the safer approach in the commit history in case we want to revisit that idea. Let me know if you prefer a neatly squashed commit history instead.

## Context

@sjarva invited me to contribute and I picked an issue off the top.

Fixes #887

Although the bug report was about conditional expressions the same bug could be reproduced with other expressions as well. This PR covers all the cases I could think of.

### DIFF
--- a/lib/node-utils/index.ts
+++ b/lib/node-utils/index.ts
@@ -222,44 +222,31 @@ export function isPromiseHandled(nodeIdentifier: TSESTree.Identifier): boolean {
 		nodeIdentifier,
 		true
 	);
+	const callRootExpression =
+		closestCallExpressionNode == null
+			? null
+			: getRootExpression(closestCallExpressionNode);
 
-	const suspiciousNodes = [nodeIdentifier, closestCallExpressionNode].filter(
-		Boolean
+	const suspiciousNodes = [nodeIdentifier, callRootExpression].filter(
+		(node): node is NonNullable<typeof node> => node != null
 	);
 
-	for (const node of suspiciousNodes) {
-		if (!node?.parent) {
-			continue;
-		}
-		if (ASTUtils.isAwaitExpression(node.parent)) {
-			return true;
-		}
-
+	return suspiciousNodes.some((node) => {
+		if (!node.parent) return false;
+		if (ASTUtils.isAwaitExpression(node.parent)) return true;
 		if (
 			isArrowFunctionExpression(node.parent) ||
 			isReturnStatement(node.parent)
-		) {
+		)
 			return true;
-		}
-
-		if (hasClosestExpectResolvesRejects(node.parent)) {
-			return true;
-		}
-
-		if (hasChainedThen(node)) {
-			return true;
-		}
-
-		if (isPromisesArrayResolved(node)) {
-			return true;
-		}
-	}
-
-	return false;
+		if (hasClosestExpectResolvesRejects(node.parent)) return true;
+		if (hasChainedThen(node)) return true;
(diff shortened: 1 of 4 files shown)


Chamion:
I've now shuffled around the statements such that the branch not visited in tests is on the same line as a branch that is visited. The code is worse as a result but hopefully this passes the code coverage check 😞.

Belco90:
Thanks for your contribution @Chamion! I'll try to review this during the week.

Belco90 (approved):
LGTM, great implementation and detailed tests. Thanks for your contribution @Chamion

sjarva:
Thank you so much @Belco90 for reviewing! 🙇‍♀️ 

And huge thanks and congrats to @Chamion for contributing and getting your first PR reviewed and merged!! 🎉 🎈 🥳

PR #895, opened 2024-04-12
fix(await-async-events): false positives for userEvent.setup() returned

### LINKED ISSUE #881: `await-async-events` produces false-positives

Yes

### Plugin version

v6.2.0

### ESLint version

v8.56.0

### Node.js version

v20.10.0

pnpm 7.33.7

### Operating system

macOS Sonoma, version 14.3.1

### Bug description

The `userEvent.setup()` method is treated as if it were returning a `Promise` when wrapped by an arrow function expression.

### Steps to reproduce

While exposing a `getUser` wrapper around `userEvent.setup` for reuse:

```ts
export const getUser = () =>
  userEvent.setup({
    advanceTimers: (msToRun: number) => jest.advanceTimersByTime(msToRun),
  });
```

Usages like:

```ts
const user = getUser();
```

Produce a false-positive error as follows:

```
error  Promise returned from `getUser` wrapper over async event method must be handled  testing-library/await-async-events
```

However, `userEvent.setup` isn‘t an async method, so there isn‘t anything to await.

### Error output/screenshots

_No response_

### ESLint configuration

```json
{
  "extends": ["plugin:testing-library/react"]
}
```

or

```json
{
  "plugins": ["testing-library"],
  "rules": {
    "testing-library/await-async-events": "error"
  }
}
```

### Rule(s) affected

`testing-library/await-async-events`

### Anything else?

_No response_

No

Chamion:
## Checks

## Changes

The circuit-breaker checking for `userEvent.setup()` was executed after `detectEventMethodWrapper` so the circuit-breaking only applied to cases where the expression is not returned. It's intended to be applied generally so this PR moves it to the top of its code block.

## Context

My last contribution was received well so I picked another open issue.

Fixes #881

This is another false-positive where the root problem is defaulting to report unless a reason not to report is found. If new methods are introduced to `userEvent` the rule will assume they return promises. I'd like to change the behaviour to check for an allow-list of method names instead of checking it's not specifically `setup` so future bugs would be false negatives instead of false positives. I didn't do that in this PR because maybe it's the way it is by design.

### DIFF
--- a/lib/rules/await-async-events.ts
+++ b/lib/rules/await-async-events.ts
@@ -122,6 +122,10 @@ export default createTestingLibraryRule<Options, MessageIds>({
 					(isFireEventEnabled && helpers.isFireEventMethod(node)) ||
 					(isUserEventEnabled && helpers.isUserEventMethod(node))
 				) {
+					if (node.name === USER_EVENT_SETUP_FUNCTION_NAME) {
+						return;
+					}
+
 					detectEventMethodWrapper(node);
 
 					const closestCallExpression = findClosestCallExpressionNode(
@@ -130,10 +134,6 @@ export default createTestingLibraryRule<Options, MessageIds>({
 					);
 
 					if (!closestCallExpression?.parent) {
-						return;
-					}
-
-					if (node.name === USER_EVENT_SETUP_FUNCTION_NAME) {
 						return;
 					}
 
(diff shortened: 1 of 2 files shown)


Belco90 (approved):
LGTM, thanks @Chamion!

Belco90:
@all-contributors please add @Chamion for code, test

PR #902, opened 2024-04-27
feat: support ESLint flat config

Belco90:
## Checks

- [X] I have read the [contributing guidelines](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/CONTRIBUTING.md).

## Changes

- Following [the official ESLint guide for making the plugin compatible with flat config](https://eslint.org/docs/latest/extend/plugin-migration-flat-config)

## Context

Closes #853

PR #921, opened 2024-08-04
fix: add plugin name and version metadata for ESLint v9

G-Rath:
## Checks

## Changes

- Add plugin name and version metadata

## Context

Relates to #902 - this metadata is technically not required but preferred for ESLint v9 to be able to work.. better? (iirc its used for cache busting).

I've not added any tests for now as I think it's a small enough change that we can verify it's fine with our eyes :)

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,10 +1,12 @@
-import {
-	name as packageName,
-	version as packageVersion,
-} from '../package.json';
-
 import configs from './configs';
 import rules from './rules';
+
+// we can't natively import package.json as tsc will copy it into dist/
+const {
+	name: packageName,
+	version: packageVersion,
+	// eslint-disable-next-line @typescript-eslint/no-var-requires
+} = require('../package.json') as { name: string; version: string };
 
 export = {
 	meta: {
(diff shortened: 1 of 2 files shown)


Belco90 (dismissed):
Thanks for taking care of this for the current version ❤️

Belco90:
There are some types failing tho, can you take a look at that @G-Rath?

G-Rath:
@Belco90 I don't think the CI failure is related as it's happening on all my PRs

G-Rath:
ah right ok the issue is that now that we're importing from `package.json`, that gets copied by tsc as part of the build, and the package structure changes - tbh it might be easiest to just switch to using babel for compiling 🤔

Belco90:
That's what I was dealing with in the next major (introducing tsup), but don't have time right now to finish it.

G-Rath:
Right, well in the meantime we can use plain `require` so now this should be good to land

G-Rath:
@Belco90 fwiw #923 is also now passing and includes this change, so might be better to just land that one - I'll leave both open though in case you want to stagger them for dedicated changelog entries :)

Belco90:
Closing in favor of #923

PR #922, opened 2024-08-04
chore: enable `eslint-plugin-prettier`

G-Rath:
## Checks

## Changes

- enables `eslint-plugin-prettier` via it's recommended ruleset

## Context

Extending `prettier` enables `eslint-config-prettier` which only disables ESLint rules that conflict with `prettier` - it's the `eslint-plugin-prettier` config that enables _both_ the `prettier/prettier` rule _and_ `eslint-config-prettier`.


Belco90:
This was done on purpose. We don't want to run stylistic rules as part of ESLint.

PR #923, opened 2024-08-04
feat: add support for flat config

### LINKED ISSUE #853: Prepare the plugin for flat config
### Plugin version

n/a

### What problem do you want to solve?

- [Prepare the plugin for flat config](https://eslint.org/docs/latest/extend/plugin-migration-flat-config).

### Your take on the correct solution?

Apply the changes mentioned in the linked migration docs so the plugin is compatible with flat config mechanism

### Anything else?

_No response_

### Do you want to submit a pull request to implement this change?

Yes

G-Rath:
## Checks

## Changes

- Adds plugin name and version metadata (aka #921)
- Adds flat versions of configs, for ESLint v9

## Context

Resolves #902
Resolves #853

This uses a very similar approach to what I'm using for `eslint-plugin-jest` - the types should actually be pretty much correct, but I've just gone straight to casting since things are a bit of a mess right now ecosystem wise, which'll improve once we have everyone on ESLint v9.

Similarly I've not worried about adding any particular tests since the output is pretty stable and easily validated manually (which I have already), and I've based the documentation off what we use in `eslint-plugin-jest` too

### DIFF
--- a/.eslint-doc-generatorrc.js
+++ b/.eslint-doc-generatorrc.js
@@ -3,6 +3,13 @@ const prettierConfig = require('./.prettierrc.js');
 
 /** @type {import('eslint-doc-generator').GenerateOptions} */
 const config = {
+	ignoreConfig: [
+		'flat/angular',
+		'flat/dom',
+		'flat/marko',
+		'flat/react',
+		'flat/vue',
+	],
 	postprocess: (content) =>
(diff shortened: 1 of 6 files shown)


Belco90 (approved):
Thanks for adding flat config support in the current version ❤️

PR #926, opened 2024-08-14
test: run against ESLint v9

G-Rath:
## Checks

## Changes

- introduces a new "flat compat" rule tester that accepts tests written for legacy config, and automatically converts them to flat config when running against ESLint v9
  - this is a class I have used in `eslint-plugin-jest`, `eslint-plugin-jest-dom`, and `eslint-plugin-jest-extended` with great successful
- updates tests to pass when using ESLint v9
- updates CI to run test suite against ESLint v9

## Context

Before saying the plugin officially supports ESLint v9, it should be testing against v9 which in turn requires accounting for v9's defaulting to flat config and the improvements that have been made to the rule tester to help plugin authors catch bugs.

Besides flat config in general, the two notable changes to the rule tester that require changes to this test suite are:
  - it's now an error to have `output` that matches `code`
  - it's now an error to have duplicate cases

There's only a couple of occurrences of that first one and the fix is to just specify `null` instead which means "this rule supports autofix, but for this test case it does not do anything".

The second though requires a lot more work especially because of how much mapping this test suite does meaning addressing them in full would have huge changes to the layout so instead for now I've made all the tests unique typically by adding a comment that mentions the testing framework whose loop they're part of - ideally though a number of these tests should be restructured to pull such cases out of the mapping loops, which is something I'm happy to help with via follow-up and parallel PRs.

~I have also cherry-picked the straightforward removal of a number of duplicates into #927 which if landed first would help reduce the overall diff of this pull request.~

### DIFF
--- a/.github/workflows/verifications.yml
+++ b/.github/workflows/verifications.yml
@@ -63,6 +63,10 @@ jobs:
       - name: Install dependencies
         run: npm install
 
+        # see https://github.com/npm/cli/issues/7349
+      - if: ${{ matrix.eslint == 9 }}
+        run: npm un eslint-plugin-jest eslint-doc-generator
+
       - name: Install ESLint v${{ matrix.eslint }}
         run: npm install --no-save --force eslint@${{ matrix.eslint }}
 
--- a/.github/workflows/verifications.yml
+++ b/.github/workflows/verifications.yml
@@ -31,6 +31,7 @@ jobs:
   tests:
     name: Tests (Node v${{ matrix.node }} - ESLint v${{ matrix.eslint }})
     runs-on: ubuntu-latest
+    timeout-minutes: 3
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/verifications.yml
+++ b/.github/workflows/verifications.yml
@@ -66,7 +66,7 @@ jobs:
 
         # see https://github.com/npm/cli/issues/7349
       - if: ${{ matrix.eslint == 9 }}
-        run: npm un eslint-plugin-jest eslint-doc-generator
+        run: npm un @typescript-eslint/eslint-plugin eslint-plugin-jest eslint-doc-generator
 
       - name: Install ESLint v${{ matrix.eslint }}
         run: npm install --no-save --force eslint@${{ matrix.eslint }}
--- a/.github/workflows/verifications.yml
+++ b/.github/workflows/verifications.yml
@@ -35,7 +35,21 @@ jobs:
       fail-fast: false
       matrix:
         node: [12.22.0, 12, 14.17.0, 14, 16, 17, 18, 19, 20, 22]
-        eslint: [7.5, 7, 8]
+        eslint: [7.5, 7, 8, 9]
+        exclude:
+          # eslint@9 doesn't support < Node v18
+          - node: 17
+            eslint: 9
+          - node: 16
+            eslint: 9
+          - node: 14
+            eslint: 9
+          - node: 14.17.0
+            eslint: 9
+          - node: 12
+            eslint: 9
+          - node: 12.22.0
+            eslint: 9
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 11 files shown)


.github/workflows/verifications.yml
```diff
@@ -47,6 +62,12 @@ jobs:
 
       - name: Install dependencies
         uses: bahmutov/npm-install@v1
```
Belco90, line 64: We have merged another PR removing  `npm-install` usage, can you adapt your changes to this?

.github/workflows/verifications.yml
```diff
@@ -30,12 +30,27 @@ jobs:
   tests:
     name: Tests (Node v${{ matrix.node }} - ESLint v${{ matrix.eslint }})
     runs-on: ubuntu-latest
+    timeout-minutes: 3
```
Belco90, line 33: This is great, thanks!

Belco90 (commented):
The changes LGTM, but I'll give it another review after being rebased.

tests/lib/rules/await-async-events.test.ts
```diff
@@ -154,7 +154,7 @@ ruleTester.run(RULE_NAME, rule, {
 					'testing-library/utils-module': 'test-utils',
 				},
 				code: `
-        import { fireEvent } from 'test-utils'
+        import { fireEvent } from 'test-utils' // implicitly ${testingFramework}
```
MichaelDeBoey, line 157:
```suggestion
        import { fireEvent } from 'test-utils' // implicitly using ${testingFramework}
```

tests/lib/rules/await-async-events.test.ts
```diff
@@ -142,7 +142,7 @@ ruleTester.run(RULE_NAME, rule, {
 					'testing-library/utils-module': 'test-utils',
 				},
 				code: `
-        import { fireEvent } from 'somewhere-else'
+        import { fireEvent } from 'somewhere-else' // not ${testingFramework}
```
MichaelDeBoey, line 145:
```suggestion
        import { fireEvent } from 'somewhere-else' // not using ${testingFramework}
```

tests/lib/rules/await-async-utils.test.ts
```diff
@@ -561,6 +562,7 @@ ruleTester.run(RULE_NAME, rule, {
 			(asyncUtil) =>
 				({
 					code: `
+				// using ${testingFramework} implicitly
```
MichaelDeBoey, line 565:
```suggestion
				// implicitly using ${testingFramework}
```

tests/lib/rules/await-async-utils.test.ts
```diff
@@ -590,6 +592,7 @@ ruleTester.run(RULE_NAME, rule, {
...
```
MichaelDeBoey, line 595:
```suggestion
				// implicitly using ${testingFramework}
```

tests/lib/rules/await-async-utils.test.ts
```diff
@@ -620,6 +623,7 @@ ruleTester.run(RULE_NAME, rule, {
...
```
MichaelDeBoey, line 626:
```suggestion
				// implicitly using ${testingFramework}
```

tests/lib/rules/await-async-utils.test.ts
```diff
@@ -649,6 +653,7 @@ ruleTester.run(RULE_NAME, rule, {
...
```
MichaelDeBoey, line 656:
```suggestion
				// implicitly using ${testingFramework}
```

tests/lib/rules/await-async-utils.test.ts
```diff
@@ -678,6 +683,7 @@ ruleTester.run(RULE_NAME, rule, {
...
```
MichaelDeBoey, line 686:
```suggestion
				// implicitly using ${testingFramework}
```

tests/lib/rules/await-async-utils.test.ts
```diff
@@ -706,6 +712,7 @@ ruleTester.run(RULE_NAME, rule, {
...
```
MichaelDeBoey, line 715:
```suggestion
				// implicitly using ${testingFramework}
```

MichaelDeBoey (approved):
Some minor tweaks to the comments to make them consistent all over the place, but nothing big

MichaelDeBoey (approved):

Belco90 (approved):
Great job 👏

PR #925, opened 2024-08-14
fix: replace use of deprecated methods

G-Rath:
## Checks

- [x] I have read the [contributing guidelines](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/CONTRIBUTING.md).

## Changes

- introduced compat utilities that enable support for ESLint v9
- replaced usages of `context` functions that were removed in v9 in favor of compat utils

## Context

Resolves #924
Resolves #899

PR #928, opened 2024-08-15
feat: support ESLint v9

G-Rath:
## Checks

## Changes

- adds ESLint v9 as an allowed peer dependency

## Context

v9 can be officially supported given that #926 has the plugin running against v9 without issue


niklaas:
Is this the proper PR to subscribe to for getting updated about making this plugin compatible working with eslint v9? I'm asking because it's been in draft state since August. Anything I could potentially help with to get it into review state?

G-Rath:
@niklaas this is good to go but currently a draft as it's branched off of #926 (which proves the plugin works with ESLint v9) and in turn that's branched off of #925 (which is the runtime fix required for the plugin to work with v9) - I've done them as dedicated PRs as I feel it's nice to have them in their own commits and released independently in case something comes up along the line.

While @Belco90 doesn't have to land my PRs, I suspect that he'll be happy with them so I think it's the one to subscribe to - alternatively you could subscribe to releases for this whole codebase since it's most likely the next release will include v9 support:

image
image

niklaas:
Thanks a lot! Appreciated the update. 🙇🏼

Issue #855, opened 2023-12-12
Upgrade core dependencies


Belco90:
I'm creating this issue to keep discussing the upgrade of some core dependencies of this plugin, originated in #852 (@lesha1201).

### typescript-eslint
We could support both v5 and v6, actually. I think it will make things more complicated, so let me think about it. There are other things more critical to be updated now.

### Prettier
I already tried to upgrade to v3, but had to revert it because of breaking changes introduced. [You can find more details in the revert PR](https://github.com/testing-library/eslint-plugin-testing-library/pull/826).

### TypeScript
Definitely interested in upgrading this one, and it's actually needed to upgrade prettier to v3 as described in the previous revert PR.

### Jest
Not sure why this was updated in #852, but we need to upgrade it anyway. ~Shouldn't be blocked by anything~ This is blocked by Node.js v12 which is not supported by Jest v29.

So our priorities should be:

- [X] Upgrade TS to v5 --> #858
- [x] Upgrade Prettier to v3 --> #932
- [x] Upgrade typescript-eslint to v7 (breaking change)
- [x] Update Node.js engine requirement to `^18.18.0 || >=20.0.0` (breaking change) --> #898 

### Other improvements
When Node.js compatibility for non-supported versions is dropped, we can apply more improvements:

- [x] ~Upgrade Jest to v29~ Migrate test runner to vitest --> #1102
- [x] Upgrade TypeScript to the last v5.X (enabling pending things from #858) --> #1132
- [x] Replace npm with pnpm as the package manager --> #966
- [x] Use tsup for bundling in both CJS and ESM formats --> #900

Belco90:
Just thinking out loud: another migration update plan for this plugin could be releasing a new major that upgrades typescript-eslint to v6 + drops Node.js support for the same versions as the future ESLint v9. Then when ESLint v9 gets released we don't release another major but a minor supporting both ESLint v8 and v9.

lesha1201:
@Belco90 Do you want to upgrade TS or I can open PR? I saw your notes about updating `tsconfig` in this PR https://github.com/testing-library/eslint-plugin-testing-library/pull/826 but not sure what exactly you want to update. I can open PR with these changes - https://github.com/testing-library/eslint-plugin-testing-library/pull/852/commits/20c87e82f8d66d6a22a38773379d66676a8edadd and you can either ask me to add some changes to `tsconfig` or push them directly.

Belco90:
[...]

PR #932, opened 2024-09-27
chore(deps-dev): upgrade Prettier to v3

G-Rath:
## Checks

## Changes

- Upgrades Prettier to v3
- Removes unused `eslint-plugin-prettier`
- Rewrites `tools/generate-configs` to be async

## Context

Prettier v3 removes its sync API, so to upgrade we now have to use promises

Relates to #826 and #855


.prettierrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+	trailingComma: 'es5',
```
G-Rath, line 2: note: in v3 the default changes to "all"

I've locked this to the v2 default to reduce the diff in this pull request since as an option it will be very easy to do a follow-up PR switching this default if desired

MichaelDeBoey (dismissed):

Belco90:
@G-Rath can you resolve the conflicts with `main`? So I merge this one before other changes.

PR #934, opened 2024-10-08
fix(no-manual-cleanup): properly check all `@testing-library` imports

G-Rath:
## Checks

## Changes

- Changes `no-manual-cleanup` so it checks every import rather than just the first one

## Context

Currently the rule assumes the first `@testing-library` import will be one with a `cleanup`, which isn't true


MichaelDeBoey (approved):

Belco90 (approved):
Amazing, thanks!

PR #935, opened 2024-10-08
chore: add editorconfig file

G-Rath:
## Checks

## Changes

- Adds a `.editorconfig` file to tell tools that tabs should be used instead of spaces, along with a few other config options

## Context

Most editors these days understand `.editorconfig` out of the box, and in particular this will help ensure that tabs are used by default over spaces when writing code including in situations that Prettier won't catch such as in string literals

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = tab


MichaelDeBoey (commented):
We have Prettier installed, so I'd rather have people use their IDE's on save functionality tbh

G-Rath:
As I said in the context section, Prettier doesn't help here because it doesn't apply in string literals - this isn't about automatic formatting, but what IDEs use for indentation by default i.e. when you start a newline from an already intended line, editors typically keep you on the same level of indenting, which they do by inserting either spaces or tabs

MichaelDeBoey (approved):

Belco90 (approved):

Belco90:
@all-contributors please add @G-Rath for code, maintenance

PR #939, opened 2024-10-16
fix: generate provenance statements on release

MichaelDeBoey:
This will also add a new Provenance section + a verified badge next to the version (in the sidebar + versions tab) to the NPM page

More info can be found at https://docs.npmjs.com/generating-provenance-statements

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,12 @@ concurrency:
   group: release
   cancel-in-progress: false
 
+permissions:
+  contents: write # to be able to publish a GitHub release
+  id-token: write # to enable use of OIDC for npm provenance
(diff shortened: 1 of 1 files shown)


Belco90 (approved):
TIL about provenance, thanks!

Belco90:
@MichaelDeBoey I don't think this worked as expected? v6.3.2 was removed or hidden from npm registry somehow, [so the latest is still v6.3.1](https://www.npmjs.com/package/eslint-plugin-testing-library). Perhaps because its content is the same as the previous one?

Belco90:
Interesting... it's back again in npm registry.

MichaelDeBoey:
@antfu Maybe you have an idea what went wrong here? 🤔

PR #940, opened 2024-10-16
fix: update dependencies

MichaelDeBoey:


package.json
```diff
@@ -51,42 +51,42 @@
...
-		"@typescript-eslint/eslint-plugin": "^5.58.0",
-		"@typescript-eslint/parser": "^5.58.0",
-		"del-cli": "^5.0.0",
-		"eslint": "^8.38.0",
-		"eslint-config-prettier": "^8.8.0",
-		"eslint-doc-generator": "^1.4.3",
-		"eslint-import-resolver-typescript": "^3.6.1",
-		"eslint-plugin-import": "^2.29.0",
-		"eslint-plugin-jest": "^27.2.1",
+		"@types/node": "^16.18.113",
```
Belco90, line 65: This one should actually be v20 since we changed the version in `nvmrc`. Could you change that here?

package.json
```diff
@@ -51,42 +51,42 @@
...
+		"eslint-remote-tester": "^3.0.1",
 		"eslint-remote-tester-repositories": "^1.0.1",
 		"husky": "^8.0.3",
 		"is-ci": "^3.0.1",
 		"jest": "^28.1.3",
-		"lint-staged": "^13.2.1",
-		"npm-run-all": "^4.1.5",
-		"prettier": "2.8.7",
+		"lint-staged": "^13.3.0",
+		"npm-run-all2": "^5.0.2",
```
Belco90, line 85: What is this `npm-run-all2`?

Belco90 (changes requested):

package.json
MichaelDeBoey: It's the maintained version of `npm-run-all`

Belco90: Oh, interesting. Hopefully, after releasing the new major and upgrading all core dependencies, I'll be able to migrate to pnpm, which provides utils to run several scripts in a similar way.

Belco90:
We are having issues with Node v12. Perhaps is better just to wait for the next plugin major, where we drop support for outdated Node versions.

MichaelDeBoey:
@Belco90 Node 12 seems to be fixed now

Belco90 (approved):

Belco90 (approved):

PR #956, opened 2024-11-01
add `ByA11yState` to `ALL_QUERIES_METHODS` const

virenpepper:
## Checks

## Changes

- add accessibility queries to `ALL_QUERIES_METHODS` so we can successfully target those queries while pattern matching

## Context
- `prefer-screen-queries` was failing for destructured a11y queries
- this fixes that

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -43,6 +43,7 @@ const ALL_QUERIES_METHODS = [
 	'ByDisplayValue',
 	'ByRole',
 	'ByTestId',
+	'ByA11yState',
 ];
 
 const SYNC_QUERIES_COMBINATIONS = combineQueries(


virenpepper:
> Hi @virenpepper! Could you give us more context on this query? Where is it coming from?

`findByA11yState` and `queryAllByA11yState` are part of the React Native Testing library [1] so when we ran this codemod [2] it didn't catch tests that looked like this ~

```tsx
describe('test', () => {
      it('prefer-screen-queries does not throw an error here', async () => {
        const { findByA11yState } = render(<TestComponent/>);
       expect(queryAllByA11yState({ disabled: true }).length).toBe(1);
        const test = await findByA11yState({
          selected: true,
        });
      });
    });
```

but a couple things to callout,
- it seems like this library **does not support** react-native-testing-library out of the gate, and
- `*ByA11yState` is deprecated

so the tl;dr is maybe this PR can be closed? we are also facing issues where the `prefer-screen-queries` rule also fails to catch destructured `rerender` ie

```tsx 
    const { rerender } = render(<TestComponent />);
```

unsure how to proceed here, as of now we hacked the local library to catch these gaps

[1] https://callstack.github.io/react-native-testing-library/docs/api/queries#by-accessibility-state
[2] https://github.com/atsikov/eslint-jest-testing-library-codemod

MichaelDeBoey:
@virenpepper As you correctly stated: this plugin (currently) doesn't support `@testing-library/react-native`.

I'll let @Belco90 decide what to do with this, but imo we should not add this and add full support in 1 go instead

Belco90 (commented):
Hey @virenpepper! Sorry it took me so long to react to this.

I've seen there are several PRs for adding support for React Native in this plugin (this PR, #958, others in the past). Instead of having separated PRs for adding support here and there, we should unify the changes to support RN entirely in one go, similar to what was done in #973.

So we need to create an umbrella issue to gather what's missing in this plugin to fully support RN, then move the current work there, including the RN preset.

Belco90:
Closing in favor of adding React Native support properly. Added to https://github.com/orgs/testing-library/projects/20/views/1 to track necessary changes.

PR #958, opened 2024-11-07
feat(prefer-user-event): add support for React Native events

### LINKED ISSUE #957: Support React Native events in prefer-user-event

prefer-user-event

More warnings

Add more fireEvent -> userEvent mappings.

React Native Testing Library has its own fireEvent function and the same idea of proposing to use userEvent if possible.

According to the [docs](https://callstack.github.io/react-native-testing-library/docs/api/events/fire-event), it's recommended to use:
* User Event `press` instead of `fireEvent.press`
* User Event `type` instead of `fireEvent.changeText`
* User Event `scrollTo` instead of `fireEvent.scroll`

### Example code

```
fireEvent.press(screen.getByText('Press me'))
```

It doesn't suggest any changes to `fireEvent` calls for React Native projects.

It'll suggest to use User Event as per the docs.

### Anything else?

_No response_

Yes

ddrozdov:
## Checks

## Changes

- add support for React Native events

## Context

Fixes #957

### DIFF
--- a/docs/rules/prefer-user-event.md
+++ b/docs/rules/prefer-user-event.md
@@ -118,6 +118,7 @@ The following table lists all the possible equivalents from the low-level API `f
 | ---------------- | ----------------------------------------------------------------------------------------------------------- |
 | `click`          | <ul><li>`click`</li><li>`type`</li><li>`selectOptions`</li><li>`deselectOptions`</li></ul>                  |
 | `change`         | <ul><li>`upload`</li><li>`type`</li><li>`clear`</li><li>`selectOptions`</li><li>`deselectOptions`</li></ul> |
+| `changeText`     | <ul><li>`type`</li></ul>                                                                                    |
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (commented):
@ddrozdov This plugin (currently) doesn't support `@testing-library/react-native`.

I'll let @Belco90 decide what to do with this, but imo we should not add this and add full support in 1 go instead

Belco90:
Hey @ddrozdov! Sorry it took me so long to react to this.

I've seen there are several PRs for adding support for React Native in this plugin (this PR, #956, others in the past). Instead of having separated PRs for adding support here and there, we should unify the changes to support RN entirely in one go, similar to what was done in #973.

So we need to create an umbrella issue to gather what's missing in this plugin to fully support RN, then move the current work there, including the RN preset.

Belco90:
Closing in favor of adding React Native support properly. Added to https://github.com/orgs/testing-library/projects/20/views/1 to track necessary changes.

PR #963, opened 2024-11-19
feat: add TypeScript types

G-Rath:
## Checks

## Changes

- adds types so that usage with `eslint.config.js` in TypeScript codebases don't result in errors

## Context

Now that plugins are imported "naturally" they can be type-checked by TypeScript so it's useful to ship types - for now I've just manually typed the plugin since primarily the only useful thing to have typed is the configs, as things like rules and settings are still not implicitly inferred and trying to automate this would be overkill.

These types match those that we've got in `eslint-plugin-jest`, `eslint-plugin-jest-dom`, and other places

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,25 @@
+import type { Linter, Rule } from 'eslint';
+
+declare const plugin: {
+	meta: {
+		name: string;
+		version: string;
+	};
+	configs: {
+		angular: Linter.LegacyConfig;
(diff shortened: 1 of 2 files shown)


Belco90 (approved):

Belco90:
Thanks! I'll autogenerate this with tsup when I finish my pending task for setting up a proper bundle to export CJS and ESM formats.

PR #964, opened 2024-11-20
feat: cleanup main types

MichaelDeBoey:
Follow-up of #963

CC/ @G-Rath

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,25 +1,15 @@
-import type { Linter, Rule } from 'eslint';
+import type { TSESLint } from '@typescript-eslint/utils';
+
+import { SupportedTestingFramework } from './lib/utils';
 
 declare const plugin: {
 	meta: {
 		name: string;
 		version: string;
 	};
-	configs: {
-		angular: Linter.LegacyConfig;
-		dom: Linter.LegacyConfig;
-		marko: Linter.LegacyConfig;
-		react: Linter.LegacyConfig;
-		vue: Linter.LegacyConfig;
-		'flat/angular': Linter.FlatConfig;
-		'flat/dom': Linter.FlatConfig;
-		'flat/marko': Linter.FlatConfig;
-		'flat/react': Linter.FlatConfig;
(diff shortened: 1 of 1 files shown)


G-Rath:
I don't really see this as strictly better - the types as they are pretty straightforward and maintainable as they are imo, and I in particular like having the configs written out individually to make it clear what is available without having to mentally unroll some type logic.

Unlike your PRs to the other plugins, this one does has a slight advantage due to the number of configs, but given I don't expect this to need changing pretty much ever (aside from when the `flat` configs go away which'll half the number of properties), I still wouldn't say it's worth landing.

At the very least, this is not a "feat" or a "fix"

MichaelDeBoey:
@G-Rath We're also using TSESLint's types here, which is a dependency we already have

If we use ESLint's dependency, we would need to add `@types/eslint` to the `dependencies` in `package.json` to work correctly

G-Rath:
@MichaelDeBoey you'll already need `@types/eslint` as a dependency if you're type-checking `eslint.config.js`, and ESLint v9 ships with its own types that are compatible with these ones, so that shouldn't be an issue

MichaelDeBoey:
@G-Rath with these changes, you don't need `@types/eslint` anymore when having ESLint v8

G-Rath:
@MichaelDeBoey do you have an actual example of a project where you are getting an error with these types because `@types/eslint` isn't present?

Because I'm willing to bet most people will have `@types/eslint` available if they're using v8 since to have a properly typed `eslint.config.js` you should be typing your export as a flat config (so you either need to source that type from `@types/eslint` or `@typescript-eslint`), and most plugins I've seen exporting types are using types from `eslint` rather than `@typescript-eslint`.

So the only situation I can think of where this'd come up is if you're running _just_ this plugin with ESLint v8 which seems kind of weird - why would you go to all the trouble of setting up ESLint just to apply linting rules for your testing code?

Meanwhile, everyone is busy upgrading to ESLint v9 which as mentioned before is now typed - so I feel the audience for this is very niche.

Issue #80, opened 2020-02-07
Add new shareable configurations for libraries


thomlom:
Currently, we are supporting Angular, React and Vue for shareable configurations. The configuration contains the following rules:

- await-async-query
- await-async-utils
- no-await-async-query
- no-debug
- no-dom-import

It would be nice to add shareable configurations for other libraries such as:

- [Native Testing Library](https://testing-library.com/docs/native-testing-library/intro)
- [Preact Testing Library](https://testing-library.com/docs/preact-testing-library/intro)
- [Svelte Testing Library](https://testing-library.com/docs/svelte-testing-library/intro)
- [Marko Testing Library](https://testing-library.com/docs/marko-testing-library/intro)

We would need to check that each rule can be applied to these packages.

Belco90:
Indeed! Would be nice to include all those presets, and even check if there are special cases for them as async fire event for vue so we can add additional rules to those presets in a next step.

I'll try to find more info about what of our current rules could/should be applied for each and share it here for further discussion.

Belco90:
I did a quick research for Native Testing Library. Even if it's implemented in a completely different way under the hood (as it's not looking for elements in the DOM of course) the main core of the API remains the same. These are the main differences I've found:

- Additional [NativeTestInstance](https://www.native-testing-library.com/docs/api-test-instance) with additional properties. This would need additional rules maybe, but it doesn't affect the current ones.
- Additional [ByHintText](https://www.native-testing-library.com/docs/api-queries#byhinttext) query. Just including this one in the list of queries would be enough? It would be enabled for other presets but I don't know if that's a problem.
- `ByRole` query doesn't exist. We could _disable_ checks for this query when enabling native testing library preset, but I don't think it's worth the effort?
-  Methods from `fireEvent` are different. I don't think this affects the plugin as we don't check any specific `fireEvent` method.

So apart from adding the additional query I don't think we would have to do anything else for that one.

thomlom:
Thanks for the details! There is also a difference for Marko Testing Library for `no-dom-import` as the package name is `@marko/testing-library` and not `@testing-library/marko`.

Belco90:
Maybe we should split this in a project to be able to implement and review each library separated and then merge everything together.

SimenB:
One not mentioned in the OP is [`cypress`](https://testing-library.com/cypress)

Belco90:
[...]

PR #973, opened 2024-12-03
feat: add config for Svelte

nostrorom:
## Checks

## Changes

- Adds recommended config for `svelte`
- Updates the `Linter.FlatConfig` type (deprecated)

## Context

Adds to #80


Belco90 (approved):
Hey @nostrorom! Thanks for your contribution, this LGTM.

Belco90:
@all-contributors please add @nostrorom for code

index.d.ts
```diff
@@ -10,12 +10,14 @@ declare const plugin: {
...
-		'flat/dom': Linter.FlatConfig;
-		'flat/marko': Linter.FlatConfig;
-		'flat/react': Linter.FlatConfig;
-		'flat/vue': Linter.FlatConfig;
+		'flat/angular': Linter.Config<Linter.RulesRecord>;
+		'flat/dom': Linter.Config<Linter.RulesRecord>;
+		'flat/marko': Linter.Config<Linter.RulesRecord>;
+		'flat/react': Linter.Config<Linter.RulesRecord>;
+		'flat/svelte': Linter.Config<Linter.RulesRecord>;
+		'flat/vue': Linter.Config<Linter.RulesRecord>;
```
G-Rath, line 20: note that these types should not have been changed - while `Linter.FlatConfig` is marked as deprecated in v9, it's the correct type to be using for v8 since `Linter.Config` there resolves to the legacy configuration.

For plugins like this which support both v8 and v9, we should be using the explicit `Linter.LegacyConfig` and `Linter.FlatConfig` types to ensure users of both receive the right types.

I have opened #976 to revert this change

nostrorom: Indeed I didn't realise these were necessary for v8 compatibility

Thanks for the fix !

PR #984, opened 2025-01-27
[react] Disable `testing-library/no-await-sync-events` by default

eps1lon:
## Changes

`@testing-library/react` will require `await act()`. At that point the rule is useless. Kept for backwards compatibility.

## Context

https://github.com/testing-library/react-testing-library/pull/1214/


Belco90:
Hi @eps1lon! Thanks for this PR and sharing context, I wasn't aware of such change.

Belco90 (approved):

eps1lon:
It hasn't landed yet. This is just preparation.

PR #1006, opened 2025-05-11
feat: add `no-test-id-queries` rule

### LINKED ISSUE #279: New rule suggestion: no-test-id-queries
Hi! I've been wondering if there was ever a discussion about implementing a rule to warn about accessible queries use. Since I don't think it's really necessary to throw an error or do an autofix, a warning should be just fine. So we could get a warning to possibly refactor this

```javascript
const loginButton = screen.getByTestId('login-button');
```

into this

```javascript
const loginButton = screen.getByRole('button', { name: /login/i });
```

What do you think? :)

StyleShit:
Closes #279


docs/rules/no-test-id-queries.md
```diff
@@ -0,0 +1,31 @@
...
+Examples of **correct** code for this rule:
+
+```js
+const button = screen.getByRole('button');
+const input = screen.getByRole('textbox');
+```
+
+## Further Reading
+
+- [about `getByTestId`](https://testing-library.com/docs/queries/bytestid)
```
Belco90, line 29: suggestion (blocking): Please add a link to the "[Not querying by text](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text)" section from "Common mistakes with React Testing Library" post.

lib/rules/no-test-id-queries.ts
```diff
@@ -0,0 +1,46 @@
+import { TSESTree } from '@typescript-eslint/utils';
+
+import { createTestingLibraryRule } from '../create-testing-library-rule';
+
+export const RULE_NAME = 'no-test-id-queries';
+export type MessageIds = 'noTestIdQueries';
+type Options = [];
+
+const QUERIES_REGEX = /^(get|query|getAll|queryAll|find|findAll)ByTestId$/;
```
Belco90, line 9: suggestion (blocking): Please build this regex using the `ALL_QUERIES_VARIANTS` from the utils file.

tests/lib/rules/no-test-id-queries.test.ts
```diff
@@ -0,0 +1,78 @@
...
+	'getByTestId',
+	'queryByTestId',
+	'getAllByTestId',
+	'queryAllByTestId',
+	'findByTestId',
+	'findAllByTestId',
+];
+
+ruleTester.run(RULE_NAME, rule, {
+	valid: [
```
Belco90, line 24: suggestion (blocking): Please add another valid test that includes calls to methods containing `*TestId` at the end but not matching the regex reported by the rule (e.g. `caculateTestId`, `getTestId`, `searchByTestId`)

Belco90 (changes requested):

docs/rules/no-test-id-queries.md
StyleShit: Nice, done

lib/rules/no-test-id-queries.ts
StyleShit: Good idea, done

tests/lib/rules/no-test-id-queries.test.ts
StyleShit: Done

lib/rules/no-test-id-queries.ts
```diff
@@ -0,0 +1,47 @@
...
+			description: 'Ensure no `data-testid` queries are used',
+			recommendedConfig: {
+				dom: 'error',
+				angular: 'error',
+				react: 'error',
+				vue: 'error',
+				svelte: 'error',
+				marko: 'error',
+			},
+		},
```
Belco90, line 26: suggestion (blocking): Please, remove this rule from all the recommended configs.

Based on [the package Semantic Versioning Policy](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/semantic-versioning-policy.md), adding more rules to existing shareable configurations is a breaking change. So we will release the new rule as a minor, then in the next major will be added to all configs.

Belco90 (changes requested):

lib/rules/no-test-id-queries.ts
StyleShit: oops, sorry, it was copy-pasted

Belco90: No worries, it happens all the time! Would be a good idea to set up an internal tool for creating new rules.

Belco90:
@all-contributors please add @StyleShit for code, test, docs

PR #1013, opened 2025-05-28
feat(prefer-find-by): add fixer for waitFor wrapping findBy queries

### LINKED ISSUE #910: Disallow using findBy* inside a waitFor block
### Name for new rule

no-findby-in-waitfor

This new rule should prevent the usage of a `findBy`* method inside a `waitFor`.

### Testing Library feature

`waitFor` and `findBy`

all

_No response_

### Code examples

**Incorrect usage**
```ts
await waitFor(() => {
  const button = await findByRole("button", { name: "Submit" });
  expect(button).toBeInTheDocument()
};
```

**Correct usage**
```ts
  const button = await findByRole("button", { name: "Submit" });
  expect(button).toBeInTheDocument()
```

### Anything else?

_No response_

Yes

y-hsgw:
Hi 👋 This is my first contribution — thank you for the great work on this project!

## Checks

## Changes

- Introduced auto-fix support to the `prefer-find-by` rule.
- Fixes cases where `findBy*` queries are unnecessarily wrapped inside `waitFor`.
 
## Context

Fixes #910

### DIFF
--- a/lib/rules/prefer-find-by.ts
+++ b/lib/rules/prefer-find-by.ts
@@ -2,12 +2,15 @@ import { TSESTree, ASTUtils, TSESLint } from '@typescript-eslint/utils';
 
 import { createTestingLibraryRule } from '../create-testing-library-rule';
 import {
+	getDeepestIdentifierNode,
 	isArrowFunctionExpression,
+	isBlockStatement,
 	isCallExpression,
 	isMemberExpression,
 	isObjectExpression,
 	isObjectPattern,
 	isProperty,
+	isVariableDeclaration,
 } from '../node-utils';
 import { getScope, getSourceCode } from '../utils';
 
@@ -19,6 +22,10 @@ export function getFindByQueryVariant(
 	queryMethod: string
 ): 'findAllBy' | 'findBy' {
 	return queryMethod.includes('All') ? 'findAllBy' : 'findBy';
+}
+
+function isFindByQuery(name: string): boolean {
+	return /^find(All)?By/.test(name);
 }
 
 function findRenderDefinitionDeclaration(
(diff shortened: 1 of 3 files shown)


lib/rules/prefer-find-by.ts
```diff
@@ -329,20 +336,82 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+							!ASTUtils.isAwaitExpression(declaration.init) ||
+							!isCallExpression(declaration.init.argument)
+						) {
+							return false;
+						}
+
+						const { callee } = declaration.init.argument;
+
+						const name = getDeepestIdentifierNode(callee)?.name;
+						return name ? isFindByQuery(name) : false;
```
Belco90, line 372: question(blocking): In the `helpers` received as the 3rd param of the `create`, you have a helper called [`isFindQueryVariant`](https://github.com/testing-library/eslint-plugin-testing-library/blob/e4c0daaabbee97e560f79c5c4a0a4cb67d3d35bd/lib/create-testing-library-rule/detect-testing-library-utils.ts#L397C9-L397C27). Why not using that one instead?

lib/rules/prefer-find-by.ts
```diff
@@ -329,20 +336,82 @@ export default createTestingLibraryRule<Options, MessageIds>({
 		}
 
 		return {
-			'AwaitExpression > CallExpression'(node: TSESTree.CallExpression) {
+			'AwaitExpression > CallExpression'(
+				node: TSESTree.CallExpression & { parent: TSESTree.AwaitExpression }
```
Belco90, line 340: praise: Good trick!

lib/rules/prefer-find-by.ts
```diff
@@ -329,20 +336,82 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+					const init = ASTUtils.isAwaitExpression(findByDeclarator?.init)
+						? findByDeclarator.init?.argument
+						: null;
+
+					if (!isCallExpression(init)) {
+						return;
+					}
+					const queryIdentifier = getDeepestIdentifierNode(init.callee);
+
+					if (!queryIdentifier || !helpers.isAsyncQuery(queryIdentifier)) {
```
Belco90, line 384: suggestion(blocking): Could you add some comments before each `if` to clarify the intentions of the block? Similar to the previous comment you found that said `// ensure the only argument is an arrow function expression`

Belco90 (changes requested):

Belco90 (changes requested):
Hi @y-hsgw! Thanks for your contribution 🙏.

I've left a couple of comments. You also need to add more tests to cover extra scenarios (codecov left you some hints about what's not covered).

lib/rules/prefer-find-by.ts
y-hsgw: Added in [64c976f](https://github.com/testing-library/eslint-plugin-testing-library/pull/1013/commits/64c976f1025ab794653a8e1185c6e2b9e4321603).

y-hsgw:
@Belco90 
Thanks for the comments! I’ve made the changes and added some tests — appreciate it if you could take another look 🙌

y-hsgw:
Added in [e68646b](https://github.com/testing-library/eslint-plugin-testing-library/pull/1013/commits/e68646bd9dc45c557d9ca97ccd052c43f9ba63cb)!📕

Belco90:
@all-contributors please add @y-hsgw for code and test

Issue #1024, opened 2025-06-09
`no-node-access` rule output false-positive warning after 7.5.0 release


Simek:
### Have you read the Troubleshooting section?

Yes

### Plugin version

7.5.0

### ESLint version

9.28.0

### Node.js version

22.15.0

### Bug description

When upgrading the `eslint-plugin-testing-library` to the latest release today I have found out few new ESLint `no-node-access` rule warnings which seems to be a false-positives.

### Steps to reproduce

```jsx
import { render, screen } from '@testing-library/react';
import userEvent from '@testing-library/user-event';

it('outputs false-positive warnings', async () => {
  render(<button>Test</button>);

  const user = userEvent.setup();
  await user.click(screen.getByText('Test'));
});
```

### Error output/screenshots

![Image](https://github.com/user-attachments/assets/c789257a-7a6e-44ec-963a-4ee714515d86)

### ESLint configuration

ESLint flat config chunk related to Testing Library:

```js
  {
    files: ['**/*-test.[jt]s?(x)'],
    ...testingLibrary.configs['flat/react'],
  }
```

### Rule(s) affected

no-node-access

### Do you want to submit a pull request to fix this bug?

No

Belco90:
@Simek thanks for reporting.

@y-hsgw do you think you can handle this one? I think just adding `"user"` to `EVENTS_SIMULATORS` should be enough.

Simek:
What if I name the constant differently? I think it's a bit too much to expect that everyone will call it the same. 

Can we instead validate the typing on which the event trigger is called? In this case it would be `UserEvent` coming from `@testing-library/user-event` package.

G-Rath:
> Can we instead validate the typing on which the event trigger is called?

That would make TypeScript a requirement for this to be usable, which not everyone uses (though an enhanced version of this rule for TypeScript-based projects would be a cool addition)

You could improve accuracy by using logic similar to what we do in [`eslint-plugin-jest` with our `parseJestFnCall` utility](
https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/utils/parseJestFnCall.ts), which reasonably accurately resolves Jest functions in vanilla JS contexts - it works by taking a call expression and attempts to resolve it to where it comes "into" the file.

i.e. 

```
import { test as somethingCompletelyDifferent } from '@jest/globals`;

const oneAlias = somethingCompletelyDifferent;
const twoAlias = oneAlias;
[...]

PR #1025, opened 2025-06-10
fix(no-node-access): exclude `user` to avoid false positives

y-hsgw:
I have applied the temporary workaround that was proposed in #1024.
Sorry for the inconvenience, but I would appreciate your review.🙇‍♂️

## Checks

## Changes

- Temporarily exclude the variable name 'user' from the no-node-access rule to avoid false positive errors when using `const user = userEvent.setup()`.

## Context

#1024 （partially）

### DIFF
--- a/lib/rules/no-node-access.ts
+++ b/lib/rules/no-node-access.ts
@@ -73,7 +73,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
 				ALL_PROHIBITED_MEMBERS.some(
 					(allReturningNode) => allReturningNode === propertyName
 				) &&
-				!EVENTS_SIMULATORS.some((simulator) => simulator === objectName)
+				![
+					...EVENTS_SIMULATORS,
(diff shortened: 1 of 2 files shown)


lib/rules/no-node-access.ts
```diff
@@ -73,7 +73,12 @@ export default createTestingLibraryRule<Options, MessageIds>({
 				ALL_PROHIBITED_MEMBERS.some(
 					(allReturningNode) => allReturningNode === propertyName
 				) &&
-				!EVENTS_SIMULATORS.some((simulator) => simulator === objectName)
+				![
+					...EVENTS_SIMULATORS,
+					// TODO: As discussed in https://github.com/testing-library/eslint-plugin-testing-library/issues/1024, this is just a temporary workaround.
+					// We should address the root cause and implement a proper solution instead of explicitly excluding 'user' here.
+					'user',
```
Belco90, line 80: thought(non-blocking): I'd have even added `'user'` to `EVENTS_SIMULATORS`, but it might have side effects in other rules you don't need to handle here.

Belco90 (approved):
Thanks 🙏

Issue #1039, opened 2025-07-10
Support additional async matchers when detecting handled promises


puglyfe:
### What rule do you want to change?

await-async-queries

### Does this change cause the rule to produce more or fewer warnings?

Fewer warnings

### How will the change be implemented?

Expand the matching logic in `hasClosestExpectResolvesRejects()` to include `expectAsync` as well as the [async matchers](https://jasmine.github.io/api/edge/async-matchers) provided by jasmine. The implementation diff would look something like this (I'd probably also rename this function):

```diff
diff --git a/lib/node-utils/index.ts b/lib/node-utils/index.ts
index 4aa8a47..7a5a858 100644
--- a/lib/node-utils/index.ts
+++ b/lib/node-utils/index.ts
@@ -525,6 +525,12 @@ const matcherNamesHandlePromise = [
 	'rejects',
 	'toResolve',
 	'toReject',
+	'toBeRejected',
+	'toBeRejectedWith',
+	'toBeRejectedWithError',
+	'toBePending',
+	'toBeResolved',
+	'toBeResolvedTo',
 ];
 
 /**
@@ -539,7 +545,7 @@ export function hasClosestExpectResolvesRejects(node: TSESTree.Node): boolean {
 		ASTUtils.isIdentifier(node.callee) &&
 		node.parent &&
 		isMemberExpression(node.parent) &&
-		node.callee.name === 'expect'
+		['expect', 'expectAsync'].includes(node.callee.name)
 	) {
 		const expectMatcher = node.parent.property;
 		return (
```

### Example code

```
expectAsync(
	screen.findByText('some text that should never appear')
).toBeRejected();
```

### How does the current rule affect the code?

This code would fail the lint check and would need to be annotated with `/* eslint-disable-next-line testing-library/await-async-utils */`.

### How will the new rule affect the code?

This code would be valid and would require no additional annotation.

### Anything else?

I would understand the hesitancy to add additional framework-specific logic, but it's a fairly trivial addition that would reduce a number of false positives.

### Do you want to submit a pull request to change the rule?

Yes

Belco90:
Hi @puglyfe, thanks for your suggestion! Even better with a PR, it's harder to say no when the feature request comes with a contribution, isn't it? 😉 

I'm not against this idea, actually! This plugin shouldn't be tied to a specific testing runner, so I'm happy to add support to this `exectAsync` (TIL), even if it's not the recommended syntax.

[...]

PR #1040, opened 2025-07-10
feat: expand async matcher support

puglyfe:
## Checks

## Changes

This expands support for additional async matchers (specifically, [jasmine](https://jasmine.github.io/api/edge/async-matchers.html)) to reduce false positives for `await-async-queries` and `await-async-utils`. As mentioned in the linked issue, I understand any hesitancy to include framework-specific logic, but I think this is a relatively trivial addition to open up support to another testing framework.

## Context

Fixes https://github.com/testing-library/eslint-plugin-testing-library/issues/1039

### DIFF
--- a/lib/node-utils/index.ts
+++ b/lib/node-utils/index.ts
@@ -525,6 +525,12 @@ const matcherNamesHandlePromise = [
 	'rejects',
 	'toResolve',
 	'toReject',
+	'toBeRejected',
+	'toBeRejectedWith',
+	'toBeRejectedWithError',
+	'toBePending',
+	'toBeResolved',
+	'toBeResolvedTo',
 ];
 
 /**
@@ -539,7 +545,7 @@ export function hasClosestExpectResolvesRejects(node: TSESTree.Node): boolean {
 		ASTUtils.isIdentifier(node.callee) &&
 		node.parent &&
 		isMemberExpression(node.parent) &&
-		node.callee.name === 'expect'
+		['expect', 'expectAsync'].includes(node.callee.name)
 	) {
 		const expectMatcher = node.parent.property;
 		return (
--- a/lib/node-utils/index.ts
+++ b/lib/node-utils/index.ts
@@ -241,7 +241,7 @@ export function isPromiseHandled(nodeIdentifier: TSESTree.Identifier): boolean {
 			isReturnStatement(node.parent)
 		)
 			return true;
-		if (hasClosestExpectResolvesRejects(node.parent)) return true;
+		if (hasClosestExpectHandlesPromise(node.parent)) return true;
 		if (hasChainedThen(node)) return true;
 		if (isPromisesArrayResolved(node)) return true;
 	});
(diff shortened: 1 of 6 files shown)


lib/node-utils/index.ts
```diff
@@ -525,21 +525,27 @@ const matcherNamesHandlePromise = [
 	'rejects',
 	'toResolve',
 	'toReject',
+	'toBeRejected',
+	'toBeRejectedWith',
```
Belco90, line 529: Can you add a comment before this line specifying these are coming from jasmine?

lib/node-utils/index.ts
```diff
@@ -525,21 +525,27 @@ const matcherNamesHandlePromise = [
...
  */
-export function hasClosestExpectResolvesRejects(node: TSESTree.Node): boolean {
+export function hasClosestExpectHandlesPromise(node: TSESTree.Node): boolean {
 	if (
 		isCallExpression(node) &&
 		ASTUtils.isIdentifier(node.callee) &&
 		node.parent &&
 		isMemberExpression(node.parent) &&
-		node.callee.name === 'expect'
+		['expect', 'expectAsync'].includes(node.callee.name)
```
Belco90, line 548: I'm not sure if we should check that `expect` is combined with the right subset handlers (`resolves`, `rejects`, etc), and `expectAsync` is combined with its subset handlers (`toBeRejectedWith`, `toBeResolved`, etc).

I think it's fine as it is, so we don't need to check if the combination between expect and matcher exists, as the code won't work if doesn't.

tests/lib/rules/await-async-queries.test.ts
```diff
@@ -261,6 +261,24 @@ ruleTester.run(RULE_NAME, rule, {
       `
 		),
 
+		// jasmine async matchers are valid
```
Belco90, line 264: I'm trying to think any invalid case that should be added, but there is none probably, since the improvement only includes more valid combinations.

lib/node-utils/index.ts
```diff
@@ -525,21 +525,27 @@ const matcherNamesHandlePromise = [
...
- * fulfilled by `resolves` or `rejects` properties or
- *  by `toResolve` or `toReject` jest-extended matchers
- *
+ * Determines whether a node belongs to an async assertion that is fulfilled by:
+ * - `resolves` or `rejects` properties
+ * - `toResolve` or `toReject` jest-extended matchers
+ * - jasmine async matchers
  */
-export function hasClosestExpectResolvesRejects(node: TSESTree.Node): boolean {
+export function hasClosestExpectHandlesPromise(node: TSESTree.Node): boolean {
```
Belco90, line 542: This util is used by `isPromiseHandled`, which is used by these rules:
- `await-async-queries`
- `await-async-utils`
- `await-async-events`

You added valid tests for the first two, can you add similar tests for the third one?

Belco90 (commented):
Thanks for your contribution! I added some comments, but the implementation seems fine in general.

lib/node-utils/index.ts
puglyfe: i had the same thought and came to the same conclusion. there will be plenty of other signals elsewhere if an incompatible combination is used - TS, runtime errors, etc.

lib/node-utils/index.ts
puglyfe: added in db05d13d673bfa2dd4f5d175c42c3e3f092e2040

lib/node-utils/index.ts
puglyfe: added in db05d13d673bfa2dd4f5d175c42c3e3f092e2040. i noticed there weren't any `await-async-events` tests for the other async matchers either, so i added those as well.

Belco90 (approved):

Belco90:
@all-contributors please add @puglyfe for code and test

Issue #1041, opened 2025-07-10
no-node-access rule incorrectly triggering from RTK Query syntax in 7.5.4


MrZipper7:
### Have you read the Troubleshooting section?

Yes

### Plugin version

7.5.4

### ESLint version

v9.30.1

### Node.js version

20.19.3

### Bug description

After updating to 7.5.4, all instances of RTK Query's `endpoint.select()` method ([https://redux-toolkit.js.org/rtk-query/api/created-api/endpoints#select](https://redux-toolkit.js.org/rtk-query/api/created-api/endpoints#select)) are triggering `no-node-access` errors.

Ex:
```js
const data = api.endpoints.getAllPosts.select(undefined)(store.getState()).data?.data
                                       ~~~~~~
expect(data).toEqual(expectedData)
```

### Steps to reproduce

1. Define an RTK Query endpoint ([https://redux-toolkit.js.org/rtk-query/api/createApi#createapi-parameters](https://redux-toolkit.js.org/rtk-query/api/createApi#createapi-parameters).
2. Call the `select` function of the endpoint in a test.
3. Observe `the no-node-access` error.

### Error output/screenshots

<img width="1285" height="343" alt="Image" src="https://github.com/user-attachments/assets/49bd4bea-891d-4ceb-9dc7-53d5bbf24e9c" />

### ESLint configuration

```
  {
    files: ['src/**/__test__/*', 'src/**/*.test.ts', 'src/**/*.test.tsx'],
    extends: [
      testingLibrary.configs['flat/react'], 
      vitest.configs.recommended
    ],
    rules: {
      'testing-library/render-result-naming-convention': 'off',
    }
  },
```

### Rule(s) affected

`no-node-access`

### Do you want to submit a pull request to fix this bug?

No

WikiRik:
I'm experiencing a similar issue I think. I'm accidentally running these rules on non-test files as well and I get a lint error on the following piece of code;
```
    const anchor = document.createElement("a");
    anchor.href = "/api/download?fileName=test.csv";
    anchor.click();
```

spennine:
similar issue in our codebase:  element.click() is triggering the lint error for us as well

none of the files cited as violating the rule in our codebase are tests, they're all runtime code

```
                        const element = document.getElementById(id);
                        if (element && !checked) {
                          element.click();
                        }
```

joshkel:
[...]

PR #1042, opened 2025-07-12
fix(no-node-access): Avoid false positives when Testing Library is not imported

y-hsgw:
## Checks

## Changes

- Avoid false positives by reporting only when Testing Library or custom utils module is imported.

## Context

These comments should be resolved with this.
- https://github.com/testing-library/eslint-plugin-testing-library/issues/1041#issuecomment-3061393495
- https://github.com/testing-library/eslint-plugin-testing-library/issues/1041#issuecomment-3062213266


Belco90:
I don't think this is necessary, the plugin should be run only against test files.

PR #1045, opened 2025-07-17
fix: disallow extra properties in rule options

andreww2012:
## Checks

## Changes

Some rules, for example [no-node-access](https://github.com/testing-library/eslint-plugin-testing-library/blob/35e2b406241c816bf52439075dfc61f0964ca494/lib/rules/no-node-access.ts#L37) currently allow extra properties to be passed in options object, which should not be allowed. This makes it easier for typos in rule options to go unnoticed.

This PR simply disallows extra properties in rules' schemas which currently allow them.


Belco90:
@all-contributors please add @andreww2012 for code

Issue #1058, opened 2025-08-08
Promises with `catch` are not considered handled


puglyfe:
### Have you read the Troubleshooting section?

Yes

### Plugin version

7.6.1

### ESLint version

8.57.0

### Node.js version

20.11

### Bug description

We have a number of test cases where we expect async queries to fail, and then we either make assertions against the failure, or throw an error to indicate a particular failure mode. However, async queries with `.catch()` chained are not considered "handled" promises by the plugin.

For example, when testing a custom query:
```
screen.findByMyCustomQuery('thing-that-doesnt-exist').catch((err) => {
  expect(err.message).toMatch(...)
});
```

These tests could be re-written to explicitly use async matchers, but this should be a valid pattern as far as I can tell.

### Steps to reproduce

add this to the `valid` entries in the `await-async-queries` test suite:

```
...createTestCase(
  (query) => `
    ${query}('foo').catch((error) => {
      expect(error.message).toMatch(/my error message/)
    })
  `
)
```

### ESLint configuration

n/a

### Rule(s) affected

`await-async-queries`, `await-async-utils`, `await-async-events`

### Do you want to submit a pull request to fix this bug?

Yes

PR #1059, opened 2025-08-08
fix: `catch` handles promise

puglyfe:
## Checks

## Changes

- replaces `hasChainedThen` with `hasChainedPromiseHandler` in the `isPromiseHandled` logic
- adds tests to all async rules that utilize this code path

I'm a bit curious if there's a reason why this pattern wasn't supported previously. I know it's possible to rewrite tests to use async matchers, but this seems like something that the plugin should allow. Let me know if there's something I'm not considering here...

edit: github cached the PR title from my previous contribution 🤔 not sure what happened there.

## Context

closes https://github.com/testing-library/eslint-plugin-testing-library/issues/1058


Belco90 (approved):
This is great, thanks for your contribution!

PR #1061, opened 2025-08-08
fix: allow un-called async util references

### LINKED ISSUE #1060: `await-async-utils` flags un-called function references

Yes

### Plugin version

7.6.1

### ESLint version

8.57.0

### Node.js version

20.11

### Bug description

I maintain an internal library that wraps testing-library with some additional options and bits of functionality. Part of our test suite asserts that we're exporting everything we think we are and includes one test case that's essentially equivalent to `expect(waitFor).toBeDefined()`. This gets flagged as an unhandled promise.

### Steps to reproduce

Add this to the `await-async-utils` test suite:

```
...ASYNC_UTILS.map((asyncUtil) => ({
  code: `
    import { ${asyncUtil} } from '${testingFramework}';
    test('${asyncUtil} util not called is valid', () => {
      expect(${asyncUtil}).toBeDefined();
    });
  `,
})),
```

### Error output/screenshots

_No response_

### ESLint configuration

n/a

### Rule(s) affected

`await-async-utils` only. `await-async-events` explicitly tests against this pattern, and `await-async-queries` supports it as well (though it isn't explicitly tested against).

### Anything else?

_No response_

Yes

puglyfe:
## Checks

## Changes

This prevents false positives for the case where async utils are referenced but not called. This matches the behavior in both `await-async-events` and `await-async-queries`. `await-async-events` explicitly tests for this behavior [here](https://github.com/puglyfe/eslint-plugin-testing-library/blob/allow-async-refs/tests/lib/rules/await-async-events.test.ts#L43-L51) and [here](https://github.com/puglyfe/eslint-plugin-testing-library/blob/allow-async-refs/tests/lib/rules/await-async-events.test.ts#L282-L290). `await-async-queries` doesn't explicitly test for this, but follows a similar implementation of only processing `CallExpression` nodes rather than `CallExpression Identifier` nodes.

- Allow un-called references in `await-async-utils`
- Add tests to `await-async-queries` to prevent regressions in existing behavior

## Context

fixes #1060

### DIFF
--- a/tests/lib/rules/await-async-queries.test.ts
+++ b/tests/lib/rules/await-async-queries.test.ts
@@ -113,6 +113,11 @@ ruleTester.run(RULE_NAME, rule, {
 			testingFramework: '@marko/testing-library',
 		}),
 
+		// async queries not called are valid
+		...createTestCase((query) => `expect(screen.${query}).toBeDefined()`, {
+			isAsync: false,
+		}),
+
 		// async queries are valid with await operator
 		...createTestCase(
 			(query) => `
--- a/lib/rules/await-async-utils.ts
+++ b/lib/rules/await-async-utils.ts
@@ -121,20 +121,29 @@ export default createTestingLibraryRule<Options, MessageIds>({
 					functionWrappersNames.push((node.id as TSESTree.Identifier).name);
 				}
 			},
-			'CallExpression Identifier'(node: TSESTree.Identifier) {
+			CallExpression(node: TSESTree.CallExpression) {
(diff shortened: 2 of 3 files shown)


tests/lib/rules/await-async-queries.test.ts
```diff
@@ -113,6 +113,11 @@ ruleTester.run(RULE_NAME, rule, {
 			testingFramework: '@marko/testing-library',
 		}),
 
+		// async queries not called are valid
```
puglyfe, line 116: this behavior already existed - just codifying it via test.

Belco90 (approved):

Issue #1063, opened 2025-08-14
no-node-access reports issue with object method


ribeiroguilherme:
### Have you read the Troubleshooting section?

Yes

### Plugin version

v7.6.6

### ESLint version

v9.30.1

### Node.js version

22.17.1

### Bug description

When using a object that has a method named similar to HTML element methods, the no-node-access complains that function calls should not directly access a node.

```js
test('should submit the payment', async () => {
    const pm = new PaymentMethod();
    pm.submit();
    //              ^ - Avoid direct Node access. Prefer using the methods from Testing Library. eslint testing-library/no-node-access 
```

I took a look at the `Aggressive Reporting` section and also in the optional configuration that can be passed to the `no-node-access` rule but couldn't manage to go around this issue.

### Steps to reproduce

Add the following code to a test file:

```js
class PaymentMethod {
    submit() {}
    click() {}
}

test('should submit the payment', () => {
    const pm = new PaymentMethod();
    pm.click();
    //            ^ - Avoid direct Node access. Prefer using the methods from Testing Library. eslint testing-library/no-node-access
    pm.submit();
    //            ^ - Avoid direct Node access. Prefer using the methods from Testing Library. eslint testing-library/no-node-access
    // ...
});
```

### ESLint configuration

```ts
import testingLibrary from 'eslint-plugin-testing-library';
// ...

export default [
  // ...
  {
        name: 'Testing files rules',
        files: ['**/?(*.)+(spec|test).[jt]s?(x)'],
        plugins: {
            'testing-library': testingLibrary
        },
        ...testingLibrary.configs['flat/dom'],
        ...testingLibrary.configs['flat/react']
    }
];
```

### Rule(s) affected

no-node-access

### Do you want to submit a pull request to fix this bug?

No

PR #1064, opened 2025-08-14
fix(no-node-access): narrow detection to Testing Library queries

y-hsgw:
## Checks

## Changes

This PR changes the rule to only report native DOM event method calls (e.g. `click()`, `select()`…) when the receiver is an element obtained via Testing Library queries (e.g. `screen.getBy*,` `findBy*,` `queryBy*`,  chains).

Previously: the rule broadly reported any event handler method that was not fireEvent or userEvent.

Now: the rule reports only when event handler methods are called on results of Testing Library queries.

This reduces false positives while keeping the intended guidance.

## Background

The prior, wide-scoped detection caused false positives when non–Testing Library APIs exposed methods with the same names as DOM event handlers.

Related reports:

- #1063
- #1041

**Limitation**: This rule does not report cases where an element is obtained directly from document, for example:

```ts
document.getElementById('hoge').click();
```

However, we consider this acceptable because, in our setup, the direct element retrieval itself (in this example, `getElementById`) is already reported, so the lack of a report here should not be an issue.

---
This is a proposal PR. Please review, and if there are no concerns, I’ll proceed to broaden the test coverage afterward!

### DIFF
--- a/tests/lib/utils/resolve-to-testing-library-fn.test.ts
+++ b/tests/lib/utils/resolve-to-testing-library-fn.test.ts
@@ -63,6 +63,16 @@ ruleTester.run('esm', rule, {
 			function userClick() {
 				userEvent.click(document.body);
 			}
+    	userClick()
+		`,
+		},
+		{
+			code: `
+			import { userEvent } from './test-utils';
+
+			function userClick() {
+				userEvent.click(document.body);
+			}
     	[].forEach(userClick);
 		`,
 		},


Belco90 (commented):
I like this approach, which is in the same direction of getting rid of the aggressive reporting (I'm still writing an RFC for it), so we only report for Testing Library utils.

> the direct element retrieval itself (in this example, getElementById) is already reported, so the lack of a report here should not be an issue.

I'm not sure about what rule would report `document.getElementById('hoge').click();`, @y-hsgw can you clarify this?

y-hsgw:
Thanks! I’ll proceed with this approach and focus on increasing the test coverage.

> I'm not sure about what rule would report document.getElementById('hoge').click();,

You’re right—that was my mistake; I misunderstood. We don’t currently report `document.getElementById('hoge').click()`. Sorry for the confusion on my part 🙇‍♂️

y-hsgw:
I’ve increased the test coverage and completed the implementation. Please review.

PR #1075, opened 2025-09-03
fix(await-async-event): prevent adding async/await in forEach loops

### LINKED ISSUE #843: await-async-event adds async and await in forEach loop

Yes

### Plugin version

v6.1.2

### ESLint version

v8.45.0

### Node.js version

20.8.1

yarn 1.22.19

### Operating system

macos 13.6.1

### Bug description

async and await doesn't work for `forEach` loop. I think we should give up autofixing this issue.

### Steps to reproduce

```
foo.forEach(() => {
  userEvent.click();
});
```

After autofix, it becomes

```
foo.forEach(async () => {
  await userEvent.click();
});
```

### Error output/screenshots

_No response_

### ESLint configuration

{
   ...
   "testing-library/await-async-event": "error",
}

### Rule(s) affected

await-async-event

### Anything else?

_No response_

Yes

y-hsgw:
## Checks

## Changes

- add support for handling `forEach` in async event method detection

## Context

Fixes #843

### DIFF
--- a/lib/rules/await-async-events.ts
+++ b/lib/rules/await-async-events.ts
@@ -4,6 +4,7 @@ import { createTestingLibraryRule } from '../create-testing-library-rule';
 import {
 	findClosestCallExpressionNode,
 	findClosestFunctionExpressionNode,
+	getDeepestIdentifierNode,
 	getFunctionName,
 	getInnermostReturningFunction,
 	getVariableReferences,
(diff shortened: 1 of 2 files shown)


lib/rules/await-async-events.ts
```diff
@@ -153,6 +154,13 @@ export default createTestingLibraryRule<Options, MessageIds>({
 										findClosestFunctionExpressionNode(node);
 
 									if (functionExpression) {
+										const deepestCalleeIdentifier = getDeepestIdentifierNode(
+											functionExpression.parent
+										);
+										if (deepestCalleeIdentifier?.name === 'forEach') {
```
bonkevin, line 160: (not a maintainer of this package) I would add these commonly mis-used iterator methods
```suggestion
										if (['reduce', 'forEach', 'map'].includes(deepestCalleeIdentifier?.name)) {
```

bonkevin: disregard this comment, it will require more complexity (understand if something has been returned)

PR #1091, opened 2025-09-27
feat(no-wait-for-multiple-assertions): add fix support

y-hsgw:
Follow-up to https://github.com/testing-library/eslint-plugin-testing-library/pull/1090

## Checks

## Changes

- Added partial fix support to the no-wait-for-multiple-assertions rule.
- Enabled autofix for `BlockStatement` cases.

## Context

Fixes https://github.com/testing-library/eslint-plugin-testing-library/issues/1081

### DIFF
--- a/lib/rules/no-wait-for-side-effects.ts
+++ b/lib/rules/no-wait-for-side-effects.ts
@@ -42,6 +42,8 @@ export default createTestingLibraryRule<Options, MessageIds>({
 	},
 	defaultOptions: [],
 	create(context, _, helpers) {
+		const sourceCode = getSourceCode(context);
+
 		function isCallerWaitFor(
 			node:
 				| TSESTree.AssignmentExpression
@@ -155,8 +157,6 @@ export default createTestingLibraryRule<Options, MessageIds>({
 				}
 				return false;
 			});
-
-			return false;
 		}
 
 		function getSideEffectNodes(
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-wait-for-side-effects.test.ts
```diff
@@ -407,6 +445,13 @@ ruleTester.run(RULE_NAME, rule, {
         })
       `,
 					errors: [{ line: 5, column: 11, messageId: 'noSideEffectsWaitFor' }],
+					output: `
+        import { waitFor } from '${testingFramework}';
+        import { renderHelper } from 'somewhere-else';
+        renderHelper(<App />)
+        await waitFor(() => {
+        })
+      `,
```
Belco90, line 454: question: Shouldn't this one remove the empty `waitFor`?

Similar to the same case but with an arrow function (the case above).

tests/lib/rules/no-wait-for-side-effects.test.ts
```diff
@@ -675,6 +825,13 @@ ruleTester.run(RULE_NAME, rule, {
         })
       `,
 					errors: [{ line: 5, column: 11, messageId: 'noSideEffectsWaitFor' }],
+					output: `
+        import { waitFor } from '${testingFramework}';
+        fireEvent.keyDown(input, {key: 'ArrowDown'})
+        await waitFor(() => {
+          expect(b).toEqual('b')
+        })
```
Belco90, line 833: praise: Nice! 🏆

lib/rules/no-wait-for-side-effects.ts
```diff
@@ -204,10 +214,58 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+						);
+						const sideEffectNodeText = sideEffectLines.join('\n').trimStart();
+						if (
+							sideEffects.length === node.body.length &&
+							sideEffects.length === 1
+						) {
+							const tokenAfter = sourceCode.getTokenAfter(targetNode);
+							return [
+								fixer.insertTextBefore(targetNode, sideEffectNodeText),
+								tokenAfter?.value === ';'
```
Belco90, line 243: thought(non-blocking): I'd say this is a brittle implementation, since the `;` operator is optional. However, from what I've seen in the tests, the basic scenarios are covered, so I'm happy to go ahead with this and iterate later if necessary.

PR #1097, opened 2025-10-03
feat(no-wait-for-multiple-assertions): avoid reporting unrelated assertions

### LINKED ISSUE #1084: Refine no-wait-for-multiple-assertions: Only forbid multiple assertions on the same async target

no-wait-for-multiple-assertions

Fewer warnings

The rule no-wait-for-multiple-assertions currently forbids any multiple assertions inside a waitFor callback.
The proposal is to refine the rule so that it only reports an error when multiple assertions are made against the same asynchronous target.

This keeps the original intention (avoid useless timeouts when combining multiple checks on the same async call) while avoiding unnecessary errors in cases where separate async targets are being tested together.

### Example code

✅ Current rule reports (but should still report)
```ts
await waitFor(() => {
  expect(window.fetch).toHaveBeenCalledWith('/foo');
  expect(window.fetch).toHaveBeenCalledTimes(1);
});
```

Here, both assertions relate to the same async target (window.fetch).
If one of them is permanently failing, waitFor will unnecessarily retry until timeout.
The rule should still report this as a violation.

❌ Current rule reports (but should NOT report anymore)
```ts
await waitFor(() => {
  expect(window.fetch).toHaveBeenCalledWith('/foo');
  expect(localStorage.setItem).toHaveBeenCalledWith('bar', 'baz');
});
```

In this case, two different async side-effects are being checked (fetch and localStorage).
If both are eventually satisfied, waiting time is identical whether assertions are combined or separated.
If one fails permanently, the timeout will still happen regardless of being combined or separated.
So forcing developers to split them adds verbosity without clear benefit.

- It flags all multiple assertions inside waitFor, even when they target different async behaviors.
- This often forces developers to artificially split assertions into multiple waitFor calls, which can hurt readability and lead to more verbose tests.

- Multiple assertions against the same async target will still be forbidden.
- Multiple assertions against different async targets will be allowed.
- Developers will no longer need to split unrelated assertions into separate waitFor calls.

### Anything else?

There are also practical cases where multiple assertions in the same waitFor should be acceptable, because they are targeting different async targets and are expected to resolve independently. For example:

```tsx
export function Example() {
  const [a, setA] = useState('');
  const [b, setB] = useState('');

  useEffect(() => {
    setTimeout(() => setA('a'), 100);
    setTimeout(() => setB('b'), 500);
  }, []);

  return (
    <div>
      <div data-testid="a">{a}</div>
      <div data-testid="b">{b}</div>
    </div>
  );
}

it('Example', async () => {
  render(<Example />);

  await waitFor(() => {
    expect(screen.getByTestId('a')).toHaveTextContent('a');
    expect(screen.getByTestId('b')).toHaveTextContent('b');
  });
});

```

In this case, both `a` and `b` are independent async state updates.
Splitting them into two separate waitFor calls does not reduce the timeout risk (if one fails, it will still time out), but it does add verbosity.
Allowing multiple assertions on different async targets would make this kind of test more concise without sacrificing reliability.

Yes

y-hsgw:
## Checks

## Changes

- Updated rule logic to better distinguish between unique async elements.
- Update doc and test.

## Context

Fixes #1084 

---
This is completely unrelated to the changes in this PR, but I noticed an [error](https://github.com/testing-library/eslint-plugin-testing-library/actions/runs/18224863863/job/51893411818?pr=1097#step:8:273) in the "Upload coverage reports to Codecov" step, and the coverage notification is not being sent.  
Is this something we should be concerned about, or is it fine to ignore?

### DIFF
--- a/docs/rules/no-wait-for-multiple-assertions.md
+++ b/docs/rules/no-wait-for-multiple-assertions.md
@@ -7,9 +7,13 @@
 ## Rule Details
 
 This rule aims to ensure the correct usage of `expect` inside `waitFor`, in the way that they're intended to be used.
-When using multiple assertions inside `waitFor`, if one fails, you have to wait for a timeout before seeing it failing.
-Putting one assertion, you can both wait for the UI to settle to the state you want to assert on,
-and also fail faster if one of the assertions do end up failing
+
+If you use multiple assertions against the **same asynchronous target** inside `waitFor`,
+you may have to wait for a timeout before seeing a test failure, which is inefficient.
+Therefore, you should avoid using multiple assertions on the same async target inside a single `waitFor` callback.
+
+However, multiple assertions against **different async targets** (for example, independent state updates or different function calls) are allowed.
+This avoids unnecessary verbosity and maintains readability, without increasing the risk of missing failures.
 
 Example of **incorrect** code for this rule:
 
@@ -17,13 +21,13 @@ Example of **incorrect** code for this rule:
 const foo = async () => {
 	await waitFor(() => {
 		expect(a).toEqual('a');
-		expect(b).toEqual('b');
+		expect(a).toEqual('a');
 	});
 
 	// or
 	await waitFor(function () {
 		expect(a).toEqual('a');
-		expect(b).toEqual('b');
+		expect(a).toEqual('a');
 	});
 };
 ```
@@ -33,13 +37,13 @@ Examples of **correct** code for this rule:
 ```js
 const foo = async () => {
 	await waitFor(() => expect(a).toEqual('a'));
-	expect(b).toEqual('b');
+	expect(a).toEqual('a');
 
 	// or
 	await waitFor(function () {
 		expect(a).toEqual('a');
 	});
-	expect(b).toEqual('b');
+	expect(a).toEqual('a');
 
 	// it only detects expect
 	// so this case doesn't generate warnings
--- a/docs/rules/no-wait-for-multiple-assertions.md
+++ b/docs/rules/no-wait-for-multiple-assertions.md
@@ -20,14 +20,14 @@ Example of **incorrect** code for this rule:
 ```js
 const foo = async () => {
 	await waitFor(() => {
-		expect(a).toEqual('a');
-		expect(a).toEqual('a');
+		expect(window.fetch).toHaveBeenCalledWith('/foo');
+		expect(window.fetch).toHaveBeenCalledWith('/foo');
 	});
 
 	// or
 	await waitFor(function () {
-		expect(a).toEqual('a');
-		expect(a).toEqual('a');
+		expect(window.fetch).toHaveBeenCalledWith('/foo');
+		expect(window.fetch).toHaveBeenCalledWith('/foo');
 	});
 };
 ```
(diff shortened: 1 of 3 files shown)


docs/rules/no-wait-for-multiple-assertions.md
```diff
@@ -7,23 +7,27 @@
...
+This avoids unnecessary verbosity and maintains readability, without increasing the risk of missing failures.
 
 Example of **incorrect** code for this rule:
 
 ```js
 const foo = async () => {
 	await waitFor(() => {
 		expect(a).toEqual('a');
-		expect(b).toEqual('b');
+		expect(a).toEqual('a');
```
Belco90, line 24: suggestion: please update the examples with a more realistic scenario.

The example described in the original issue works better, where 2 assertions were checking number of calls and args called for the same mock.

lib/rules/no-wait-for-multiple-assertions.ts
```diff
@@ -62,14 +85,30 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+				expectArgumentMap.set(argumentText, newTargetNodes);
 			}
 
-			for (let i = 0; i < expectNodes.length; i++) {
-				if (i !== 0) {
+			for (const expressionStatements of expectArgumentMap.values()) {
+				if (expressionStatements.length <= 1) {
+					continue;
+				}
+				for (const expressionStatement of expressionStatements.slice(1)) {
```
Belco90, line 109: suggestion: please add a comment clarifying why we are slicing from index 1.

I know this is to report from the 2nd occurrence of the same target, but better to leave an explicit comment about it.

lib/rules/no-wait-for-multiple-assertions.ts
```diff
@@ -62,14 +85,30 @@ export default createTestingLibraryRule<Options, MessageIds>({
...
+				const argumentText = getSourceCode(context).getText(argument);
+				const existingNodes = expectArgumentMap.get(argumentText) ?? [];
+				const newTargetNodes = [...existingNodes, expectNode];
+				expectArgumentMap.set(argumentText, newTargetNodes);
 			}
 
-			for (let i = 0; i < expectNodes.length; i++) {
-				if (i !== 0) {
+			for (const expressionStatements of expectArgumentMap.values()) {
+				if (expressionStatements.length <= 1) {
```
Belco90, line 106: question: is this actually needed?

If `expressionStatements` has 0 or 1 element, the slice in the for loop would produce an empty array, avoiding any iteration and ending in the same result.

lib/rules/no-wait-for-multiple-assertions.ts
y-hsgw: Added a comment to clarify why we slice from index 1. Thanks for the suggestion! [908e84e](https://github.com/testing-library/eslint-plugin-testing-library/pull/1097/commits/908e84e5dfe0857e470160b4278e439030b887cb)

docs/rules/no-wait-for-multiple-assertions.md
y-hsgw: You're absolutely right! I've updated the sample code. [9a7199b](https://github.com/testing-library/eslint-plugin-testing-library/pull/1097/commits/9a7199b64d5970ee3b8f5bdc6c38dca41a5da888)

lib/rules/no-wait-for-multiple-assertions.ts
y-hsgw: You're right — that conditional was unnecessary. I've removed it. Thanks for pointing it out! [ed30651](https://github.com/testing-library/eslint-plugin-testing-library/pull/1097/commits/ed3065130ad64f3970278a151f1b65609d46af6b)

Belco90:
> This is completely unrelated to the changes in this PR, but I noticed an [error](https://github.com/testing-library/eslint-plugin-testing-library/actions/runs/18224863863/job/51893411818?pr=1097#step:8:273) in the "Upload coverage reports to Codecov" step, and the coverage notification is not being sent.
> Is this something we should be concerned about, or is it fine to ignore?

I need to check this independently. Would you mind creating a separate issue for it? @y-hsgw

PR #1102, opened 2025-10-09
fix: generate configs and rules statically

Belco90:
## Checks

## Changes

- Replace jest with vitest, including scripts and eslint plugin
- In order to get certain tests running correctly with vitest, I've removed the dynamic generation for configs and rules, so they are generated statically. The plugin should work the same way for the final user.

## Context

Relates to #855

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
 		"@types/node": "^22.9.3",
 		"@typescript-eslint/rule-tester": "^8.15.0",
 		"@vitest/coverage-v8": "^3.2.4",
-		"@vitest/eslint-plugin": "^1.3.16",
+		"@vitest/eslint-plugin": "^1.5.2",
 		"del-cli": "^6.0.0",
 		"eslint": "^9.35.0",
 		"eslint-config-prettier": "^10.1.8",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -40,7 +40,7 @@ importers:
         specifier: ^3.2.4
         version: 3.2.4(vitest@3.2.4(@types/node@22.15.29)(jiti@2.5.1)(yaml@2.7.0))
       '@vitest/eslint-plugin':
-        specifier: ^1.3.16
+        specifier: ^1.5.2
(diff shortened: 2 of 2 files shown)


lib/configs/index.ts
```diff
@@ -1,19 +1,20 @@
...
+const legacyConfigs: Record<SupportedTestingFramework, Linter.LegacyConfig> = {
+	dom,
+	angular,
+	react,
+	vue,
+	svelte,
+	marko,
+};
 
-export default SUPPORTED_TESTING_FRAMEWORKS.reduce(
```
Belco90: Remove dynamic generation in favor of static generation.

lib/rules/index.ts
```diff
@@ -1,22 +1,59 @@
...
+import renderResultNamingConvention from './render-result-naming-convention';
 
-import { importDefault } from '../utils';
-
-import type { TestingLibraryPluginRuleModule } from '../utils';
-
-const rulesDir = __dirname;
-const excludedFiles = ['index'];
-
-export default readdirSync(rulesDir)
```
Belco90: Remove dynamic generation in favor of static generation.

tests/index.test.ts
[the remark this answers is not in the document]
Belco90: Fixed in [c2ddd3a](https://github.com/testing-library/eslint-plugin-testing-library/pull/1102/commits/c2ddd3a660055d33faad6c1e3ebf93b7926489f9)

tests/index.test.ts
```diff
@@ -1,48 +1,46 @@
...
+
+import { it, expect } from 'vitest';
 
 import plugin from '../lib';
 
-const numberOfRules = 28;
 const ruleNames = Object.keys(plugin.rules);
 
-// eslint-disable-next-line jest/expect-expect
+it('should export all existing rules', () => {
```
Belco90: This new test will take care of making sure that all rules within `libs/rules` are actually re-exported, so can be consumed by users installing the plugin.

The purpose for this check is to improve the DX when working on new rules, since it's quite easy to forget re-exporting the rules to the public API of the plugin.

Here, a demo of the test catching a rule not re-exported properly:

https://github.com/user-attachments/assets/f92eb389-e323-4d2b-aa72-72e351ece88d

PR #1106, opened 2025-10-14
add react-native library module

deecewan:
## Checks

## Changes

the current system does not support react native, which means that some rules (like `no-node-access`) never work with react native. it's possible to set a custom module, but only one of them, so I can't set our actual custom module and `@testing-library/react-native`.

-

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -25,6 +25,7 @@ const LIBRARY_MODULES = [
 	'@testing-library/dom',
 	'@testing-library/angular',
 	'@testing-library/react',
+	'@testing-library/react-native',
 	'@testing-library/preact',
 	'@testing-library/vue',
 	'@testing-library/svelte',


Belco90:
Hi @deecewan. The plugin doesn't support yet React Native. We have a project to support this framework, so I'm closing this PR in favor of including it in that project.

deecewan:
@Belco90 we currently use this eslint plugin for react-native-testing-library and it works great for the most part.

Can you point me to the eslint plugin specific to react native?

Belco90:
> @Belco90 we currently use this eslint plugin for react-native-testing-library and it works great for the most part.
> 
> Can you point me to the eslint plugin specific to react native?

@deecewan I'm afraid there is no specific plugin for react-native.

I'm happy to implement the proper React Native support in the plugin, but I need to understand the impact of that framework across the rules in this plugin. Please, [add more info to this discussion](https://github.com/testing-library/eslint-plugin-testing-library/discussions/1136) so we can add such framework config to the plugin.

PR #1125, opened 2025-12-10
feat(prefer-user-event-setup): add new rule

### LINKED ISSUE #646: Recommend using the methods on the instances returned by userEvent.setup() from userEvent
### Name for new rule

prefer-userEvent-setup

Rule should recommend using the methods on the instances returned by userEvent.setup() from userEvent as inline or using setup functions.

### Testing Library feature

Enforce using the methods on the instances returned by userEvent.setup()

React

_No response_

### Code examples

// inlining
test('trigger some awesome feature when clicking the button', async () => {
  const user = userEvent.setup()
  render(<MyComponent />)

  await user.click(screen.getByRole('button', {name: /click me!/i}))

  // ...assertions...
})

// setup function
function setup(jsx) {
  return {
    user: userEvent.setup(),
    ...render(jsx),
  }
}

test('render with a setup function', async () => {
  const {user} = setup(<MyComponent />)
  // ...
})

### Anything else?

_No response_

Yes, but need help

gipcompany:
## Checks

## Changes

- Add prefer-user-event-setup rule to encourage using [userEvent.setup() pattern](https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent) over direct method calls
- Add comprehensive test coverage with 44 test cases covering various usage scenarios  
- Add documentation explaining the rule benefits and correct usage patterns
- Update README.md with new rule entry in supported rules table

## Context

Closes #646

### DIFF
--- a/lib/rules/index.ts
+++ b/lib/rules/index.ts
@@ -25,6 +25,7 @@ import preferQueryByDisappearance from './prefer-query-by-disappearance';
 import preferQueryMatchers from './prefer-query-matchers';
 import preferScreenQueries from './prefer-screen-queries';
 import preferUserEvent from './prefer-user-event';
+import preferUserEventSetup from './prefer-user-event-setup';
 import renderResultNamingConvention from './render-result-naming-convention';
 
 export default {
@@ -55,5 +56,6 @@ export default {
 	'prefer-query-matchers': preferQueryMatchers,
 	'prefer-screen-queries': preferScreenQueries,
 	'prefer-user-event': preferUserEvent,
+	'prefer-user-event-setup': preferUserEventSetup,
 	'render-result-naming-convention': renderResultNamingConvention,
 };


docs/rules/prefer-user-event-setup.md
```diff
@@ -0,0 +1,118 @@
...
+
+test('using custom render', async () => {
+	const { user, getByRole } = renderWithUser(<MyComponent />);
+	await user.click(getByRole('button'));
+});
+```
+
+## When Not To Use This Rule
+
+If you're using an older version of user-event (< v14) that doesn't support or require the setup pattern, you may want to disable this rule.
```
Belco90: issue(blocking): We need to mention here another reason for not using this rule: if the `userEvent.setup()` is called in a custom render outside the test file.

lib/rules/prefer-user-event-setup.ts
```diff
@@ -0,0 +1,221 @@
...
+				}
+			},
+
+			// Track variables assigned from userEvent.setup()
+			VariableDeclarator(node: TSESTree.VariableDeclarator) {
+				if (!userEventIdentifier || !node.init) return;
+
+				// Direct assignment: const user = userEvent.setup()
+				if (
+					isUserEventSetupCall(node.init) &&
```
Belco90: question(blocking): can't we reuse `isUserEventMethod` helper for this?

tests/lib/rules/prefer-user-event-setup.test.ts
```diff
@@ -0,0 +1,311 @@
...
+				
+				test('example', async () => {
+					const user = userEvent.setup();
+					await user.click(element);
+				});
+			`,
+		},
+	],
+
+	invalid: [
```
Belco90: issue(blocking): Please check `line` and `column` for errors in invalid cases.

lib/rules/prefer-user-event-setup.ts
gipcompany: Good catch! Updated to use `helpers.isUserEventMethod` instead of the custom `isDirectUserEventMethodCall` function.

f75b0f8787556ef9f1821e42871a8eccdd57bfab

tests/lib/rules/prefer-user-event-setup.test.ts
```diff
@@ -0,0 +1,311 @@
+import rule, { RULE_NAME } from '../../../lib/rules/prefer-user-event-setup';
+import { createRuleTester } from '../test-utils';
+
+import type { MessageIds } from '../../../lib/rules/prefer-user-event-setup';
+
+const ruleTester = createRuleTester();
+
+const USER_EVENT_METHODS = [
```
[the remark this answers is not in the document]
gipcompany: Done! Extracted `USER_EVENT_METHODS` to `lib/utils/index.ts` and updated all references.

49c67549d0f0b399c7cfc55d6bef0c767c292efd

gipcompany:
> The rule must be manually exported in `lib/rules/index.ts`

This was addressed with 8510411d2fd094cbe418a569e8a122341f9fe437.

Belco90:
@all-contributors please add @gipcompany for code, test, docs

bobaaaaa:
awesome new rule. thx @gipcompany ❤️

PR #1211, opened 2026-02-08
feat: support ESLint v10

G-Rath:
## Checks

## Changes

- update peer dependency range to allow ESLint v10
- update `typescript-eslint` dependencies (in particular `rule-tester`) to work with ESLint v10

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,8 +54,8 @@
 		"type-check": "tsc --noEmit"
 	},
 	"dependencies": {
-		"@typescript-eslint/scope-manager": "^8.51.0",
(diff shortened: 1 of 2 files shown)


G-Rath:
Pretty sure this requires https://github.com/typescript-eslint/typescript-eslint/issues/11952 to be resolved, maybe due to the use of `@typescript-eslint/rule-tester` since in other codebases like `eslint-plugin-jest` and `eslint-plugin-jest-dom` we've been able to support v10 without error.

I'll try dig into this later today

PR #1278, opened 2026-03-31
feat: add `no-unsettled-absence-query` rule

### LINKED ISSUE #1237: New Rule Proposal: no-unsettled-absence-query
### Name for new rule

no-unsettled-absence-query

Asserting absence with `queryBy*` + `.not.toBeInTheDocument()` immediately after `render()`, before the component has settled, can produce a false positive. The element isn't there *yet*, not because it *won't* be there. The community calls this _Testing Ghosts_.

Testing Library's `render()` does flush the initial synchronous render (it's wrapped in `act()` internally), so for a plain component with no async behavior, checking absence right after `render()` is technically correct. The problem shows up when the component has async side effects like data fetching, lazy loading, state updates from `useEffect`, etc. In those cases the UI hasn't finished updating, and the absence check passes vacuously. Since the rule can't know whether a component has async behavior, it flags all unsettled absence assertions and lets the developer confirm intent by adding a settling expression.

This rule would enforce that an absence assertion (`expect(screen.queryBy*(...)).not.toBeInTheDocument()` / `.not.toBeVisible()`) must be preceded by a "settling" expression within the test body.

**Why existing rules don't cover this:**

| Existing Rule | Why it doesn't solve this |
|---|---|
| `prefer-presence-queries` | Enforces correct query type, but doesn't track timing relative to render |
| `no-await-sync-queries` | Only flags `await` directly on sync queries, not timing of assertions |
| `await-async-queries` / `await-async-utils` | About promise handling, not assertion ordering |
| `prefer-find-by` | Simplifies `waitFor + getBy` -> `findBy`, doesn't detect premature absence checks |

Related issues: #269, #411, #518 - all adjacent but none address this specific pattern.

### Testing Library feature

This rule relates to the absence/presence assertion pattern described in the [Appearance and Disappearance guide](https://testing-library.com/docs/guide-disappearance/). Specifically, it enforces correct ordering of `queryBy*` absence assertions relative to async settling utilities (`findBy*`, `waitFor`, `act`) and sync settling queries (`getBy*`).

All frameworks: React, Angular, Vue, Svelte, etc. The pattern applies anywhere `screen.queryBy*` + `.not.toBeInTheDocument()` is used.

Warns about a potential error. Less valuable as an error given the call-out above of vanilla components.

### Code examples

**Fail:**

```js
// Absence assertion before component has settled, "Testing a Ghost"
test('shows no error', () => {
  render(<AsyncComponent />);
  expect(screen.queryByText('error')).not.toBeInTheDocument();
});

// Absence assertion BEFORE the awai, order matters
test('shows no error', async () => {
  render(<AsyncComponent />);
  expect(screen.queryByText('error')).not.toBeInTheDocument();
  await screen.findByText('loaded');
});

// queryAllBy variant
test('shows no alerts', () => {
  render(<AsyncComponent />);
  expect(screen.queryAllByRole('alert')).not.toBeInTheDocument();
});

// Absence assertion inside waitFor - passes on first retry, still a ghost
test('shows no error', async () => {
  render(<AsyncComponent />);
  await waitFor(() => {
    expect(screen.queryByText('error')).not.toBeInTheDocument();
  });
});
```

**Pass:**

```js
// findBy* settles the component first
test('shows no error', async () => {
  render(<AsyncComponent />);
  await screen.findByText('loaded');
  expect(screen.queryByText('error')).not.toBeInTheDocument();
});

// waitFor settles the component first
test('shows no error', async () => {
  render(<AsyncComponent />);
  await waitFor(() => expect(something).toBe(true));
  expect(screen.queryByRole('alert')).not.toBeInTheDocument();
});

// act settles the component first
test('shows no error', async () => {
  await act(() => render(<AsyncComponent />));
  expect(screen.queryByText('error')).not.toBeInTheDocument();
});

// getBy* proves sync render completed
test('shows no error', () => {
  render(<Component />);
  screen.getByText('visible heading');
  expect(screen.queryByText('error')).not.toBeInTheDocument();
});

```

**What counts as "settled":**

1. **Any `await` expression on a preceding statement** — covers `findBy*`, `waitFor`, `act`, custom async helpers
2. **A `screen.getBy*` / `screen.getAllBy*` call** — proves the sync render produced expected output

**Matchers checked:** `.not.toBeInTheDocument()`, `.not.toBeVisible()`

### Anything else?

**References:**
- [Testing Library: Appearance and Disappearance guide](https://testing-library.com/docs/guide-disappearance/)
- [Kent C. Dodds: Common mistakes with React Testing Library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library)
- [Gerardo Perrucci: Stop Testing Ghosts](https://www.gperrucci.com/blog/react/assert-non-existence-react-testing-library)
- [react-testing-library #865](https://github.com/testing-library/react-testing-library/issues/865) (vague similarity; in the same realm with component mounting/unmounting states)

I have a working implementation with full test coverage that I'm happy to submit as a PR if this proposal is accepted.

JMiltier:
## Checks

## Changes

1. Add `no-unsettled-absence-query` rule that detects absence assertions (`queryBy*` + `.not.toBeInTheDocument()` / `.not.toBeVisible()` / `.toBeNull()` / `.toBeFalsy()`) before the component has settled                  
2. Flags absence assertions inside `waitFor` callbacks (pass on first invocation)
3. Recognizes settling via `await` expressions (`findBy*`, `waitFor`, `act`) or `getBy*`/`getAllBy*` calls
4. Scope-aware tree-walking: await inside a nested function does not settle the outer scope
5. 37 test cases across React, Vue, Angular, DOM, and Marko
6. `recommendedConfig` set to `false` for all frameworks (opt-in only) - open to adding as `warn` in presets if preferred 
7. Documentation: https://github.com/JMiltier/eslint-plugin-testing-library/blob/pr/no-unsettled-absence-query/docs/rules/no-unsettled-absence-query.md

## Context

Closes #1237


Belco90:
@JMiltier thanks for your PR, all checks passing now! I'll try to review this during the week.

src/rules/no-unsettled-absence-query.ts
```diff
@@ -0,0 +1,225 @@
...
+			while (current) {
+				if (
+					(isArrowFunctionExpression(current) ||
+						isFunctionExpression(current)) &&
+					isCallExpression(current.parent)
+				) {
+					const calleeIdentifier = getDeepestIdentifierNode(
+						current.parent.callee
+					);
+					if (calleeIdentifier && helpers.isAsyncUtil(calleeIdentifier)) {
```
Belco90, line 122: I'm not sure if `isAsyncUtil` is the right util to use here. This helper matches both `waitFor` and `waitForElementToBeRemoved` by default, but the latter shouldn't be reported, right? It's not even mentioned in the docs, and IIRC `waitForElementToBeRemoved` throws an error if the first argument (the element that should be removed) is already empty (i.e. it's a settling expression on its own).

You can limit the subset of async utils checked passing a second arg to `isAsyncUtil` (other rules like `no-wait-for-side-effects` are doing this).

Additionally, mention this in the docs, and add extra tests to check the usage of `waitForElementToBeRemoved` is valid. Something like:

```js
{
    code: `
    import { render, screen, waitForElementToBeRemoved } from '@testing-library/react'

    test('waitForElementToBeRemoved callback is not a waitFor', async () => {
      render(<Component />)
      await waitForElementToBeRemoved(() => screen.queryByText('loading'))
      expect(screen.queryByText('error')).not.toBeInTheDocument()
    })
    `,
  },
```

docs/rules/no-unsettled-absence-query.md
```diff
@@ -0,0 +1,86 @@
...
+});
+
+// getBy* proves sync render completed
+test('shows no error', () => {
+	render(<Component />);
+	screen.getByText('visible heading');
+	expect(screen.queryByText('error')).not.toBeInTheDocument();
+});
+```
+
```
Belco90, line 81: We need a "Limitations" section to mention that some untracked helpers (custom setup, fake timers, etc) might produce false positives.

src/rules/no-unsettled-absence-query.ts
```diff
@@ -0,0 +1,225 @@
...
+
+function isNestedFunction(node: TSESTree.Node): boolean {
+	return (
+		isArrowFunctionExpression(node) ||
+		isFunctionExpression(node) ||
+		isFunctionDeclaration(node)
+	);
+}
+
+function containsNode(
```
Belco90, line 32: I'm worried about the performance of this util because of the following issues:

1. It revisits the tree per assertion. `hasSettlingExpression` calls this util twice for every preceding statement of every absence assertion, resulting in O(n²) complexity.
2. It iterates over non-AST keys. Skipping only `parent` means that it also traverse into `loc`, `range`, comment arrays, etc. It works because those don't carry a `type` but it's fragile, especially in terms of maintenance.
3. It's not consistent with the rest of the rules in the repo. We have a bunch of utils to walk a subtree in `node-utils`. You should be able to report awaited expressions with ESLint's visitor pattern instead (take a look at https://github.com/testing-library/eslint-plugin-testing-library/blob/main/src/rules/no-await-sync-events.ts#L99 or https://github.com/testing-library/eslint-plugin-testing-library/blob/2002150540c9fc4b59bb95b63fc221334f76f2f2/src/rules/prefer-find-by.ts#L338 for reference).

The rule implementation must be refactored to get rid of this `containsNode` function and let ESLint do the single pass, collect all settling nodes (`AwaitExpression` + expression, checking they are `getQuery` with the corresponding helper), then check what absence assertions must be reported at `Program:exit`.

src/rules/no-unsettled-absence-query.ts
```diff
@@ -0,0 +1,225 @@
...
+				vue: false,
+				svelte: false,
+				marko: false,
+			},
+		},
+		messages: {
+			noUnsettledAbsenceQuery:
+				'Absence assertion on `{{queryMethod}}` appears before the component has settled. ' +
+				'The element may not have rendered yet, resulting in a false positive. ' +
+				'Add an `await` expression (e.g. `findBy*`, `waitFor`, `act`) or a `getBy*` call before this assertion.',
```
Belco90, line 90: Let's make the error message shorter. Right now it includes the "what" + "why" + "how to fix", but with the "what" + "how to fix" should be enough. If the user wants to know more about the why, they can go to the docs to get more info.

```suggestion
				'Absence assertion on `{{queryMethod}}` appears before the component has settled. ' +
				'Add an `await` expression (e.g. `findBy*`, `waitFor`, `act`) or a `getBy*` call before this assertion.',
```

Belco90 (changes requested):

JMiltier:
Thanks @Belco90! I'll take a look at these requested changes soon

docs/rules/no-unsettled-absence-query.md
JMiltier: https://github.com/testing-library/eslint-plugin-testing-library/pull/1278/changes/647d457b67c3dd988a0ff0baf30fde03ec59b35d Added a "Limitations" section with four specific call-outs from what I found. Included a way to suppress any warnings based on how users set this rule in their config

src/rules/no-unsettled-absence-query.ts
JMiltier: https://github.com/testing-library/eslint-plugin-testing-library/pull/1278/commits/45c273706f9d60d2bc94dc010d9459d1430dc308 updated - matched the patterns from no-wait-for-side-effects and prefer-find-by, along with some renaming to align. Docs were updated as well so `waitForElementToBeRemoved` is listed as a settling expression. Test added verbatim (thanks!)

src/rules/no-unsettled-absence-query.ts
JMiltier: https://github.com/testing-library/eslint-plugin-testing-library/pull/1278/commits/ee2c337f6a9320fc4459b91671ad4e2bed65cd8b committed this suggested change, but debating if the `act` portion of this conflicts with the `no-unnecessary-act` rule that is enabled by default for the recommended configs ([react](https://github.com/testing-library/eslint-plugin-testing-library/blob/a53330b638ea8f64ee8f90994d1586ba7797eae0/src/configs/react.ts#L28)). Can drop that out of this as well if that could cause some confusion/conflict. Would need to rework the associated test as well (below)

```
{
	code: `
	import { render, screen, act } from '@testing-library/react'

	test('settled by act wrapping render', async () => {
		await act(() => render(<Component />))
		expect(screen.queryByText('error')).not.toBeInTheDocument()
	})
	`,
},
```

src/rules/no-unsettled-absence-query.ts
JMiltier: https://github.com/testing-library/eslint-plugin-testing-library/pull/1278/commits/e14c293b4759d6e1254a9c45918ac9c46b8eec76 this was a good learning session to do a single AST pass (the visitor pattern you mentioned) instead of revisiting. Kept the same (or very similar) collection/resolution shape as `prefer-explicit-assert` and `prefer-implicit-assert`. Also ran some synthetic benchmarking before/after, which is now closer to linear instead of quadratic. Both `containsNode` and `hasSettlingExpression` are gone since the single pass replaces what they were doing

JMiltier:
@Belco90 - let me know if you still find this PR valuable. I may close this and the issue, since it's an optional and not fully encompassing feature addition. Haven't seen any `+` support on the issue (edit: from others*)

Belco90:
@JMiltier I find it valuable! It's just I didn't have the proper time to finish the review.

